Issue #186, opened 2014-11-06
Remove FunctionalTests.swift and FunctionalTests+ObjC.m


modocache:
Part of the aim for #127 was to get rid of these test files, which have only ever served as a basic Quick sanity check (as opposed to a comprehensive set of tests).

Functional tests (using `qck_runSpec`) should be added to cover any functionality tested in these two files but not covered by the tests in [the `FunctionalTests` directory](https://github.com/Quick/Quick/tree/b51fd886741ed4765bc4b66bd58461c65a9b0262/QuickTests/FunctionalTests). Once that's done, these two should be removed.

modocache:
Addressed in #193! Thanks again, @sidraval! :+1:

PR #193, opened 2014-11-10
Remove functional test files.

sidraval:
In reference to issue #186


modocache:
Thanks, @sidraval! It took a while, but it's finally in! :100: :sparkles:

Issue #199, opened 2014-11-19
README: Add instructions for including Quick in a project via Carthage


modocache:
See: https://github.com/Carthage/Carthage

Instructions should live in the "Installing Quick" section, above Git submodules and next to the CocoaPods instructions in #197. They should include a sample Cartfile.

raven:
This has been merged.

modocache:
Good call, thanks @raven! :+1:

PR #203, opened 2014-11-23
Add Carthage instructions

raven:
Added Carthage instructions and notes regarding the required **Copy Files** build phase.
For #199


raven:
@paulyoung added note regarding this is but one way to use Carthage.

paulyoung:
:cool: :sunglasses:

I think we should merge this and address any further concerns / updates to Carthage as they arise.

modocache:
Go for it, @paulyoung! Hit the big green button! :sunglasses:

PR #234, opened 2015-01-16
Carthage install: Add Nimble to cartfile.private

jmig:
Finally trying out Quick (& Carthage at the same time), great test library (was using Kiwi before).
In the README.md, the carthage installation seems to be missing the Nimble line (used afterwards).

Thanks!


modocache:
Seems reasonable--the other installation instructions all list Quick _and_ Nimble, so we should probably do so here as well. Thanks @jmig! :+1:

jmig:
You're welcome @modocache

PR #237, opened 2015-01-20
Simplify framework search paths to avoid finding the wrong xctest.framework

mattdelves:
I've found that xcodebuild will often select the wrong version of xctest.framework and as such fail to compile Quick.framework.

This pull request removes the inclusion of xctest.framework from Quick-iOS and Quick-OSX. It also simplifies the framework search path to only include $(inherited) $(PLATFORM_DIR)/Developer/Library/Frameworks.

### DIFF
--- a/Quick.xcodeproj/project.pbxproj
+++ b/Quick.xcodeproj/project.pbxproj
@@ -1176,10 +1176,7 @@
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(PLATFORM_DIR)/Developer/Library/Frameworks",
-					"$(SDKROOT)/Developer/Library/Frameworks",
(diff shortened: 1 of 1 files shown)


paulyoung:
Thanks @mattdelves!

I [mentioned](https://github.com/Quick/Quick/pull/201#issuecomment-63849057) something like this before but never got around to it.

Does anyone have any concerns about this change?

/cc @jspahrsummers

jeffh:
I think this also probably worth fixing up the unused frameworks path on Nimble too.

modocache:
I think this change is plainly more correct. So long as someone can confirm that this works with an OS X and iOS project using Quick, we should merge it! :+1: 

To make this process easier in the future, I added #238.

modocache:
This worked for me--thanks, @mattdelves! :cherry_blossom:

PR #240, opened 2015-01-23
Fixes build

ashfurrow:
OK, so first, my sincere apologies for breaking the build on master in the first place. I made [this commit](https://github.com/Quick/Quick/commit/d7d104864e904c097fcd5989c35f81559a5c0187) to master, which *should not* have affected things at all. I have no explanation for why this broke things, but I branched off and reverted it, then repeated the change. This time I did so on my local machine instead of in the GitHub web interface. Maybe there's some sort of encoding error or something.

### DIFF
--- a/Quick.podspec
+++ b/Quick.podspec
@@ -18,5 +18,4 @@ Pod::Spec.new do |s|
   s.source_files  = "Quick", "Quick/**/*.{swift,h,m}"
 
   s.framework = "XCTest"
-  s.requires_arc = true
 end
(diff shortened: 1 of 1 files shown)


ashfurrow:
Suuuuper weird. Since there are effectively no files changed, I'm going to merge this now as to get that red CI status off of the README ASAP. Again, sorry for causing the problem in the first place.

modocache:
Hey @ashfurrow, no worries! Thanks for the updates and the quick fix. :100: 

You're absolutely right that d7d104864e904c097fcd5989c35f81559a5c0187 couldn't have possibly broke the build, so I re-ran the build that failed and, voilà! [It passed](https://travis-ci.org/Quick/Quick/builds/48038617)!

Thanks again for submitting a change that would kick off another build, though. You were absolutely right about not wanting a red CI badge on the README! :+1:

ashfurrow:
Very strange! I did restart the Travis build after the first failure and it failed again, and I had (somehow?) reproduced the locally, even.

PR #242, opened 2015-02-01
Removed beta xcode warning in README.

jeffh:
We might support Xcode versions greater than 6 beta 4 :wink:.

Also worth asking: are the version numbers enough to indicate beta-ness of Quick? If not, I can leave that part of the notice there.


modocache:
No, Quick is pretty stable these days. Thanks!

PR #250, opened 2015-02-21
Add "How to Install Quick using Carthage" to table of contents

mkauppila:
I noticed that the readme was missing a link to "How to Install Quick using Carthage" section from the table of contents, so I added it. Sublime Text also trimmed some unnecessary whitespace from it :convenience_store:

### DIFF
--- a/README.md
+++ b/README.md
@@ -975,7 +975,7 @@ pod 'Quick', :head
 ```
 
 ## How to Install Quick using [Carthage](https://github.com/Carthage/Carthage)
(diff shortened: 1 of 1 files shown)


mkauppila:
hmm, the CI error is interesting :flags: I wonder if it's actually relevant :smile:

modocache:
@mkauppila Yep, sorry--the CI error is due to the fact that Travis CI doesn't support Xcode 6.3/Swift 1.2 (see: https://github.com/Quick/Quick/blob/35e002f2a3bfa590b5030a4416555f9f6bc83a83/CONTRIBUTING.md#pull-requests). But thanks for the PR! :+1:

Issue #140, opened 2014-09-08
Quick does not work on physical devices


jbosse:
Well, the codesign issue of #79 is resolved but it seems to have made room for a new problem.

I am getting the following even when starting with an new iOS project I created to make sure it wasn't something else.

> Ld /Users/jbosse/Library/Developer/Xcode/DerivedData/Sample-bvrevcrvgjvtuigbogtgeilpdfcl/Build/Intermediates/Quick.build/Debug-iphoneos/Quick-iOS.build/Objects-normal/armv7/Quick normal armv7
>     cd /Users/jbosse/Dev/Sample/Vendor/Quick/Quick
>     export IPHONEOS_DEPLOYMENT_TARGET=8.0
>     export PATH="/Applications/Xcode6-Beta7.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin:/Applications/Xcode6-Beta7.app/Contents/Developer/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin"
[...]

PR #252, opened 2015-02-24
Add instructions for running Quick specs on device

mkauppila:
As I mentioned in #140 , adding the frameworks as embedded binaries to the host application is a way to run specs on device. Here’s my suggestion for adding instructions for this to the readme. All sorts of feedback would be appreciated :feet:


modocache:
Pullled this into #254. Thanks, @mkauppila! :+1:

PR #255, opened 2015-02-28
Updates tests in ArrangeActAssert to be correct

ScottPetit:
The tests for offering to peal the banana were testing the wrong cases (testing the peeled message when the banana wasn't peeled and vice versa).


modocache:
Nice catch--thanks, @ScottPetit! :+1:

PR #258, opened 2015-03-08
Bumped Podspec version to 0.3.0

gedward:
Worth noting is on v0.2.2 (what the current podspec is pointing to), in DSL.swift, the func `it` reads:

```public func it(description: String, closure: () -> (), flags: FilterFlags = [:], file: String = __FILE__, line: Int = __LINE__) ```

Which will not allow the syntax in the readme

```it("has everything you need to get started") { }```

to work, as the closure is not yet updated to be at the end of the func declaration. I would expect this to be causing problems with users who are using the cocoapods beta.

### DIFF
--- a/Quick.podspec
+++ b/Quick.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Quick"
-  s.version      = "0.2.2"
+  s.version      = "0.3.0"
   s.summary      = "The Swift (and Objective-C) testing framework."
 
   s.description  = <<-DESC


ashfurrow:
Can you elaborate on the README problem? The closure works as-is, AFAIK, because the other paeans have default values.

ashfurrow:
looks fine to merge, but I'd rather not push it to CocoaPods trunk yet since 0.3 requires an Xcode beta. Devs using that beta can just point to the fit repo (instead of everyone else having to specify a pre-0.3 version).

modocache:
@gedward, are you using Swift 1.2? Xcode 6.3β2 warns when a closure doesn't come after all parameters that take default values, which may explain the errors you're seeing. Use `v0.3.0` when using Xcode 6.3β2.

If the error is occurring on `v0.3.0` when using Xcode 6.3β2, then that's a serious problem we'll need to fix right away! Let us know here or file an issue. :warning:

gedward:
@modocache, I was using `v0.2.2` on Xcode 6.3β2, which is why the issue exists, everything is all good on `v0.3.0`

@ashfurrow, That makes sense, I suppose I may be in a niche of devs on Xcode 6.3β2 using cocoapods to install Quick. This can be closed if the plan is to wait until swift 1.2 is out of beta.

modocache:
@ashfurrow I'd actually love to have v0.3.0 on CocoaPods. I think Nimble already is...? https://github.com/Quick/Nimble/issues/90

ashfurrow:
Oh, hmm. Well, I don't feel that strongly about it. Whatever you think.

modocache:
Sorry it took so long, @gedward, but you can now use Quick v0.3.0 via CocoaPods. :+1:

gedward:
Awesome, thanks!

PR #266, opened 2015-03-20
Update TestingApps.md

dpretty:


modocache:
Haha, thanks @dpretty--nice catch! :fish:

PR #267, opened 2015-03-20
Updated InstallingQuick Documentation for Cocoapods 0.36.0

iv-mexx:
Cocoapods Version 0.36.0 with Swift Support has been released, so the install-instructions for Cocoapods had to be updated.


jeffh:
Thanks @iv-mexx!

Issue #276, opened 2015-04-02
Pick iOS 7 support commits into master branch & release v0.3.1


modocache:
See #206.

jeffh:
Merged and pushed as v0.3.1, closing. Thanks again!

PR #277, opened 2015-04-03
Quick.xcodeproj: Lower iOS deployment target to 7.0

modocache:
Depends on https://github.com/Quick/Nimble/pull/118. Once that's merged, this pull request should be update Externals/Nimble to point at v0.4.2.

Quick and Nimble can be used to test code running against on iOS 7.0, but their deployment targets were set too high in order to do so. Lower Quick's deployment target to make this possible.

Unfortunately, a warning is generated during the linking phase for Quick-iOS: "ld: warning: embedded dylibs/frameworks only run on iOS 8 or later". I'm not sure if this is avoidable. I'm no expert, though, I just skimmed ReactiveCocoa/ReactiveCocoa#1480.

A new release should be made after merging these changes.

Addresses #276.

### DIFF
--- a/Quick.podspec
+++ b/Quick.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Quick"
-  s.version      = "0.3.0"
+  s.version      = "0.3.1"
   s.summary      = "The Swift (and Objective-C) testing framework."
 
   s.description  = <<-DESC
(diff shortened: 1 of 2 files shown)


ashfurrow:
Cooool.

jeffh:
Thanks @marciok and @modocache. I just pushed v0.3.1 to trunk.

modocache:
Awesome, thanks! Great way to start a Monday. :+1:

PR #285, opened 2015-04-19
Contributing guide clarifications

agelber:
Hey there!

These are 2 really small things I ran into when working on something else.

- The first is that the command for running ios and osx tests is just `rake` rather than `rake test`. If you prefer, I can change the Rakefile so that it'll be `rake test` instead of changing the docs.
- The second is just a comment about updating the submodule just so other people forking and cloning for the first time will know what to do if they run into that issue.

Thank you very much! :rocket:


jeffh:
Thanks @agelber!

PR #287, opened 2015-05-01
Create QuickCore.framework

modocache:
**See individual commit messages for details on each change.**

One of Quick's greatest design flaws is due to the combination of four factors:

1. Quick has a predilection for Swift--one of it's goals was to be written entirely in Swift if possible.
2. Quick supports tests written in both Swift and Objective-C.
3. XCTest relies upon subclassing for test discovery.
4. Objective-C classes cannot subclass Swift base classes.

As such, Quick is unable to define QuickSpec, a base XCTestCase subclass, in Swift, and then have both Objective-C and Swift tests subclass QuickSpec.

Instead, in order to be used from both Swift and Objective-C, QuickSpec must be implemented in Objective-C. This forces internal Quick code, such as Quick.World and Quick.ExampleGroup, to be marked as public. People using Quick to write tests were thus able to call Quick.World directly, even though that class is meant to be an implementation detail.

In order to mitigate this architectural flaw, create a new QuickCore.framework. Quick.framework has a dependency on QuickCore, but end-users are not exposed to the code in that framework unless they explicitly import it.

I'm hoping that in a later commit, Quick.framework will be split up into two user-facing frameworks:

1. Quick.framework, for clients writing tests in Swift.
2. Objective-Quick.framework, for clients writing tests in Objective-C.

Both frameworks will define an XCTestCase subclass to be used in each language, and will internally use QuickCore.framework.

---

This pull request introduces breaking changes: in order to use `QuickConfiguration` (and by extension, the shared examples feature), they'll have to import *both* Quick and QuickCore. This is because `Configuration` is defined in QuickCore. I'd really appreciate suggestions on how to avoid this breaking change.


paulyoung:
Wow :raised_hands:

Will take a look over this at some point soon!

dimsumthinking:
I like this idea of splitting into two frameworks -- but it may be because I'm moving to Swift development exclusively and am interested in a framework that is tailored to my needs.

paulyoung:
> I'd really appreciate suggestions on how to avoid this breaking change.

I haven't give this too much thought or looked into the implementation yet, but maybe you can use a forward declaration of `QuickConfiguration`.

modocache:
> I like this idea of splitting into two frameworks -- but it may be because I'm moving to Swift development exclusively and am interested in a framework that is tailored to my needs.

@dimsumthinking, you can see a discussion exclusive to that issue in #288. It's already been closed, but feel free to chime in!

> ...maybe you can use a forward declaration of `QuickConfiguration`.

No dice, since users actually need to interact with that class. From [Documentation/ConfiguringQuick.md](https://github.com/Quick/Quick/blob/0931946b96e9ea82ef6bb7b6f427732c78f0ef97/Documentation/ConfiguringQuick.md):

```swift
class FinConfiguration: QuickConfiguration {
  override class func configure(configuration: Configuration) {
    configuration.beforeEach {
      Dorsal.sharedFin().height = 0
    }
  }
}
```

jeffh:
But you can forward declare Quick.World in objective-c for internal use perhaps?




—

modocache:
The merge conflict is due to changes to the Xcode project file--should be easy to fix.

> But you can forward declare `Quick.World` in Objective-C for internal use perhaps?

Hmm... I'm not exactly sure what you mean. :sweat_smile:

jeffh:
Sorry, I was unclear. I was thinking kind of like `NMBWait` in this commit https://github.com/Quick/Nimble/commit/fcdc7641e141c5184c9ff90c92c5f97e8efb8fd5. It still might be kind of icky.

modocache:
> But you can forward declare `Quick.World` in Objective-C for internal use perhaps?

@jeffh Just making sure--are you suggesting forward declaring `Quick.World` instead of splitting up Quick and QuickCore? I guess that's possible but I'm not a fan of using `SWIFT_CLASS()`...

jeffh:
Yeah. It's not ideal, but it does allow you to keep it private. It's up to you though.

modocache:
> But you can forward declare `Quick.World` in Objective-C for internal use perhaps?

@jeffh Actually, this probably *is* the way to go here--thanks so much for the direction! Splitting up the frameworks sounded like a great idea, but adds a dependency for end users, who have to add both `Quick.xcodeproj` and `QuickCore.xcodeproj` to their workspace--which is just too much of a hassle.

I'll close this pull request in favor of another one that forward-declares `Quick.World`, `Quick.Example`, etc. :point_up:

jeffh:
:ok_hand: :+1:

Issue #231, opened 2015-01-07
Simpler style for declaring shared variables


listrophy:
Apple's builtin XCTest framework has a much simpler way to instantiate shared variables. This mechanism does not work with Quick.

Given the simple class:

``` swift
class Mutator {
    var mutated: Bool
    init() {
        mutated = false
    }

    func mutate() {
        mutated = true
    }
}
```

I'd like to see a way to do this:

``` swift
// Please note that this style does not work, but I'd like it to!

class MyQuickTest: QuickSpec {
    override func spec() {

        describe("Mutator") {
            let subject = Mutator()

            it("test 1") {
                expect(subject.mutated).to(beFalse())
                subject.mutate()
                expect(subject.mutated).to(beTrue())
            }

            it("test 2") {
                expect(subject.mutated).to(beFalse()) // currently fails because `subject` is same mutated object from "test 1"
                subject.mutate()
                expect(subject.mutated).to(beTrue())
            }
        }
    }
}
```

Apple does this in XCTest by instantiating a new object for each test method.

Currently, I'd have to do this, which seems overly verbose and has those gross `!` optional-smashers everywhere:

``` swift
// undesirables:
//   1. must declare shared variable as "var" and not "let".
//   2. must use `beforeEach` to instantiate before each test run
//   3. must use "!" (or worse `if let`) to access underlying value since it's now an optional

class MyQuickTest: QuickSpec {
    override func spec() {

        describe("Mutator") {
            var subject: Mutator?
            beforeEach {
                subject = Mutator()
            }

            it("test 1") {
                expect(subject!.mutated).to(beFalse())
                subject!.mutate()
                expect(subject!.mutated).to(beTrue())
            }

            it("test 2") {
                expect(subject!.mutated).to(beFalse())
                subject!.mutate()
                expect(subject!.mutated).to(beTrue())
            }
        }
    }
}
```

modocache:
Will it, though? As @listrophy points out, a fresh instsance of `XCTestCase` is used every time `-[XCTestCase invokeTest]` is called. Since `QuickSpec` is a humble `XCTestCase` subclass, it's possible this might work. :bulb: 

Gonna have to actually try it locally, though! :sweat_smile:

paulyoung:
I had to write is as follows:
```swift
class MyQuickTest: QuickSpec {
    let subject = Mutator()
    
    override func spec() {
        describe("Mutator") {
            it("test 1") {
                expect(self.subject.mutated).to(beFalse())
[...]

PR #290, opened 2015-05-03
[WIP] Variable definitions

agelber:
Hi there!

__Background:__
I really like the way `let`s work in rspec, and also saw issue #231, so I tried implementing something similar for quick.

__Implementation:__
I added a `Definitions` class which holds a dictionary which maps "variable names" as strings to closures and a dictionary which holds memoized values of those closures.
I added a `definitions` property to `Configuration` which can also be used through `World`, in a similar fashion to the way hooks are used.
So now from anywhere in your tests, you can use `define` (`let` was taken :smile_cat:) which takes a name and a closure to define a variable and `fetch` which takes a variable name and returns the value for that variable. The values are memoized for example and reset between examples (just like rspec).

__My Problem:__
In swift, this works just as expected, the tests I wrote are passing and everything is happy.
In Objective-C, however, my last test (which is identical to the swift one) is failing. Instead of `number` being `4`, I get `2` again. My assumption is that the value of `mutatingNumber` is being captured in the block even though I marked it as `__block`. I am not completely sure why and I couldn't get this to work, so I though one of you guys might have a solution to this.

I really appreciate the work you've all been putting into Quick and would really love to see this feature making it's way in if we can figure this out.
Let me know if there's any more information I could give or any other thing I can do!

Thanks! :beers:

### DIFF
--- a/Quick/Example.swift
+++ b/Quick/Example.swift
@@ -71,6 +71,8 @@ private var numberOfExamplesRun = 0
             after(exampleMetadata: exampleMetadata)
         }
         world.exampleHooks.executeAfters(exampleMetadata)
+        
+        group!.clearMemoizedValues()
 
         ++numberOfExamplesRun
 
--- a/Quick/Example.swift
+++ b/Quick/Example.swift
@@ -72,7 +72,7 @@ private var numberOfExamplesRun = 0
         }
         world.exampleHooks.executeAfters(exampleMetadata)
         
-        group!.clearMemoizedValues()
+        world.definitions.clearMemoizedValues()
 
         ++numberOfExamplesRun
 
--- a/QuickTests/FunctionalTests/DefinitionsTests+ObjC.m
+++ b/QuickTests/FunctionalTests/DefinitionsTests+ObjC.m
@@ -1,0 +1,61 @@
+#import <XCTest/XCTest.h>
+#import <Quick/Quick.h>
+#import <Nimble/Nimble.h>
+
+#import "QCKSpecRunner.h"
+
+static NSMutableArray *fetches;
+
+QuickSpecBegin(FunctionalTests_DefinitionsSpec)
+
+define(@"one", ^NSObject *{ return @1; });
+
+it(@"should evaluate and return the defined variable", ^{
+    NSNumber *one = (NSNumber *)fetch(@"one");
+    [fetches addObject:one];
+    expect(one).to(equal(@1));
+});
+
+__block NSNumber *mutatingNumber = @2;
+define(@"number", ^NSObject *{ return mutatingNumber; });
+
+it(@"should memoize the value of the first evaluation in each example", ^{
+    mutatingNumber = @2;
+    NSNumber *firstFetch = (NSNumber *)fetch(@"number");
+    [fetches addObject:firstFetch];
+    mutatingNumber = @3;
+    NSNumber *secondFetch = (NSNumber *)fetch(@"number");
+    [fetches addObject:secondFetch];
+    expect(secondFetch).to(equal(firstFetch));
+});
+
+it(@"should clear memoized values between examples", ^{
+    mutatingNumber = @4;
+    NSNumber *number = (NSNumber *)fetch(@"number");
+    [fetches addObject:number];
+    expect(number).to(equal(@4));
+});
+
+QuickSpecEnd
+
+@interface DefinitionsTests : XCTestCase; @end
+
+@implementation DefinitionsTests
+
+- (void)setUp {
+    [super setUp];
+    fetches = [NSMutableArray array];
+}
+
+- (void)tearDown {
+    fetches = [NSMutableArray array];
+    [super tearDown];
+}
+
+- (void)testFetchesReturnsTheExpectedValues {
+    qck_runSpec([FunctionalTests_DefinitionsSpec class]);
+    NSArray *expectedFetches = @[@1, @2, @2, @4];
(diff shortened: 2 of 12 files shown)


QuickTests/FunctionalTests/DefinitionsTests.swift
```diff
@@ -0,0 +1,54 @@
+import XCTest
+import Quick
+import Nimble
+
+private var fetches = [Int]()
+
+class FunctionalTests_DefinitionsSpec: QuickSpec {
+    override func spec() {
+        define("one") { return 1 }
```
modocache: Interesting. You could probably make this even shorter:

```swift
define("one") { 1 }
```

I bet you could also have `define` take an `@autoclosure(escaping)` parameter, which would allow you to write:

```swift
define("one", 1)
```

QuickTests/FunctionalTests/DefinitionsTests.swift
```diff
@@ -0,0 +1,54 @@
...
+
+        it("should evaluate and return the defined variable") {
+            let one = fetch("one") as! Int
```
modocache: Hmm... sorry to be negative, but I have doubts as to whether this addresses #231 in a satisfactory way. That issue criticizes how Quick recommends test setup that uses `!`. I doubt the author of that issue will appreciate the kind of forced casting you have here, which not only uses `!` but also performs a cast from `AnyObject`.

This approach works *around* the type system, instead of *leveraging* it, so it doesn't seem like "idiomatic" Swift.

QuickTests/FunctionalTests/DefinitionsTests+ObjC.m
```diff
@@ -0,0 +1,60 @@
...
+__block NSNumber *mutatingNumber = @2;
+define(@"number", ^NSObject *{ return mutatingNumber; });
+
+it(@"should memoize the value of the first evaluation in each example", ^{
+    NSNumber *firstFetch = (NSNumber *)fetch(@"number");
+    [fetches addObject:firstFetch];
+    mutatingNumber = @3;
+    NSNumber *secondFetch = (NSNumber *)fetch(@"number");
+    [fetches addObject:secondFetch];
+    expect(secondFetch).to(equal(firstFetch));
```
modocache: Yeah, I think I can see why this test would fail, while the Swift one would pass.

My theory: here, the block captures a pointer to an `NSNumber` object. You then mutate what that pointer points to. In the Swift test, the closure captures the primitive `Int` value, not a pointer.

Quick/Example.swift
```diff
@@ -71,6 +71,8 @@ private var numberOfExamplesRun = 0
             after(exampleMetadata: exampleMetadata)
         }
         world.exampleHooks.executeAfters(exampleMetadata)
+        
+        world.definitions.clearMemoizedValues()
```
modocache: So, I'm leaning towards the opinion that while this shouldn't be included in Quick itself, there's no reason Quick can't provide hooks that would allow you to release a library that made it possible for people to use `define` within their specs.

Perhaps Quick (QuickCore as of #287) can provide a hook here that would allow you to execute code. You could define a `Definitions` singleton and clear its memoized values here.

modocache:
OK, so this pull request blew me away. Thanks a ton for the test coverage and the detailed message explaining what you did. :+1: 

And while overall I think this is an interesting approach, ultimately I think it's missing type safety. The casting between `AnyObject` and the desired type allows for too much room for user error. :rotating_light:

I do think, however, Quick should allow users to customize it's behavior as much as possible. We could allow for `define` to be vended as a third-party Quick extension with a little work. If that sounds interesting, @agelber, I'll create an issue on the Quick repository to track that change.

I think we should look to RSpec as much as possible for guidance here. Some key questions I'd want to research before proceeding here are:

1. How does RSpec implement `let`? Yes, it's Ruby, so we probably won't be able to use it's approach directly, but their implementation would be a great frame of reference.
2. [RSpec 3.0 extracted `its` into a separate RubyGem](https://gist.github.com/myronmarston/4503509). I assume [rspec/rspec-its](https://github.com/rspec/rspec-its) also needs to integrate with RSpec. How does RSpec 3.0 allow external gems to be integrated with how it runs specs?

QuickTests/FunctionalTests/DefinitionsTests.swift
agelber: :+1:

agelber:
Yeah, I see what you mean about the type safety issue, I didn't quite think about that, coming from ruby.
Another approach I had was to have `define` methods for types, so for example, `defineInt`, `defineString` and so on, but that would still leave you with the issue for most use-cases.
The other option would be to use generics. Is it possible to have method-level generics as opposed to class-level? i.e. `define<T>(name: String, closure: () -> T` and `fetch<T>() -> T`?

Either way, a plugin system sounds like an awesome idea. I'd be happy to help with anything you need. Feel free to contact me via email as well.

jeffh:
You can have generic methods and functions, but you'll need to link the generic type from its definition to its usage in a way the Swift compiler can track and infer its type.




—

agelber:
I see.. I'll try to think of other ways, I guess. Would be glad to get some pointers or suggestions :beers:

modocache:
Thanks again for the great code and discussion, @agelber! :bouquet: 

To keep the pull request tab clean--I'd like only requests in active review to stay there--I'm going to close this pull request in favor of two issues:

- #394 to track enabling integrations that would allow you to write this feature without adding it to Quick directly
- #231 to track implementing a simpler syntax for defining shared variables

PR #292, opened 2015-05-05
Markdown syntax fix in link to Carthage.private docs

mokagio:
This will make the link to Carthage.private look like this [`Carthage.private`](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#cartfileprivate). Most likely a typo while writing the documentation.

Documentation which is great btw, very informative and thorough. I especially like how it first focuses on teaching the bases of good tests, then _sells_ Quick. Very well done. Thanks!


jeffh:
Thanks for the PR @mokagio and the warm compliments :smile:.

PR #295, opened 2015-05-15
Shared examples with different context

hffmnn:
When calling `itBehavesLike` multiple times from within the same ObjC spec, the `sharedExamples` gets always called with the same context. Problem is, that the dynamically generated selector for the instance of the `Example` is always the same same. see c5675e53f55ab05a0e861556ad03f012b9ce4d4a
This PR tries to fix this by adding more unique properties (filename and line number) to the generated selector name. see 563f61570ce3b3cfce7fa5de6e566b5d6c767913

### DIFF
--- a/Quick/QuickSpec.m
+++ b/Quick/QuickSpec.m
@@ -101,7 +101,8 @@ const void * const QCKExampleKey = &QCKExampleKey;
         [example run];
     });
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
-    SEL selector = NSSelectorFromString(example.name.qck_selectorName);
+    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld", example.name.qck_selectorName, example.callsite.file, (long)example.callsite.line];
+    SEL selector = NSSelectorFromString(selectorName);
     class_addMethod(self, selector, implementation, types);
 
     return selector;
--- a/Quick/QuickSpec.m
+++ b/Quick/QuickSpec.m
@@ -101,7 +101,10 @@ const void * const QCKExampleKey = &QCKExampleKey;
         [example run];
     });
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
-    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld", example.name.qck_selectorName, example.callsite.file, (long)example.callsite.line];
+    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld",
+                              example.name.qck_selectorName,
+                              example.callsite.file,
+                              (long)example.callsite.line];
     SEL selector = NSSelectorFromString(selectorName);
     class_addMethod(self, selector, implementation, types);
 
--- a/QuickTests/FunctionalTests/SharedExamplesTests+ObjC.m
+++ b/QuickTests/FunctionalTests/SharedExamplesTests+ObjC.m
@@ -20,7 +20,7 @@ QuickSpecEnd
 
 QuickSpecBegin(FunctionalTests_SharedExamples_SameContextSpec)
 
-__block NSInteger counter;
+__block NSInteger counter = 0;
 
 afterEach(^{
     counter++;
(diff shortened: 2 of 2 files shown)


Quick/QuickSpec.m
```diff
@@ -101,7 +101,11 @@ + (SEL)addInstanceMethodForExample:(Example *)example {
         [example run];
     });
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
-    SEL selector = NSSelectorFromString(example.name.qck_selectorName);
+    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld",
+                              example.name.qck_selectorName,
+                              example.callsite.file,
+                              (long)example.callsite.line];
+    SEL selector = NSSelectorFromString(selectorName);
```
modocache: I'm curious: does this effect the test names that are output in the logs, and shown in the test navigator in Xcode?

QuickTests/FunctionalTests/SharedExamplesTests+ObjC.m
```diff
@@ -18,6 +18,39 @@
...
+        BOOL expected = [payload isEqualToString:[NSString stringWithFormat:@"%ld", (long)counter]];
+        expect(@(expected)).to(beTrue());
+    });
+    
+});
+
+itBehavesLike(@"gets called with a different context from within the same spec file", ^{
+    return @{
+             @"payload" : @"0"
+             };
```
modocache: nitpick: I'd put this dictionary literal on a single line. :expressionless:

QuickTests/FunctionalTests/SharedExamplesTests+ObjC.m
```diff
@@ -18,6 +18,39 @@
...
+});
+
+itBehavesLike(@"gets called with a different context from within the same spec file", ^{
+    return @{
+             @"payload" : @"1"
+             };
```
modocache: Same as above, this literal on a single line maybe? :point_up:

modocache:
This is awesome, thanks a lot for this pull request! :+1: 

I had one question, otherwise this looks great!

modocache:
@jeffh Do you have any insight into the Circle CI failure? If not no worries, I'm planning on learning more about how Circle CI works in the upcoming few weeks.

Quick/QuickSpec.m
hffmnn: New names are even worse, because the complete path is used:
`[Quick_OSXTests.FunctionalTests_AfterEachSpec executes_the_outer_afterEach_closures_once__but_not_before_this_closure__1__/<Path_to_your_project>/Quick/QuickTests/FunctionalTests/AfterEachTests.swift_22]`.

These get logged and are shown in the test navigator (but mostly only for a second, then Xcode shows something else... maybe another bug?).

What do you think would be a appropriate naming scheme? Another option would be, to check for `.isSharedExample == YES` to only mangle shared examples.

jeffh:
I think the circle ci failures were just an out of date version of nimble. Master should have this fix.

Quick/QuickSpec.m
jeffh: A better naming scheme is probably to strip the invalid characters for an identifier or replace them with an `_`. Historically, Xcode's test browser only accepted function-styled identifier names.

hffmnn: My question is: Isn't the new selector name much to long? Shouldn't we e.g. strip the path and only use the filename?

modocache: @hffmnn Yes, I would agree that the new names are too long--I prefer the way they are now...

Is there another way we can fix #299, besides changing the test selector names? If this really is the only way, couldn't we *just* use the line number, since this bug only manifests within the same file?

hffmnn: @modocache Regarding
> Is there another way we can fix #299, besides changing the test selector names?

giphy

So: Definitely Maybe! :)

If this is the only way (or the way we choose), we could use the added line number just for sharedExamples (via testing for `isSharedExample` on `Example`).

hffmnn:
@modocache Why is this still labeled waiting for contributor? I already answered: https://github.com/Quick/Quick/pull/295#discussion_r31057960

jeffh:
Thanks @hffmnn, sorry for the slow turnaround.

PR #303, opened 2015-05-23
Remove Failure.swift

modocache:
The 'Failure' class was no longer being used within the framework
(probably) as of 6a68691, and is not an object that Quick
consumers should rely on.


jeffh:
Merged. I'll see how CircleCI builds this before trying to merge the other PRs.

PR #302, opened 2015-05-23
Bridge example metadata to Objective-C

modocache:
A goal for Quick v1.0 is to no longer expose the `Quick.World` singleton. This is made harder by the fact that our very own tests use it!

Bridge `beforeEach` and `afterEach` closures that take example metadata arguments to Objective-C, in order to allow `ItTests+ObjC.m` to not use "private" APIs.

PR #304, opened 2015-05-23
Forward-declare World, which is now internal

modocache:
This must be merged **after** #302. This pull request is a revision of #287.

---

One of Quick's greatest design flaws is due to the combination
of four factors:

1. Quick has a predilection for Swift--one of it's goals was to be
   written entirely in Swift if possible.
2. Quick supports tests written in both Swift and Objective-C.
3. XCTest relies upon subclassing for test discovery.
4. Objective-C classes cannot subclass Swift base classes.

As such, Quick is unable to define QuickSpec, a base XCTestCase
subclass, in Swift, and then have both Objective-C and Swift tests
subclass QuickSpec.

Instead, in order to be used from both Swift and Objective-C, QuickSpec
must be implemented in Objective-C. This forces internal Quick code,
such as Quick.World, to be marked as public. People using Quick to
write tests were thus able to call Quick.World directly, even though
that class is meant to be an implementation detail.

In order to get around the need for World to be public, define macros
which can be used to make a forward declaration of World within an
Objective-C file.

### DIFF
--- a/Quick/Configuration/QuickConfiguration.m
+++ b/Quick/Configuration/QuickConfiguration.m
@@ -1,6 +1,9 @@
 #import "QuickConfiguration.h"
+#import "World.h"
 #import <Quick/Quick-Swift.h>
 #import <objc/runtime.h>
+
+QuickWorldBridge
 
 typedef void (^QCKClassEnumerationBlock)(Class klass);
 
--- a/Quick/Configuration/QuickConfiguration.m
+++ b/Quick/Configuration/QuickConfiguration.m
@@ -1,9 +1,6 @@
 #import "QuickConfiguration.h"
 #import "World.h"
-#import <Quick/Quick-Swift.h>
 #import <objc/runtime.h>
-
-QuickWorldBridge
 
 typedef void (^QCKClassEnumerationBlock)(Class klass);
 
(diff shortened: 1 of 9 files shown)


Quick/Configuration/QuickConfiguration.m
```diff
@@ -1,7 +1,10 @@
 #import "QuickConfiguration.h"
+#import "World.h"
 #import <Quick/Quick-Swift.h>
 #import <objc/runtime.h>
 
+QuickWorldBridge
```
jeffh: Is there a reason behind using a macro has opposed to just having `World.h` import the forward declarations?

modocache: Placing the forward declarations in header files causes compilation errors--they only compile in implementation files. The macro allows for that. I might be doing something wrong here... :fearful:

jeffh: You just need to add `#import <Quick/Quick-Swift.h>` inside `World.h` to get the `SWIFT_CLASS` macro.

jeffh:
Looks good :+1:

I removed the macro + rebased to current master (and forced pushed the branch). Feel free to merge when this goes green.

modocache:
Wow, thanks for the help pulling this past the finish line, @jeffh! I owe you one! :bow: :sparkles:

paulyoung:
:mega: :earth_americas:

Issue #311, opened 2015-06-08
Swift 2.0 Support: master, or 2.0 branch?


modocache:
A continuation of the discussion on https://github.com/Quick/Nimble/pull/131. Should we:

1. Support Swift 2.0 and the latest Xcode beta on the `master` branch. Create a branch named `swift-1.2` which will support Swift 1.2 and the latest stable release of Xcode.
2. Support Swift 1.2 and the latest stable release of Xcode on the `master` branch. Create a branch named `swift-2.0` which will support Swift 2.0 and the latest Xcode beta.

Comment with your preference, or any other feedback!

esttorhe:
I'd say move `1.2` to its own branch and keep `master` with `2.0`.

Honestly it makes more sense because `2.0` is going to be the «latest official» `Swift` and makes more sense to keep `master` up to date.

IMHO ¯\_(ツ)_/¯

aarondaub:
2.0 should be a branch until it is out of beta

MarcoSero:
I lean towards n.2 but setting the default Github branch to `swift-2.0`, the way ReactiveCocoa has done for their v3.

esttorhe:
Yeah; I guess that's what I meant to move it to `master` (when is out of «beta»).

dimsumthinking:
Agreed

twobitlabs:
+1 for branch while its in beta.

fabiomassimo:
+1 for "branch it until it's beta"

modocache:
@Quick/contributors?

ashfurrow:
I like number 2 – keep it on a branch until it's out.

paulyoung:
Part of me is all "YOLO" but generally I like to keep master able to deploy the latest stable version of a project.

modocache:
Sounds good--thanks for the input, everybody! We'll be maintaining Swift 2.0-compatible releases on branches named `swift-2.0`. As usual, **don't** depend on a specific branch of Quick or Nimble in your project! Instead, use our releases (which we will tag).

For Quick, Swift 1.2-compatible releases will continue to be tagged `v3.*`. Swift 2.0-compatible releases will be tagged `v4.*`. We're still deciding version numbers for Nimble--follow the discussion in https://github.com/Quick/Nimble/pull/131 and https://github.com/Quick/Nimble/pull/132.

PR #312, opened 2015-06-08
Support Swift 2.0

modocache:
1. Update .gitmodules to pull in Nimble Swift 2.0 support
   (https://github.com/Quick/Nimble/pull/131).
2. Run Edit > Convert > To Latest Swift Syntax.
3. XCTestObservationCenter is now a public XCTest API, although it
   doesn't expose "suspend observation". Expose that method via a category.

---

We should get a consensus on #311 before merging this, but would love to get some early feedback from @Quick/contributors! :+1:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
+*.xcscmblueprint
 
 # CocoaPods
 #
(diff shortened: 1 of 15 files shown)


modocache:
Closing in favor of #313.

(Boy, it sure would be nice if GitHub allowed you to edit the target of a pull request.)

PR #321, opened 2015-06-16
Updated Cocoapods instructions

NatashaTheRobot:
Updated instructions for Cocoapods using multiple testing targets


modocache:
Thanks, @NatashaTheRobot! The README already contains a brief blurb on CocoaPods--shouldn't these additional notes *replace* the old ones?

It'd also be great if you could update the documentation in https://github.com/Quick/Quick/blob/swift-2.0/Documentation/InstallingQuick.md. :wink:

modocache:
Awesome, thanks again, @NatashaTheRobot! :bouquet:

PR #328, opened 2015-06-21
Support XCTestAssert macros in Quick specs written in Objective-C

### LINKED ISSUE #121: XCTestAssertions do not work within a QuickSpec written in Objective-C
When using XCTest assertions or 3rd Party libraries(OCMock/Mockito), failures are not handled properly on QuickSpec.
the proper method is getting called 
````
- (void)recordFailureWithDescription:(NSString *)description inFile:(NSString *)filePath atLine:(NSUInteger)lineNumber expected:(BOOL)expected;
````
but Xcode still doesn't show test as failed.

Running through debugger I noticed that when running through nimble expectation example is set correctly, but if run from any other lib or XCTest we end up with empty example, might be something related to that?

To reproduce you could just add 
````
XCTFail(@"Fail!");
````

modocache:
Fixes #121. See individual commit messages for details.

---

I wanted to make sure `XCTAssert` macros also worked within shared examples, but that brought up a separate issue: since shared examples are now typically defined within `+[QuickConfiguration configure:]`, a class method, `XCTAssert` macros raise a warning:

> Incompatible pointer types passing 'Class' to parameter of type 'XCTestCase'

We may have to change the configuration API in order to support `XCTAssert` macros in shared examples... I'll file a separate issue for this, unless any reviewers think it's not an issue that should be considered separately from this PR.


jeffh:
Thanks @modocache :+1:.

I did have some vague ideas for removing the `currentSpec` var since I remember doing something in cedar to preserving the original spec instance. But I probably won't have free time (for a while) for that.

PR #329, opened 2015-06-21
Add QUICK_EXPORT

### LINKED ISSUE #310: Quick functions should extern properly for Objective-C++
For example:

```objective-c
extern void qck_beforeSuite(QCKDSLEmptyBlock closure);
```

This prototype definition is incorrect in (Objective-)C++ because of name mangling rules.

Instead,  we should replace the `extern` keyword with a macro like Foundation's

```objective-c
#if defined(__cplusplus)
#define FOUNDATION_EXTERN extern "C"
#else
#define FOUNDATION_EXTERN extern
#endif
```

Or just use `FOUNDATION_EXTERN` instead.

marciok:
Fix #310

### DIFF
--- a/QuickTests/Helpers/QCKSpecRunner.h
+++ b/QuickTests/Helpers/QCKSpecRunner.h
@@ -10,7 +10,7 @@
  @param specClass The class of the spec to be run.
  @return An XCTestRun instance that contains information such as the number of failures, etc.
  */
-FOUNDATION_EXTERN XCTestRun *qck_runSpec(Class specClass);
+QUICK_EXPORT XCTestRun *qck_runSpec(Class specClass);
 
 /**
  Runs an XCTestSuite instance containing the given XCTestCase subclasses, in the order provided.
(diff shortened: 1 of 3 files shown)


modocache:
Awesome, thanks @marciok! Can we do something like [this](https://github.com/Quick/Nimble/blob/swift-2.0/Nimble/objc/DSL.h#L9)? That is, define a `QUICK_EXPORT`?

marciok:
Ops..  @modocache you are right, it makes more sense

QuickTests/Helpers/QCKSpecRunner.h
```diff
@@ -10,7 +11,7 @@
  @param specClass The class of the spec to be run.
  @return An XCTestRun instance that contains information such as the number of failures, etc.
  */
-extern XCTestRun *qck_runSpec(Class specClass);
+QUICK_EXPORT XCTestRun *qck_runSpec(Class specClass);
```
modocache: This is a nice touch, but the original problem was with using the Quick DSL in Objective-C++ files. This test helper isn't a part of the Quick DSL, so it might be outside of the scope of this pull request.

If we wanted to make sure this test helper could be used from within Quick's unit tests, even if those were written in Objective-C++ (they're all Swift or Objective-C at the moment), then I would argue that we should define a `QUICK_TESTS_EXPORT` macro. I don't think this file should have a dependency on `Quick/QCKDSL.h`.

marciok: Ops(again)...  Yes, you are right, it was lack of attention, my bad. I'm going to remove.

modocache:
Awesome, thanks @marciok! :+1:

marciok:
@modocache  🎉

PR #331, opened 2015-06-22
QuickSpec.current()

### LINKED ISSUE #330: XCTestExpectation fulfill() in Quick test causes NSInternalInconsistencyException
This code:

```
class QuickExpectationBug: QuickSpec {
    override func spec() {
        it("handles expectations correctly") {
            let expectation = self.expectationWithDescription("simple expectation")
            expectation.fulfill()
            self.waitForExpectationsWithTimeout(0.1, handler: nil)
        }
    }
}
```

…causes this exception:

```
Test Case '-[SiestaTests.QuickExpectationBug handles_expectations_correctly]' started.
2015-06-21 17:08:31.230 xctest[2044:34358] *** Assertion failure in -[XCTestExpectation fulfill], /Library/Caches/com.apple.xbs/Sources/XCTest_Sim/XCTest-8102.19/XCTestFramework/Classes/XCTestCase+AsynchronousTesting.m:399
/Users/paul/work/bustout/siesta/SiestaTests/QuickExpectationBug.swift:15: error: -[SiestaTests.QuickExpectationBug handles_expectations_correctly] : failed: caught "NSInternalInconsistencyException",
  "API violation - called -[XCTestExpectation fulfill] for simple expectation after
  the test case in which the expectation was created is no longer running.
  Created in test -[QuickExpectationBug (null)],
  current test is -[QuickExpectationBug handles_expectations_correctly]"
```

The same code works in an XCTestCase:

```
class ExpectationTest: XCTestCase {
    func testExpectation() {
        let expectation = self.expectationWithDescription("simple expectation")
        expectation.fulfill()
        self.waitForExpectationsWithTimeout(0.1, handler: nil)
    }
}
```

Something to do with Quick’s dynamically created test cases?

Using XC7.0 beta (7A120f), swift-2.0 branch of Quick.

pcantrell:
One possible way to fix #330: examples use QuickSpec.current() instead of self to access XCTestCase instance methods. For example:

        it("handles expectations") {
            let expectation = QuickSpec.current().expectationWithDescription("simple expectation")
            expectation.fulfill()
            QuickSpec.current().waitForExpectationsWithTimeout(0.1, handler: nil)
        }

(This is option 3 from my comments on that bug. Option 2 is possible, but gets messy fast.)

Let me know if this approach is acceptable. If so, I’ll add docs.

### DIFF
--- a/Quick Templates/Quick Configuration Class.xctemplate/Objective-C/___FILEBASENAME___.h
+++ b/Quick Templates/Quick Configuration Class.xctemplate/Objective-C/___FILEBASENAME___.h
@@ -1,4 +1,4 @@
-#import <Quick/Quick.h>
+@import Quick;
 
 @interface ___FILEBASENAMEASIDENTIFIER___ : QuickConfiguration
 
--- a/Quick Templates/Quick Configuration Class.xctemplate/Objective-C/___FILEBASENAME___.m
+++ b/Quick Templates/Quick Configuration Class.xctemplate/Objective-C/___FILEBASENAME___.m
@@ -1,4 +1,3 @@
-#import <Quick/Quick-Swift.h>
 #import "___FILEBASENAMEASIDENTIFIER___.h"
 
 @implementation ___FILEBASENAMEASIDENTIFIER___


jeffh:
I've commented on the original issue, since it has a lot more context. But in short, don't merge yet.

PR #335, opened 2015-06-24
[Podspec] Restore support for OS X 10.7

kylef:
Talked to @modocache and the change to make this 10.10 was unintentional.


jeffh:
Thanks @kylef!

PR #336, opened 2015-06-25
Xcode7b2 migration

### LINKED ISSUE #325: Can't install Quick via Carthage under Xcode 7?
I created a brand new Swift project in Xcode 7 and followed the [installation instructions](https://github.com/Quick/Quick/blob/swift-2.0/Documentation/InstallingQuick.md). But when I try and `import Quick` at the top of the default FooTests.swift file, Xcode complains `No such module 'Quick'`.

Has anyone personally had success importing Quick as a framework in Xcode 7 via Carthage?

jeffh:
Ran through Xcode 7 beta 2's Swift Syntax Migrator.

Fixes support for carthage too. Fixes #325.

### DIFF
--- a/Quick/DSL/World+DSL.swift
+++ b/Quick/DSL/World+DSL.swift
@@ -97,6 +97,6 @@ extension World {
     }
 
     internal func pending(description: String, closure: () -> ()) {
-        print("Pending: \(description)", appendNewline: false)
+        print("Pending: \(description)")
     }
 }
--- a/Quick/DSL/World+DSL.swift
+++ b/Quick/DSL/World+DSL.swift
@@ -97,6 +97,6 @@ extension World {
     }
 
     internal func pending(description: String, closure: () -> ()) {
-        print("Pending: \(description)")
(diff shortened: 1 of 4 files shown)


Quick/DSL/World+DSL.swift
```diff
@@ -97,6 +97,6 @@ extension World {
     }
 
     internal func pending(description: String, closure: () -> ()) {
-        print("Pending: \(description)")
+        print("Pending: \(description)", appendNewline: false)
```
modocache: We need to handle pending examples in a nicer way, but for the time being, shouldn't this include a new line? Each pending example should be on its own line in the test output.

modocache:
Looks good to me, just had the one comment around printing pending examples. Feel free to merge!

Quick/DSL/World+DSL.swift
jeffh: Swift 2's print now does newline (true) by default.

jeffh:
Thanks for reviewing, @modocache. :+1:

Issue #83, opened 2014-06-29
Add file template for QuickConfiguration subclass


modocache:
Once shared examples are ported to Objective-C in #80, add a file template for shared example subclasses. Like the spec file template, it should allow users to create a new file using either Swift or Objective-C.

mokagio:
Hello there 👋I'm happy to help with this, expect a PR soon-ish 👍

BTW having the "help wanted" and "easy contribution" is such a great idea to help people get started, well done!

PR #339, opened 2015-06-30
Add template for QuickConfiguration subclass

mokagio:
See #83, nothing else to add, is a pretty simple PR :smile: 

Quick is awesome, thanks guys :clap:


jeffh:
Thanks @mokagio for the PR. :tada: Sorry for the slow turn-around for merging this in.

I made one small change, I used `@import` instead of `#import` so `Quick.h` and `Quick-Swift.h` don't need to be imported manually.

Cheers,
Jeff

PR #342, opened 2015-07-09
Swift and Xcode 7 beta 3 compatibility fix

sync:
XCTest framework doesn't have bitcode enabled, since Xcode beta 3 it's treated as an error, there is probably a better solution out there :-)


pcantrell:
Fixes #343

jeffh:
Thanks @sync!

PR #344, opened 2015-07-11
Fix access value in document

S-Shimotori:
in [Effective Tests Using XCTest: Arrange, Act, and Assert](https://github.com/Quick/Quick/blob/swift-2.0/Documentation/ArrangeActAssert.md)

you should use ```public var isEdible``` instead of ```private var isPeeled``` in Banana/Offer.swift.


jeffh:
Thanks @S-Shimotori!

Issue #299, opened 2015-05-18
[Bug] Shared Example gets called with same context from within the same spec file


hffmnn:
When calling `itBehavesLike` multiple times from within the same ObjC spec, the `sharedExamples` gets always called with the same context. Problem is, that the dynamically generated selector for the instance of the `Example` is always the same same.

https://github.com/Quick/Quick/pull/295 tries to fix this.

modocache:
Thanks a ton for bringing this up, and for the proposed fix! :100:

istx25:
Closing since this was addressed in #295.

Issue #219, opened 2014-12-11
Swift tests do not show up in test case list before being run


jspahrsummers:
Whereas Objective-C specs defined with Quick appear in the sidebar and the Test action (in the scheme editor) after Xcode's indexing has completed, Swift specs do not show up until the tests have run at least once.

codeblooded:
I was able to replicate this issue as well. Deleting derived data stops the tests from disappearing after they are run, but it doesn't cause the tests to be indexed and listed in the sidebar before execution.

> Do Swift tests written using XCTestCase show up in the test navigator before they're run?

In my tests, they do appear when the file is saved or when the project is Cleaned.  The Quick tests do not appear when the project is cleaned, only when it is run.  

> How is it that Objective-C can display these before the tests are run?

I think it is something on Apple's end.  It seems that Apple's indexer for test cases can pick up on XCTestCase(s) and is unable to recognize Quick tests.  Is there a literal crawler in the test projects looking for test cases whenever the file is saved/cleaned? 

I think this won't get any further until a radar is filed and they respond.

modocache:
@codeblooded Awesome, thanks for the info! :+1:

I plan on doing so myself as well, but if you end up filing a radar, please link to the [Open Radar](http://openradar.appspot.com/page/1) here!

Also, recently I've been wondering if maybe we can work around these issues by reverse-engineering XCTest. I've uploaded [the class-dump here](https://github.com/modocache/XCTest-Headers). Just to work around some of these tricky issues, and only sparingly, though.

modocache:
If you use Specta or Quick and are disappointed with tests disappearing in Xcode's Test Navigator, tests lingering in the Test Navigator despite no longer being defined, or even tests not being run, please read and duplicate this radar: https://openradar.appspot.com/26028557.

Xcode does not provide the Specta or Quick maintainers with an API to address these problems. The best way to contribute to fixing these issues is to communicate to the Apple XCTest team that this is a use case we wish they'd support better. The best way I know to do that is to ask you all to re-submit https://openradar.appspot.com/26028557. 🙇

modocache:
A reminder: many Apple engineers only work on issues that have been reported via Radar, and duplicate Radars are used to prioritize what gets worked on: https://twitter.com/yipe/status/776084513633935361

**The best way to contribute to Quick (and Specta) is to file a duplicate of this Radar: https://openradar.appspot.com/26028557**

[...]

PR #345, opened 2015-07-12
Fix same name test function

S-Shimotori:
for #295, #299

add function with class name and line number when you write the same name test.
in my opinion, class name is better than file name because of length and valid characters


jeffh:
Hey @S-Shimotori,

Sorry, I merged #295. In practice, class names are unique but they might be able to overlap (eg - nested class definitions, bundle injection class smashing) so I took the more conservative file name variant.

I do intend to revisit this renaming issue when I attempt to address #219. The solution would probably be more like Specta & Cedar's solution of postfixing numbers to the end of the original generated test method.

Thanks for submitting the PR,

Jeff

PR #348, opened 2015-07-18
Update cocoapods instructions

NatashaTheRobot:
CocoaPods link_with does not do what I thought it does! Updated the README and Installation Instructions with the correct way (thanks [@neonichu](https://gist.github.com/neonichu/16ce4cca9f0b8230f8ba)) of linking the testing pods against both the Unit Test and the new UITests targets in a DRY way.


jeffh:
Thanks @NatashaTheRobot! :+1: 

I merged it after tweaking some whitespace and setting using the final release of Nimble `v1.0.0`.

PR #350, opened 2015-07-22
Swift and Xcode 7 beta 4 compatibility fix

sync:
the new Beta doesn't like having a class marked @objc that isn't a subclass of NSObject

### DIFF
--- a/Quick/Example.swift
+++ b/Quick/Example.swift
@@ -4,7 +4,7 @@ private var numberOfExamplesRun = 0
     Examples, defined with the `it` function, use assertions to
     demonstrate how code should behave. These are like "tests" in XCTest.
 */
-@objc final public class Example: Equatable {
+final public class Example: NSObject {
     /**
         A boolean indicating whether the example is a shared example;
         i.e.: whether it is an example defined with `itBehavesLike`.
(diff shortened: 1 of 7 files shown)


Quick/ExampleGroup.swift
```diff
@@ -3,21 +3,25 @@
...
     internal init(description: String, flags: FilterFlags, isInternalRootExampleGroup: Bool = false) {
-        self.description = description
+        self.internalDescription = description
         self.flags = flags
         self.isInternalRootExampleGroup = isInternalRootExampleGroup
     }
+    
+    public override var description: String {
+        return internalDescription
+    }
```
sync: not happy about having to do this :-1:

Quick/Example.swift
```diff
@@ -20,16 +20,20 @@ private var numberOfExamplesRun = 0
...
-        self.description = description
+        self.internalDescription = description
         self.closure = closure
         self.callsite = callsite
         self.flags = flags
     }
+    
+    public override var description: String {
+        return internalDescription
+    }
```
sync: not happy about having to do this :-1:

Quick/Example.swift
```diff
@@ -20,16 +20,20 @@ private var numberOfExamplesRun = 0
...
 
-    private let description: String
+    private let internalDescription: String
     private let closure: () -> ()
     private let flags: FilterFlags
 
     internal init(description: String, callsite: Callsite, flags: FilterFlags, closure: () -> ()) {
-        self.description = description
+        self.internalDescription = description
         self.closure = closure
```
sync: not happy about having to do this :-1:

jeffh:
Thanks @sync!

I've merged this and reverted the submodule back to `swift-2.0` on Nimble. The `internalDescription` is not ideal, but it works.

@modocache: Do you want the Objective-C classes to be prefixed with `QCK` or leave as is?

modocache:
I think prefixing with `QCK` is a good idea if the classes are available in Objective-C. That is, `class Foo` in Swift, and if bridged to Objective-C, then `@objc(QCKFoo) class Foo`.

Sorry for the late reply, though... :grimacing:

PR #359, opened 2015-08-01
Sanitize filenames before using a part of the test method

shihang0001:


jeffh:
Did I forget to merge that back upstream? My bad, thanks for the catch @shihang0001!

PR #361, opened 2015-08-15
Switch LineNumber argument to an unsigned int

romankl:
Lines are always greater than zero so they can be unsigned - just to match `XCTests`, because they are also using unsigned ints,


modocache:
Seems reasonable--good to maintain consistency with XCTest, in any case. Thanks, @romankl! :+1:

PR #387, opened 2015-09-18
Changes XCode version on Circle CI to be 7.0

pedrovereza:
Circle CI recently announced support for XCode 7.0
https://twitter.com/circleci/status/644589240907489280

It may not be 100% stable yet, but as the swift-2.0 branch is being ignored anyway, I think it's better than what we have now.


jeffh:
Thanks @pedrovereza!

PR #388, opened 2015-09-21
Xcode7/xctool compatibility

dcaunt:
I'm using Quick 0.6.0 to run tests for an Objective-C target and everything works fine within Xcode and `xcodebuild` but not with `xctool`. xctool now supports Xcode 7 but the tests don't run with Quick due to a similar issue issue in Specta that was discussed in https://github.com/facebook/xctool/issues/572 and fixed in https://github.com/specta/specta/pull/184. 

I have to confess I don't fully understand how Quick works but I can confirm that with this fix and by building the Quick framework and adding it to my App's test target, the tests run correctly under xctool. 

Without this change, run-tests fails with a message like this:

>*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: '*** Assertion failure in void ParseClassAndMethodFromTestName(NSString **, NSString **, NSString *), /tmp/xctool20150921-10719-9r0vvv/Common/ParseTestName.m:38: Test name seems to be malformed: -[DCDateValueTransformerSpec (null)]'

### DIFF
--- a/Quick/QuickSpec.m
+++ b/Quick/QuickSpec.m
@@ -45,6 +45,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
                            @"Here's the original exception: '%@', reason: '%@', userInfo: '%@'",
                            exception.name, exception.reason, exception.userInfo];
     }
+    [self testInvocations];
 }
 
 /**


modocache:
I consider https://github.com/facebook/xctool/issues/572 to be a bug in xctool, not frameworks like Specta or Quick. That being said, I think it's better to merge this PR and fix the issue, at least in the short term.

In addition:

1. I'll create an issue to add an xctool-based build step to our CI, in order to ensure we play nice with that particular tool.
2. I'll create an issue on xctool itself to repsect the order in which XCTest builds tests. As far as I can tell, the problem here is that xctool is querying for test method names in an order different from XCTest.

dcaunt:
Much appreciated, and thanks for starting the discussion with xctool.

modocache:
Thank *you*, @dcaunt, for the report and pull request! :+1:

Issue #381, opened 2015-09-10
Support tvOS


brentleyjones:
Like https://github.com/Quick/Nimble/pull/183.

(I don't use Quick, which is why I'm not just adding this myself.)

modocache:
Thanks, @brentleyjones! I'm currently downloading the latest version of Xcode. If you'd like to submit a PR in the meantime (even if you don't use Quick), that'd be much appreciated! :100:

garnett:
I've added the same changes, but still do not see appletv in devices list to verify if it actually works, @brentleyjones how did you manage to run tests on tv as destination?

brentleyjones:
TARGETED_DEVICE_FAMILY and SUPPORTED_PLATFORMS were the key. Also, of course, Xcode 7.1 beta.

garnett:
@brentleyjones I've changed them :cry: Using beta 2 on 10.10

modocache:
@garnett Is #396 ready to go, or does it not work yet? :8ball:

garnett:
@modocache now it's ready! however I had to update Nimble submodule to latest version

modocache:
Fantastic! Merged. Thanks, everyone! :100:

PR #396, opened 2015-09-28
Add tvOS support

garnett:
Implements https://github.com/Quick/Quick/issues/381


modocache:
Excellent! Thanks so much, @garnett! :+1:

PR #400, opened 2015-10-07
add tvOS support for Cocoapods

jondwillis:


jeffh:
Thanks @jondwillis!

reidmweber:
This seems to break for me when I did a pod install. 
```
[!] Unable to load a podspec from `Quick.podspec`, skipping:
```
I was getting the pod from master since there isn't a release that disables bit code. I went back to the previous commit and pod install works again.

PR #408, opened 2015-10-23
Add missing headers to the umbrella header

TomasLinhart:
These headers were removed from the umbrella header so tests in Objective-C stopped working.


jeffh:
Using `<>` brackets seem to break Cocoapods integration

jeffh:
Thanks @TomasLinhart for the PR.

I've pushed a fix for Xcode 7.1 to the `xcode7.1` branch which readds the missing headers as relative imports (which doesn't break cocoapods integration for some reason, I'll need to retroactively investigate it later).

Closing PR. Thanks again for helping out!

TomasLinhart:
Can you give me a commit where you fixed it? I cannot find it because currently `xcode7.1` branch doesn't work with Carthage because these headers are missing.

Issue #389, opened 2015-09-21
Add xctool-based CI step


modocache:
I don't personally use xctool, but some of our users do; see #388. Our CI step should verify that Quick works with the latest version of xctool.

mokagio:
I sarted looking at this, https://github.com/mokagio/Quick/tree/mokagio/test-xctool https://circleci.com/gh/mokagio/Quick.

mokagio:
The work on #417, which is no merged into master, allows to run iOS tests via xctool. Unfortunately there is currently [an xctool issue](https://github.com/mokagio/Quick/commit/e0305a2c0b9ec94fd3ce25c7b69bb7ca90ebc6d8) that prevents running the OS X tests.

Imo this issue should stay open till the OS X tests are running as well.

mokagio:
This issue can now be closed 😉

mokagio:
I think this issue could be closed.

c9767ff975eb4f55d7b9c8725c5b04300849e739 has removed xctool because:

> xctool won't support Xcode 8 and will be deprecated.

I had heard similar news, and later heard that support for Xcode 8 will arrive, although I don't have any link/tweet to reference for it.

Looking at the repo it seems like some Xcode 8 work is on it's way ([this](https://github.com/facebook/xctool/commit/b8bb63a8596399542d10b82e075b5785a40ffd7b) and [this](https://github.com/facebook/xctool/commit/510e366a67a0434c36c8d8e61e467d57d9c153c9) for example).

I do feel that the popularity of xctool has gone down in the past year or so, and don't see much value in having our CI run through it.

istx25:
I agree!

PR #417, opened 2015-10-29
Add xctool tests to CI

mokagio:
See https://github.com/Quick/Quick/issues/389

For some reason my CircleCI setup has decided that I cannot run Xcode tests anymore, so I thought I might as well open the PR anyway and see if the tests will run for the Quick organization.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,9 @@ script:
   - rake test:ios
   - rake test:osx
   - rake test:xctool_ios
-  - rake test:xctool_osx
-
+  # Due to the `pending` specs using a `print` xctool fails generating the report,
+  # which in turn makes the tests fail.
+  # See https://github.com/facebook/xctool/issues/314#issuecomment-152988101 and comments
+  # for details.
+  # See https://circleci.com/gh/mokagio/Quick/12 for a proof that without `print` the test pass.
+  # - rake test:xctool_osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,13 @@
 osx_image: xcode7
 language: objective-c
 
-before_install: git submodule update --init --recursive
-script: "rake"
+before_install:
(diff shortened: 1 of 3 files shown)


mokagio:
Ok so the `xctool` tests for OS X are all passing but the command is failing anyway with some issue on the reporter. That's happening on my machine too but I wanted a double check. I'll look into that :hammer:

mokagio:
This is ready for review, in my opinion.

What I've done:

- Added two Rake tasks to run the iOS and OS X tests using `xctool`.
- Updated `.travis.yml` and `circle.yml` to make sure `xctool` is up to date via `brew` before running the tests.
- Tweaked the `.yml`s to workaround a recent bug with a formula which makes `brew update` fail. See https://github.com/Homebrew/homebrew/issues/45616.
- Disabled, hopefully just temporarly, the `xctool` OS X tests run due an `xctool` bug, which fails the tests due to Quick's `pending` output using a `print`. See https://github.com/facebook/xctool/issues/314#issuecomment-152988101 and comments for more info, and see https://circleci.com/gh/mokagio/Quick/14 for a proof of the `xctool` OS X tests passing if `pending` doesn't print. Imho this is an `xctool` bug, so it is not up to Quick to change to accomodate it.

I'd like feedback on the Rake tasks naming style, see https://github.com/mokagio/Quick/commit/26ee8c19ab69899fa2c40ad152ae021fc3500e44#comments

jeffh:
I'll take a look at it this weekend. Thanks for putting in the work @mokagio

jeffh:
I've merged your changes while namespacing the rake task to be `test:xctool:osx` and `test:xctool:ios`.

Thanks @mokagio for the PR.

mokagio:
I like the namespace choice you @jeffh made :)

Thanks for all the hard work you guys put in this awesome tool 💖

PR #431, opened 2015-12-01
add descriptive error when describe is used out of context

morganchen12:
Add a (hopefully) more informative error message when crashing in `describe` to make [crashes like this one](https://github.com/Quick/Nimble/issues/219) more straightforward. From what I can tell the only time the current example group can be `nil` is if the root group was never instantiated.

Includes some minor code changes that don't affect behavior as well.

### DIFF
--- a/Quick/DSL/World+DSL.swift
+++ b/Quick/DSL/World+DSL.swift
@@ -18,7 +18,7 @@ extension World {
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
         let group = ExampleGroup(description: description, flags: flags)
-        currentExampleGroup!.appendExampleGroup(group)
+        currentExampleGroup.appendExampleGroup(group)
         currentExampleGroup = group
         closure()
         currentExampleGroup = group.parent
@@ -41,28 +41,28 @@ extension World {
     }
 
     internal func beforeEach(closure: BeforeExampleClosure) {
-        currentExampleGroup!.hooks.appendBefore(closure)
+        currentExampleGroup.hooks.appendBefore(closure)
     }
 
     @objc(beforeEachWithMetadata:)
     internal func beforeEach(closure closure: BeforeExampleWithMetadataClosure) {
-        currentExampleGroup!.hooks.appendBefore(closure)
+        currentExampleGroup.hooks.appendBefore(closure)
     }
 
     internal func afterEach(closure: AfterExampleClosure) {
-        currentExampleGroup!.hooks.appendAfter(closure)
+        currentExampleGroup.hooks.appendAfter(closure)
     }
 
     @objc(afterEachWithMetadata:)
     internal func afterEach(closure closure: AfterExampleWithMetadataClosure) {
-        currentExampleGroup!.hooks.appendAfter(closure)
+        currentExampleGroup.hooks.appendAfter(closure)
     }
 
     @objc(itWithDescription:flags:file:line:closure:)
     internal func it(description: String, flags: FilterFlags, file: String, line: UInt, closure: () -> ()) {
         let callsite = Callsite(file: file, line: line)
         let example = Example(description: description, callsite: callsite, flags: flags, closure: closure)
-        currentExampleGroup!.appendExample(example)
+        currentExampleGroup.appendExample(example)
     }
 
     @objc(fitWithDescription:flags:file:line:closure:)
@@ -82,13 +82,13 @@ extension World {
     @objc(itBehavesLikeSharedExampleNamed:sharedExampleContext:flags:file:line:)
     internal func itBehavesLike(name: String, sharedExampleContext: SharedExampleContext, flags: FilterFlags, file: String, line: UInt) {
         let callsite = Callsite(file: file, line: line)
-        let closure = World.sharedWorld().sharedExample(name)
+        let closure = World.sharedWorld.sharedExample(name)
 
         let group = ExampleGroup(description: name, flags: flags)
-        currentExampleGroup!.appendExampleGroup(group)
+        currentExampleGroup.appendExampleGroup(group)
         currentExampleGroup = group
         closure(sharedExampleContext)
-        currentExampleGroup!.walkDownExamples { (example: Example) in
+        currentExampleGroup.walkDownExamples { (example: Example) in
             example.isSharedExample = true
             example.callsite = callsite
         }
--- a/Quick/DSL/World+DSL.swift
+++ b/Quick/DSL/World+DSL.swift
@@ -17,6 +17,9 @@ extension World {
     }
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
+        guard currentExampleGroup != nil else {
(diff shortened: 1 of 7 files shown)


Quick/DSL/World+DSL.swift
```diff
@@ -17,8 +17,11 @@ extension World {
     }
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
+        guard currentExampleGroup != nil else {
+            fatalError("Error: example group was not created by its parent QuickSpec spec. Check that describe() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)")
+        }
```
modocache: This is a great improvement! Thanks! :100: 

Two suggestions:

1. The end user may be using `context`, which simply calls `describe`. This error message could be updated to read "Check that context() or describe() was used in...", or something similar.
2. Would it be possible to add a unit test for this? I'm aware that the previous behavior wasn't unit tested either, but adding a test here would be a great way we catch regressions that would prevent users from seeing this helpful message. This may involve using an `NSException` instead of `fatalError`, since the latter cannot be unit tested (yet). I realize that's less "Swift-y", but from an end user's perspective, that shouldn't matter.

Thoughts? :raised_hands:

Quick/World.swift
```diff
@@ -57,12 +57,7 @@ final internal class World: NSObject {
     // MARK: Singleton Constructor
 
     private override init() {}
-    private struct Shared {
-        static let instance = World()
-    }
-    internal class func sharedWorld() -> World {
-        return Shared.instance
-    }
+    static let sharedWorld = World()
```
modocache: Very nice, thanks! :+1:

You could tell I wasn't super knowledgeable about globals in Swift when I wrote this... :stuck_out_tongue:

modocache: If it's easy to do, it might be nicer for commit history if this change was made in a separate commit. Feel free to ignore this comment if it's too much trouble to split this commit in two.

modocache:
Excellent stuff, thanks! I had a few suggestions:

1. Update the failure message to indicate the user may be using `describe()` *or* `context()`.
2. Add a unit test to ensure this change doesn't regress later.
3. Split this into two commits: one for the `sharedWorld` update, and one for the new failure message.

If you don't have time for items 1 or 3, I'll follow up by creating issues and taking care of them in a separate pull request. :cool:

Quick/World.swift
morganchen12: I'll give it a shot. I think the old private struct way was the only way to do singletons a year ago when this code was committed. :)

morganchen12:
I'll try to polish this stuff up by tonight.

Quick/DSL/World+DSL.swift
morganchen12: Agreed, unit test is good since we're coupling this error message with assumptions that come from outside this class/file. I'm okay with using an `NSException` (and perhaps even writing a Swiftier wrapper around it) for the sake of testing in this case.

Quick/DSL/World+DSL.swift
```diff
@@ -17,8 +17,15 @@ extension World {
     }
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
+        guard currentExampleGroup != nil else {
+            NSException(name: NSInternalInconsistencyException,
+                reason:"Example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)",
+                userInfo: nil).raise()
+            return
```
morganchen12: Apple should make `-[NSException raise]` a `noreturn` func so this extra `return` isn't necessary :disappointed:

QuickTests/FunctionalTests/DescribeTests.swift
```diff
@@ -0,0 +1,20 @@
...
+@testable import Quick
+import Nimble
+
+class DescribeTests: XCTestCase {
+    
+    func testDescribeErrorsIfExampleGroupsAreNotInstantiated() {
+        
+        // fiddling with global state in this test to test a singleton, what could go wrong?
+        let exampleGroup = World.sharedWorld.currentExampleGroup
+        World.sharedWorld.currentExampleGroup = nil
```
morganchen12: Messing with the `World` singleton without knowledge of whatever spec is currently using it feels dangerous.

modocache: Agreed... Would it be possible to write a test that encounters the same error a user would? That is, a test that uses a `describe()` outside of `QuickSpec.spec()`?

Something like:

```swift
import Quick

var isRunningTest = false

class DescribeThrowsTests: XCTestCase {
    func testUsingADescribeOutsideOfASpecThrows() {
        if isRunningTest {
            describe("This will cause an exception to be thrown") {}
        }
    }
}

class DescribeThrowsSpec: QuickSpec {
    override func spec() {
        expect { qck_runSpec(DescribeThrowsTests.classForCoder()) }
            .to(raiseException())
    }
}
```

This may not work because of how Quick initializes its example tree, but it might be worth a shot!

morganchen12: Seems like `QuickSpec` specs don't clean up their `World` state after running, they just rely on other specs to totally overwrite it. So in this case `describe` is still valid based on the last spec's `World` state and won't properly error here (or anywhere else) unless `World` has never been initialized by any spec, i.e. in a pure `XCTestCase` environment or a mixed environment before any `QuickSpec`s have run. I'll dig around a bit more and try to find a better way--manually mucking up `World` state and then testing for failure seems like too much of a self-fulfilling prophecy to be a good unit test.

modocache: Yeah... it might be worth it to strip the last commit that adds the unit test from this PR, and I'll just merge the first two commits. We can then create an issue to track the unit test! :+1:

morganchen12:
Looks like removing the new unit test file nuked a bunch of comments that were there.

Test has been deferred for now. :+1:

modocache:
We can probably nuke commit db1777b28aeb587cd2d51c281fc5d22b3ba00e21 entirely--the point of changing the `fatalError` to an `NSException` was to add the unit test. Without the test, I think `fatalError` is totally fine.

Sorry to send you down a rabbit hole!! :rabbit2: :sweat_drops:

morganchen12:
The rabbits have been secured. :rabbit:

jeffh:
Setting the World's `currentExampleGroup` to `nil` after [this line](https://github.com/Quick/Quick/blob/master/Quick/QuickSpec.m#L35) would make QuickSpec properly clean the example.

I don't know if there's a benefit for not cleaning it up, so that would allow this error to be more useful beside first-test uses.

morganchen12:
Updated with post-test world state cleanup and now a unit test that doesn't require fiddling with Quick internals.

In the future (especially with the open sourcing of Swift) we may want to consider de-singletonizing `World` altogether to allow for tests to run in parallel. Currently I don't think multiple async modifications of `World` will work very well.

jeffh:
Thanks @morganchen12! Yes, I agree that having a singleton world is definitely not ideal. I've creating that as a [separate issue](https://github.com/Quick/Quick/issues/435).

PR #437, opened 2015-12-07
Noted that `context` is a synonym for `describe`

mikemee:
I propose adding this sentence after spending 15+ mins trying to work out if `context` added any new syntax or behaviour beyond `describe`. Digging into the source showed that it is a synonym. I tried to make my addition match the tone of the document (_which is wonderfully written for a broad audience btw!_).


jeffh:
Thanks @mikemee! :+1:

PR #442, opened 2015-12-16
Remove a handful of inadvertently(?) committed files

briancroom:
I was browsing and noticed a bunch of files that don't seem to belong. These were added by @jeffh in c65f68f2699b0a026f0f846dddfbb5ad69de47d4 after git presumably auto-staged them in their initial state.


modocache:
Yup, I agree these were probably accidentally committed. Thanks, @briancroom!

PR #436, opened 2015-12-05
[WIP] Updates to support Swift package manager.

paulyoung:
I've marked this as WIP because:

giphy

Here's what `swift build` says:

```
Compiling Swift Module 'Quick' (16 sources)
Linking Library:  .build/debug/Quick.a
/Applications/Xcode7.1.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /Users/paulyoung/git/Quick/Quick/.build/debug/Quick.a(Closures.swift.o) has no symbols
```

PR #443, opened 2015-12-16
Ignore Swift Package Manager's `.build` directory

briancroom:
This will be useful with #436 and future work.


briancroom:
Derp, it looks like #436 includes this already. Nothing to see here.

Issue #433, opened 2015-12-03
Cross-platform support


modocache:
Quick should work on all platforms. This is made easier thanks to XCTest's front-end API being available on those platforms: https://github.com/apple/swift-corelibs-xctest

This issue tracks:
- [ ] Implementing `-[QuickSpec allTests]` so that platforms not using the Objective-C runtime can detect test methods: https://github.com/apple/swift-corelibs-xctest#additional-considerations-for-swift-on-linux
- [ ] Documenting use of `XCTMain` in the Quick documentation, or wrapping it using Quick such that the process is automatic
- [ ] Testing Quick against open-source Swift on OS X and ensuring everything works
- [ ] Testing Quick against open-source Swift on Linux and ensuring everything works
- [ ] Updating the Quick test suite such that it passes on all platforms

modocache:
Well, Quick's unit tests work when running Xcode via `xcrun launch-with-toolchain /Library/Developer/Toolchains/swift-latest.xctoolchain`. Looks promising.

modocache:
Quick imports XCTest, whereas the open-source XCTest defines a module named SwiftXCTest. Using Quick from within a test that uses SwiftXCTest results in an error:

```
'XCTestCase' is ambiguous for type lookup in this context
```

I wonder if there's a way we can detect which version of XCTest is available? Or we can provide two products: one that works with XCTest, and one that works with SwiftXCTest.

modocache:
We probably need two products: Quick-OSX, and Quick-OSX-SwiftXCTest. The second should be written entirely in Swift.

unchartedworks:
If you want support cross-platform, I don't think using framework as product is a good idea. Swift Package Manager doesn't support framework yet.

ikesyo:
The feature was implemented in #444 and #687. Also there is a dedicated documentation issue #675. So let's close this.

PR #444, opened 2015-12-17
Initial support for specs on Linux

briancroom:
This is a first pass at allowing Quick specs to be built and run on Linux (#433). There are a few pieces to this:

* Add a new basic `QuickSpec` implementation written in Swift. This is only intended for use with `swift-corelibs-xctest` and implements the `allTests` property.
* Work around a few Foundation bits not being present in `swift-corelibs-foundation`: `NSException` and `NSStringFromClass`
* Work around `@objc` not being usable on Linux (for now?) because the `NSObject` provided by `swift-corelibs-foundation` isn't recognized as being the Objective C root class

Due to the invasive nature of some of these workarounds, this probably shouldn't be merged in its current state, but I wanted to get the ball rolling at least! I've pushed a [repo](https://github.com/briancroom/QuickOnLinuxExample) demonstrating this in use.


briancroom:
Also, I discovered that it is currently not feasible to `import XCTest` from any file when building with the Swift Package Manager on OS X because it doesn't provide `swiftc` with the necessary search paths to find `XCTest.framework`.

jeffh:
Thanks for the work so far, @briancroom. @modocache mentioned that XCTest for cross-platform swift is actually the `SwiftXCTest` module.

I guess I better start dusting off my linux VM....

briancroom:
There's plenty left to do here, hopefully I can continue to make time to carry it forward.

the open-source XCTest builds the module SwiftXCTest on Mac, but it's still XCTest on Linux because they are aiming for source-level compatibility between the platforms.

Package.swift
```diff
@@ -1,5 +1,13 @@
...
 
 let package = Package(
-  name: "Quick"
+    name: "Quick",
+    targets: [
+        Target(name: "QuickTests", dependencies: [.Target(name: "Quick"), .Target(name: "QuickTestHelpers")]),
+        Target(name: "QuickFocusedTests", dependencies: [.Target(name: "Quick"), .Target(name: "QuickTestHelpers")]),
+        Target(name: "QuickTestHelpers", dependencies: [.Target(name: "Quick")]),
+    ],
+    dependencies: [ // TODO: This ought to be `testDependencies:` but there seems to be a swiftpm bug causing the lib to not get linked then
```
modocache: Could you include a Swift JIRA issue number or a link to a swift-package-manager pull request here? That way a future contributor could determine whether the bug has been fixed and this code could be changed.

Sources/Quick/Configuration/Configuration.swift
```diff
@@ -72,10 +72,16 @@ final public class Configuration: NSObject {
         provided with metadata on the example that the closure is being run
         prior to.
     */
+#if !os(Linux)
```
modocache: I see quite a few conditional macros in the Swift codebase along the lines of "if the Objective-C runtime is available". Could we change this to something like that? The rationale being that what we really want to check here is whether `@objc` is available, not what the operating system is. Even now this isn't completely correct--what if the OS is FreeBSD? :wink:

Package.swift
briancroom: I'm actually thinking now that this isn't really a bug, but rather has to do with the `swift test` command not having been built out yet. Presumably test dependencies shouldn't be linked when building the package itself, only its tests, however swiftpm hasn't been properly taught about test targets yet. I'll update the comment accordingly.

Sources/Quick/Configuration/Configuration.swift
```diff
@@ -103,10 +109,16 @@ final public class Configuration: NSObject {
         is provided with metadata on the example that the closure is being
         run after.
     */
+#if !os(Linux)
```
modocache: Same as above.

I should mention, it's very cool that these conditional macros are available in Swift, and this is a great solution to the problem here--thanks, @briancroom!

> Aside: At her Brooklyn Swift talk @tamarnachmany spoke about giving constructive feedback to theater performers. The difference between "That was a stellar rehearsal. [...] Oh, one note, you messed up a line towards the end." and "You messed up a line at the end. Excellent rehearsal, though." is **huge**!
>
> In that respect, I now see why some people prefer code review tools like Phabricator or Gerrit--when you click "submit", an overall comment is displayed first, then in-line comments. On GitHub, you'll read my in-line nitpicks before you read my overarching comment on the pull request. That's a shame, because this is amazing work and it's much appreciated, even though my in-line comments focus on what could be improved.

Sources/Quick/Configuration/QuickConfiguration.swift
```diff
@@ -0,0 +1,8 @@
+import XCTest
+
+// NOTE: This file is not intended to be included in the Xcode project. It
+//       is picked up by the Swift Package Manager during its build process.
```
modocache: Great note. Similar files in swift-package-manager always confuse me; it takes a long time to grok the build system.

Sources/Quick/DSL/World+DSL.swift
```diff
@@ -20,8 +20,13 @@ extension World {
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
         guard currentExampleGroup != nil else {
-            NSException(name: NSInternalInconsistencyException, reason:"Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)", userInfo: nil).raise()
-            return
+            let error = "Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)"
+            #if os(Linux)
```
modocache: Same as above: we want to check whether `NSException` is available, not whether the OS is Linux.

Sources/Quick/Configuration/Configuration.swift
briancroom: Oh neat, I'll try to track down those macro names so they can be used instead. Agreed that just checking for Linux is imprecise and not very future-proof.

Sources/Quick/Configuration/Configuration.swift
briancroom: That's a great insight, yeah.

I've been working on GitHub for long enough to know the drill, but still appreciate you calling it out like that :+1:

Sources/Quick/DSL/World+DSL.swift
```diff
@@ -50,42 +55,50 @@ extension World {
...
     internal func afterEach(closure closure: AfterExampleWithMetadataClosure) {
         currentExampleGroup.hooks.appendAfter(closure)
     }
+#else
+    internal func afterEach(closure closure: AfterExampleWithMetadataClosure) {
+        currentExampleGroup.hooks.appendAfter(closure)
+    }
+#endif
 
-    @objc(itWithDescription:flags:file:line:closure:)
```
modocache: Why remove these? I believe these were originally added to make the methods more readable in Objective-C:

```objc
// With the @objc name:
[[World sharedWorld] itWithDescription:@"foo" flags:@{} /* ... */];

// Without the @objc name:
[[World sharedWorld] itDescription:@"foo" flags:@{} /* ... */];
```

Sources/Quick/Example.swift
```diff
@@ -78,7 +78,7 @@ final public class Example: NSObject {
         }
         world.exampleHooks.executeAfters(exampleMetadata)
 
-        ++numberOfExamplesRun
+        numberOfExamplesRun += 1
```
modocache: :+1:

Sources/Quick/QuickMain.swift
```diff
@@ -0,0 +1,18 @@
+import XCTest
+
+@noreturn public func QCKMain(specs: [XCTestCase], configurations: [QuickConfiguration.Type] = []) {
```
modocache: So cool!! :heart_eyes:

Sources/Quick/QuickSpec.swift
```diff
@@ -0,0 +1,29 @@
...
+
+// NOTE: This file is not intended to be included in the Xcode project. It
+//       is picked up by the Swift Package Manager during its build process.
+
+public class QuickSpec: XCTestCase, XCTestCaseProvider {
+    public func spec() {}
+
+    public required init() {}
+
+    public var allTests : [(String, () -> Void)] {
```
modocache: I'm sure you're following along with https://github.com/apple/swift-corelibs-xctest/commit/e5c0cb36bc445d0522f7ec28a8342b2681f68666, so you're probably also already aware these should be converted to `[(String, () throws -> Void)]`. Still, just making a note of it here. There's also a spot below that needs to be modified as well.

> Aside: Boy, how are we ever going to version this thing? :grin:

Sources/Quick/DSL/World+DSL.swift
```diff
@@ -20,8 +20,13 @@ extension World {
...
         guard currentExampleGroup != nil else {
-            NSException(name: NSInternalInconsistencyException, reason:"Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)", userInfo: nil).raise()
-            return
+            let error = "Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)"
+            #if os(Linux)
+                fatalError(error)
+            #else
+                NSException(name: NSInternalInconsistencyException, reason: error, userInfo: nil).raise()
+                return ()
+            #endif
```
modocache: Some context on why we're using `NSException` here in the first place:

@morganchen12 was kind enough to include the very descriptive error message you see above in #431. When he did, I asked him to include a unit test for the change. Unfortunately, we can't test `fatalError` (as far as I can tell), so we stuck with `NSException` here, since we can test that an `NSException` is thrown.

We'd be able to get rid of these conditional macros and the `NSException` if we had a good plan for how to test that users actually see this nice error message.

Sources/QuickTestHelpers/SpecRunner.swift
```diff
@@ -0,0 +1,24 @@
...
+public func qck_runSpec(specClass: QuickSpec.Type) -> TestRun {
+    return qck_runSpecs([specClass])
+}
+
+public func qck_runSpecs(specClasses: [QuickSpec.Type]) -> TestRun {
+    World.sharedWorld.isRunningAdditionalSuites = true
+
+    var executionCount: UInt = 0
+
+    let fails = gatherFailingExpectations(silently: true) {
```
modocache: Wow, I had no idea about this Nimble function. Cool!

Sources/QuickTests/Helpers/XCTestCaseProvider.swift
```diff
@@ -0,0 +1,35 @@
...
+    private func assertContainsTest(name: String) {
+        let contains = self.allTests.contains({ test in
+            return test.0 == name
+        })
+
+        XCTAssert(contains, "Test '\(name)' is missing from the allTests array")
+    }
+}
+
+#endif
```
modocache: Tangential to this pull request, but we really have to solve this problem in swift-corelibs-xctest! I'm curious if it makes sense to you to have the above shims ship as part of swift-corelibs-xctest?

modocache:
As I mentioned in some of my in-line comments, this is crazy cool. I had a few suggestions, but otherwise this looks great! Is this ready to be merged once the comments are addressed? Is there anything you'd like help with/input on before that happens?

Sources/Quick/QuickSpec.swift
briancroom: I'm targeting the latest (January 11th) developer snapshot right now, which doesn't yet include that (breaking) change, but it will be necessary any day now.

Re: versioning, I'm not convinced that it makes sense to put too much effort into versioning until there's some kind of formal release (alpha, beta, whatever) that includes `corelibs` and `swiftpm`, which probably won't happen before the summer at least. In the meantime, anybody working with these tools should be aware that they are a moving target and expect breakage at every turn. To me, that is the strongest argument against merging this yet.

Sources/QuickTests/Helpers/XCTestCaseProvider.swift
briancroom: It may be reasonable to have this file present in the `swift-corelibs-xctest` repo, but actually *shipping* it with the library doesn't do much good, because it's only useful in contexts that aren't using it! (cf. my comment [here](https://bugs.swift.org/browse/SR-32?focusedCommentId=11365&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-11365))

briancroom:
I've pushed a couple updates to conditionalize using `#if _runtime(_ObjC)` now which does feel better, even though it's _clearly _a _private _API _which _could _break _sometime :wink: We'll know about it in advance if that happens though, so I'm not too worried.

I'm excited to say that this branch now builds on Linux and all tests execute and pass! I'm hoping to see if I can set up a Travis build matrix that convinces it to run a Linux job for this as well, but that could easily go into a separate PR.

On the whole, I'm a lot happier with the state of this than I am with [Nimble's PR](https://github.com/Quick/Nimble/pull/226) and would have few qualms with seeing it merged. I expect we will need to deal with multiple breaking changes in the underlying tools between now and when Swift 3.0 is released, but I don't anticipate anything too painful.

briancroom:
And with that, the Travis CI build includes a second (passing!) job which executes all the tests on Linux. Thanks to @kylef for [tools](https://github.com/kylef/swiftenv) and [examples](https://github.com/kylef/Curassow/blob/master/.travis.yml) that made this fairly straightforward.

It doesn't appear that CircleCI supports multi-OS builds at this point, so it's just Travis for now.

Sources/Quick/QuickSpec.swift
modocache: Hmm, we're not the only ones with this problem: a [recent email](https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20160111/000357.html) to the swift-corelibs-dev mailing list explained which snapshot was necessary to build swift-corelibs-foundaiton.

Until someone comes up with a little badge to put in the README, we *may* want to document somewhere which version of swift-corelibs-xctest we're targeting. Perhaps in the docblock to `QCKMain()`?

Sources/Quick/QuickSpec.swift
```diff
@@ -0,0 +1,29 @@
+import XCTest
+
+// NOTE: This file is not intended to be included in the Xcode project. It
+//       is picked up by the Swift Package Manager during its build process.
+
+public class QuickSpec: XCTestCase, XCTestCaseProvider {
```
modocache: I didn't notice this on my first pass, but this could use documentation. Perhaps include the latest snapshot we've tested against?

script/travis-install-linux
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+# See: https://github.com/kylef/swiftenv/wiki/Travis-CI
+curl -sL https://gist.github.com/kylef/5c0475ff02b7c7671d2a/raw/621ef9b29bbb852fdfd2e10ed147b321d792c1e4/swiftenv-install.sh | bash
```
modocache: Does this download a specific snapshot?

modocache:
This looks great to me! I had a few additional comments, but those can be addressed in future pull requests. Let's get this in master!

Also, it'd be great to [add documentation](https://github.com/Quick/Quick/tree/master/Documentation) on how to set up a project using Quick and the package manager.

script/travis-install-linux
kylef: It downloads the contents of `.swift-version` file (https://github.com/briancroom/Quick/blob/initial-linux/.swift-version)

modocache: Aha! Super cool, thanks @kylef :+1: 

All the more reason to add documentation in #459, though--I'm used to seeing which versions I need in a `.travis.yml` file or similar, but this additional layer of indirection threw me for a loop.

kylef: It could also be set in `.travis.yml` via an environmental variable (as `SWIFT_VERSION`) if preferred (instead of the file), but the file is more generic and then swiftenv locally will automatically use the correct version (or error if not installed).

Sources/Quick/DSL/World+DSL.swift
morganchen12: Testing precondition/assert/fatalError has been [brought up in the swift corelibs mailing list](https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20151207/000166.html), but I'm not sure if there's any work being done on it.

modocache: Thanks for the link, @morganchen12! I hadn't noticed that discussion.

There's an issue on Nimble, https://github.com/Quick/Nimble/issues/234, that discusses this topic further.

Issue #362, opened 2015-08-17
Error when the context, describe, it, beforeEach, afterEach are used incorrectly


jeffh:
Background: #357 

Quick should raise an error (using ~~`fatalError`~~ `NSException`) when any of the DSL functions are used incorrectly:

- [ ] Using `it` inside `beforeEach`
- [ ] Using `it` inside `afterEach`
- [ ] Using `afterEach` inside `it`
- [ ] Using `beforeEach` inside `it`
- [ ] Using `context` inside `it`
- [ ] Using `describe` inside `it`
- [ ] Using `it` inside `it`

An error message should look like:

> 'afterEach' cannot be used inside 'it', 'afterEach' may only be used inside 'context' or 'describe'. (/Users/../exampleTest.swift:45)

briancroom:
This was completed by #448. Thanks @jeffh and @jwfriese!

PR #448, opened 2016-01-02
Error out with more specific messages when misusing DSL functions

jwfriese:
This is a crack at solving #362. 

I spent a bit of time trying to use the properties that existed on World to reliably deduce where the current DSL function lied contextually. Unfortunately, doing so made it tough to produce exact error messages for, say, a 'beforeEach' within a 'describe' vs one within a 'context', so I had to switch gears. Overall, I'm happy with what resulted -- a new property on World to track the containing DSL function of the executing function, and a helper method that manages the logic of the throwing the error. 

Admittedly, I'm concerned that managing the 'currentExampleContextDescriptor' property might be a little clunky and hard to intuit to someone walking into the code for the first time. I'd love some suggestions here if you agree that it could be better designed.

Additionally, this is my first attempt at a PR. Let me know if there are any improvements to be made to that end, especially in this write-up.

### DIFF
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -96,6 +96,10 @@ extension World {
             NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
             return
         }
+        if aftersCurrentlyExecuting() {
+            NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
+            return
+        }
         guard currentExampleMetadata == nil else {
             NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'it', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
             return
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -20,8 +20,7 @@ extension World {
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
         guard currentExampleMetadata == nil else {
-            NSException(name: "Invalid DSL Exception", reason: "'describe' cannot be used inside 'it', 'describe' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'describe' cannot be used inside 'it', 'describe' may only be used inside 'context' or 'describe'. ")
         }
         guard currentExampleGroup != nil else {
             raiseError("Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)")
@@ -35,8 +34,7 @@ extension World {
 
     internal func context(description: String, flags: FilterFlags, closure: () -> ()) {
         guard currentExampleMetadata == nil else {
-            NSException(name: "Invalid DSL Exception", reason: "'context' cannot be used inside 'it', 'context' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'context' cannot be used inside 'it', 'context' may only be used inside 'context' or 'describe'. ")
         }
         self.describe(description, flags: flags, closure: closure)
     }
@@ -55,8 +53,7 @@ extension World {
 
     internal func beforeEach(closure: BeforeExampleClosure) {
         guard currentExampleMetadata == nil else {
-            NSException(name: "Invalid DSL Exception", reason: "'beforeEach' cannot be used inside 'it', 'beforeEach' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'beforeEach' cannot be used inside 'it', 'beforeEach' may only be used inside 'context' or 'describe'. ")
         }
         currentExampleGroup.hooks.appendBefore(closure)
     }
@@ -74,8 +71,7 @@ extension World {
 
     internal func afterEach(closure: AfterExampleClosure) {
         guard currentExampleMetadata == nil else {
-            NSException(name: "Invalid DSL Exception", reason: "'afterEach' cannot be used inside 'it', 'afterEach' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'afterEach' cannot be used inside 'it', 'afterEach' may only be used inside 'context' or 'describe'. ")
         }
         currentExampleGroup.hooks.appendAfter(closure)
     }
@@ -93,16 +89,13 @@ extension World {
 
     internal func it(description: String, flags: FilterFlags, file: String, line: UInt, closure: () -> ()) {
         if beforesCurrentlyExecuting {
-            NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ")
         }
         if aftersCurrentlyExecuting {
-            NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. ")
         }
         guard currentExampleMetadata == nil else {
-            NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'it', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'it' cannot be used inside 'it', 'it' may only be used inside 'context' or 'describe'. ")
         }
         let callsite = Callsite(file: file, line: line)
         let example = Example(description: description, callsite: callsite, flags: flags, closure: closure)
@@ -123,8 +116,7 @@ extension World {
 
     internal func itBehavesLike(name: String, sharedExampleContext: SharedExampleContext, flags: FilterFlags, file: String, line: UInt) {
         guard currentExampleMetadata == nil else {
-            NSException(name: "Invalid DSL Exception", reason: "'itBehavesLike' cannot be used inside 'it', 'itBehavesLike' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
-            return
+            raiseError("'itBehavesLike' cannot be used inside 'it', 'itBehavesLike' may only be used inside 'context' or 'describe'. ")
         }
         let callsite = Callsite(file: file, line: line)
         let closure = World.sharedWorld.sharedExample(name)
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -92,11 +92,11 @@ extension World {
 #endif
 
     internal func it(description: String, flags: FilterFlags, file: String, line: UInt, closure: () -> ()) {
-        if beforesCurrentlyExecuting() {
+        if beforesCurrentlyExecuting {
             NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
             return
         }
-        if aftersCurrentlyExecuting() {
+        if aftersCurrentlyExecuting {
             NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
             return
         }
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -73,6 +73,10 @@ extension World {
 #endif
 
     internal func afterEach(closure: AfterExampleClosure) {
+        guard currentExampleMetadata == nil else {
+            NSException(name: "Invalid DSL Exception", reason: "'afterEach' cannot be used inside 'it', 'afterEach' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
+            return
+        }
         currentExampleGroup.hooks.appendAfter(closure)
     }
 
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -19,6 +19,10 @@ extension World {
     }
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
+        guard currentExampleMetadata == nil else {
+            NSException(name: "Invalid DSL Exception", reason: "'describe' cannot be used inside 'it', 'describe' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
+            return
+        }
         guard currentExampleGroup != nil else {
             raiseError("Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)")
         }
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -34,6 +34,10 @@ extension World {
     }
 
     internal func context(description: String, flags: FilterFlags, closure: () -> ()) {
+        guard currentExampleMetadata == nil else {
+            NSException(name: "Invalid DSL Exception", reason: "'context' cannot be used inside 'it', 'context' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
+            return
+        }
         self.describe(description, flags: flags, closure: closure)
     }
 
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -92,6 +92,10 @@ extension World {
 #endif
 
     internal func it(description: String, flags: FilterFlags, file: String, line: UInt, closure: () -> ()) {
+        guard currentExampleMetadata == nil else {
+            NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'it', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
+            return
+        }
         let callsite = Callsite(file: file, line: line)
         let example = Example(description: description, callsite: callsite, flags: flags, closure: closure)
         currentExampleGroup.appendExample(example)
--- a/Sources/Quick/DSL/World+DSL.swift
+++ b/Sources/Quick/DSL/World+DSL.swift
@@ -20,7 +20,7 @@ extension World {
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
         guard currentExampleMetadata == nil else {
-            raiseError("'describe' cannot be used inside 'it', 'describe' may only be used inside 'context' or 'describe'. ")
+            raiseError("'describe' cannot be used inside '\(currentPhase)', 'describe' may only be used inside 'context' or 'describe'. ")
         }
         guard currentExampleGroup != nil else {
             raiseError("Error: example group was not created by its parent QuickSpec spec. Check that describe() or context() was used in QuickSpec.spec() and not a more general context (i.e. an XCTestCase test)")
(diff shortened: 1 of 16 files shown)


briancroom:
Hey @jwfriese! Thanks for taking the time to work on this PR. I think these diagnostics could help a lot to improve the experience for newcomers to BDD-styled tests. Here are a couple comments to get things started.

Firstly, I'd really love to see some tests around this new behavior! Though I realize that the use of `fatalError` makes that challenging. I wonder if it would make sense to raise `NSException`s instead, which could then be tested with Nimble's `raiseException` matcher. `DescribeTests.swift` is a good example of the sort of test I'd have in mind.

Secondly, I think there's an opportunity to simplify the validation logic a bit and make it more comprehensive at the same time. Quick has two phases of execution:

1. Defining examples, when the `spec` method is called on each `QuickSpec` instance to build the tree of examples. During this phase, the DSL items that are intended to allow nesting (`describe` etc.) have their closures immediately invoked, while those that don't allow nesting (`beforeEach`, `it`, `afterEach`) have their closures stashed away for the next phase
2. Running examples, when the gathered example trees are actually executed.

It seems to me that Quick could track the current phase (it sort of does already via `World`'s `currentExampleGroup` and `currentExampleMetadata` properties) and simply disallow *any* DSL method unless the 1st phase is in progress.

Sources/Quick/DSL/DSLValidation.swift
```diff
@@ -0,0 +1,28 @@
+/**
+    Descriptors intended to identify DSL function context.
+*/
+public enum ExampleContextDescriptor: String {
```
jeffh: Is this a reason for this to be public? I think we can leave this to `internal`. This project prefers being explicit about visibility scopes, so just replace `public` with `internal`.

Sources/Quick/DSL/DSLValidation.swift
```diff
@@ -0,0 +1,28 @@
...
+    case Top = "spec"
+    case Describe = "describe"
+    case Context = "context"
+    case It = "it"
+    case ItBehavesLike = "itBehavesLike"
+    case BeforeEach = "beforeEach"
+    case AfterEach = "afterEach"
+}
+
+public func validateDSLItemWithContext(currentContext: ExampleContextDescriptor, withinContext containingContext: ExampleContextDescriptor) {
```
jeffh: ditto about visibility. Change from `public` to `internal`.

Sources/Quick/DSL/DSLValidation.swift
```diff
@@ -0,0 +1,28 @@
+/**
+    Descriptors intended to identify DSL function context.
+*/
+public enum ExampleContextDescriptor: String {
+    case None = "none"
```
jeffh: I don't think this case is used and could probably be removed?

Sources/Quick/World.swift
```diff
@@ -38,6 +38,11 @@ final internal class World: NSObject {
     */
 
     internal var currentExampleMetadata: ExampleMetadata?
+    
+    /**
+        The descriptor of the current context.
+    */
```
jeffh: Nitpick: I rather have this comment removed or elaborated more. It currently doesn't provide extra value beyond the variable name.

jeffh:
Thanks for the PR @jwfriese. Overall it looks good :+1:. I agree with what @briancroom says about throwing an `NSException` so we can have test coverage for this behavior. That's my oversight when I wrote that original issue.

I left some comments in PR. Reply here to let me know when you've addresses the feedback (and changed to use `NSException` (with tests) or if you have more questions. Github doesn't notify me when the PR changes :cry:.

------------------------

@briancroom, I went with a more end-user facing error message in #362. It's difficult to succinctly explain  Quick's execution phases in one liner to make those kinds of error messages meaningful. I ended up opting to provide suggested fixes instead.

Sources/Quick/DSL/DSLValidation.swift
jwfriese: Thanks for pointing this out. I think this was simply an oversight on my part. It should absolutely be limited to internal access only. I've made that change in my in-flight update.

Sources/Quick/DSL/DSLValidation.swift
jwfriese: It's not used, and in the code there is no need for a "default" value like this. It will be removed.

Sources/Quick/World.swift
jwfriese: Agreed, it doesn't provide any extra value. I left it in there in hopes that maybe someone would have a suggestion of how to better name it :) 

After marinating for a couple days, maybe I can propose something better. Otherwise the update will remove it.

jwfriese:
@jeffh - Thanks for the thoughts and the inline code suggestions. I've responded to most of them with the action items I've taken from them! 

@briancroom - First of all, thanks for suggesting the alternative to ```fatalError``` that allows us to test. I'd wanted to put tests in for these error functions, but couldn't find how to easily do it with ```fatalError```, and wrongly assumed that there was some very special reason we wanted to use ```fatalError``` to begin with. I should have just clarified with @jeffh from the beginning. I will do so in the future!

Thank you also for the excellent high-level clarification of how Quick operates. I'm much better equipped to help this project move forward after reading it. 

To both of you, I'm working on an update that will address your concerns -- most importantly the testing. I'm hoping it will be simpler to understand, too. By this weekend I should have it.

briancroom:
@jeffh I'm totally on the same page as you regarding how the user-facing errors should be framed. I was more making a suggestion about how the 'validateDSLItem...' function could be simplified. 

Also @jwfriese, I can't shake the feeling that it *may* be possible to make this work without introducing an extra piece of global state (the ExampleContextDescriptor variable in World) since World maintains a lot of related state already, but I'm coming to the conclusion that certain parts of the desired behavior may not be possible without it, such as differentiating between 'describe' and 'context'. Anyway..food for thought.

jwfriese:
@briancroom - We're in agreement that the extra global state is nasty. I'm working to tease out an alternative solution using the existing data models to eliminate or at least reduce the extent to which global state was expanded in the original PR. 

We'll see how it goes :)

jwfriese:
Hey @briancroom & @jeffh,

Ok, so there's quite a bit to go over here. First of all, you'll notice I wrote over the previous commits. I've stashed the original commits of the PR safely away in a branch called ```first_more_descriptive_dsl_function_errors``` on my personal fork of Quick. So if we decide that this new direction is worse than the old one, we can easily jump back and I can build on that one instead. 

The first, and by far most important, item I'd like to point out is that these error cases are fully test-driven in this version of the PR. :) Thanks to both of you for guiding to me to ```NSException``` as an error-throwing option to allow this. 

Secondly, with @briancroom's advice in mind, I tried to make the code step away from the additional global state the previous PR had added to. You'll see that, for most cases, this was actually extremely easy to accomplish with simple ```guard```s like the following:

```swift
guard currentExampleMetadata == nil else {
    NSException(name: "Invalid DSL Exception", reason: "'describe' cannot be used inside 'it', 'describe' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
    return
 }
```

However, things started to get interesting when I wanted to handle 'it' blocks inside of 'beforeEach' and 'afterEach' blocks. At first, I thought it was going to be extremely simple to track the state of before- and afterEach execution patterns by simply adding one property onto the ```SuiteHooks``` class to track when hook execution began. Then, the error condition would be straightforward and look something like this:

```swift
if (suiteHooks.beforesStartedExecuting && !(suiteHooks.beforesAlreadyExecuted))  {
    NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
    return
 }
```

But, I still had tests failing. At that point I discovered that these hooks come from three places: the whole suite, the running group, and the running example. I would have to account for the execution pattern of all of these together as part of my error condition. 

For the example hooks, it seemed appropriate to me to add ```beforesStartedExecuting``` and ```beforesAlreadyExecuted``` properties, as it nicely followed the pattern already established in the ```SuiteHooks``` class. Tracking the group, however, is where it felt a little clunky. This is where I added similar properties on the ```ExampleGroup``` class: 

```swift
    internal var beforesStartedExecuting = false
    internal var beforesAlreadyExecuted = false
    internal var aftersStartedExecuting = false
    internal var aftersAlreadyExecuted = false
```
I then track these properties in the ```run()``` method on the example being executed. This smelled a little unsavory in the way that the micromanagement of global state felt in the previous PR. However, despite this, I still believe that this is a major step in the right direction, as this state management is localized to the classes relevant to the state. I feel the additions are appropriate. Especially since they end up condensing into such a nice, readable interface for the error condition: 

```swift
if beforesCurrentlyExecuting() {
    NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
    return
}
if aftersCurrentlyExecuting() {
    NSException(name: "Invalid DSL Exception", reason: "'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. ", userInfo: nil).raise()
    return
}
```

Overall, I'm much happier with this version of the PR. Thanks again @briancroom for encouraging me to look for a different, less global approach and for dropping some breadcrumbs to that end. Hopefully, this is something that we can move forward with. 

I'm looking forward to feedback on how we can improve this. I'd especially appreciate some comments on the quality of the technique for testing 'it' blocks in 'beforeEach' and 'afterEach' blocks. It is a very strange execution flow to try to test for error handling, so I had to do some unusual things there. But, we started red and went green with them, so I know they work for at least the basic cases I am testing.

QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -6,21 +6,71 @@ class FunctionalTests_ItSpec: QuickSpec {
     override func spec() {
         var exampleMetadata: ExampleMetadata?
         beforeEach { metadata in exampleMetadata = metadata }
-
+        
         it("") {
             expect(exampleMetadata!.example.name).to(equal(""))
         }
-
+
```
modocache: nit: Extra whitespace seems to have been added here. Mind removing it? :grin:

QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -6,21 +6,71 @@ class FunctionalTests_ItSpec: QuickSpec {
...
         }
-
+        
         it("has a description with セレクター名に使えない文字が入っている 👊💥") {
             let name = "has a description with セレクター名に使えない文字が入っている 👊💥"
             expect(exampleMetadata!.example.name).to(equal(name))
         }
+        
+        describe("error handling when misusing ordering") {
+            it("wraps another 'it' that will...") {
```
modocache: nit: Rather than use ellipses, I'd phrase this as `"an it"` `"a nested it throws an error"`.

I don't have a strong opinion either way, but note that the test selector for ellipses will be: `wraps_another__it__that_will_______throw_an_error`--each `.` is turned into an `_`.

Anyway, no strong opinions here, just food for thought! :cherries:

QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -6,21 +6,71 @@ class FunctionalTests_ItSpec: QuickSpec {
...
+            describe("behavior with an 'it' inside a 'beforeEach'") {
+                var exception: NSException?
+                
+                beforeEach {
+                    let capture = NMBExceptionCapture(handler: ({ e in
+                        exception = e
+                    }), finally: nil)
+                    
+                    capture.tryBlock {
+                        it("a rouge 'it' inside a 'beforeEach'") { }
```
modocache: nit/typo: I think you meant "rogue" here. http://grammarist.com/spelling/rogue-rouge/

QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -6,21 +6,71 @@ class FunctionalTests_ItSpec: QuickSpec {
...
+                
+                afterEach {
+                    let capture = NMBExceptionCapture(handler: ({ e in
+                        exception = e
+                        expect(exception).toNot(beNil())
+                        expect(exception!.reason).to(equal("'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. "))
+                    }), finally: nil)
+                    
+                    capture.tryBlock {
+                        it("a rouge 'it' inside an 'afterEach'") { }
```
modocache: Same typo as above :point_up:

QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -6,21 +6,71 @@ class FunctionalTests_ItSpec: QuickSpec {
...
+                        return
+                    }
+                }
+                
+                it("should throw an exception with the correct message after this 'it' block executes") {  }
+            }
+        }
```
modocache: These tests are great!! :heart_eyes:

Sources/Quick/World.swift
```diff
@@ -152,6 +152,28 @@ final internal class World: NSObject {
...
+    internal func aftersCurrentlyExecuting() -> Bool {
+        let suiteAftersExecuting = (suiteHooks.aftersStartedExecuting && !(suiteHooks.aftersAlreadyExecuted))
+        let exampleAftersExecuting = (exampleHooks.aftersStartedExecuting && !(exampleHooks.aftersAlreadyExecuted))
+        var groupAftersExecuting = false
+        if let runningExampleGroup = currentExampleMetadata?.example.group {
+            groupAftersExecuting = (runningExampleGroup.aftersStartedExecuting && !(runningExampleGroup.aftersAlreadyExecuted))
+        }
+        
+        return suiteAftersExecuting || exampleAftersExecuting || groupAftersExecuting
+    }
```
modocache: nit: Would these be better as derived variables? That is:

```swift
internal var beforesCurrentlyExecuting: Bool {
  // ...
  return suiteBeforesExecuting || exampleBeforesExecuting || groupBeforesExecuting
}
```

I'd be fine either way, but vars would be more consistent with the other variables, like `aftersStartedExecuting`.

QuickTests/FunctionalTests/ItTests.swift
jwfriese: Wow, thanks for the heads-up here. I had no idea that was the behavior!

QuickTests/FunctionalTests/ItTests.swift
jwfriese: Yeah, writing these tests was the most fun part of the whole PR :D

QuickTests/FunctionalTests/ItTests.swift
jwfriese: Ha! I of course was talking about a red 'it' block...

I'll fix both the typos. Thanks for the eagle-eye :)

modocache:
> @briancroom I think these diagnostics could help a lot to improve the experience for newcomers to BDD-styled tests. Here are a couple comments to get things started.

Absolutely agree. These are fantastic, thanks for working on them!

I had a few nitpicks. Aside from the typos, feel free to take the rest of the comments with a grain of salt--ignore them if you disagree.

Unfortunately I merged other pull requests before this one, and as a result this needs to be rebased on top of `master`. @jwfriese if you'd prefer, just say the word and I can pull this down and rebase myself. Otherwise, ping me after you've updated.

Thanks again!!

jwfriese:
@modocache 
Thanks for the great suggestions. I've updated the PR so the code is rebased against master, and applied your suggestions -- all of which I totally agreed with. 

After that, it was just getting the tests passing. It's all done now, but there were a couple things that I wanted to point out. 

1) Getting the tests to pass on Linux required the use of the ```_runtime(_ObjC)``` function. I saw this in the prod code, so I figured it was the best thing to use. Please let me know if there is some better way to do it. Unfortunately, these tests are effectively turned off for Linux users. :(

I'm wondering if there is some alternative testing mechanism we can use? Do you think this would be valuable as a formal issue? Or do we believe that ```NSException```  will be coming in a future version of ```swift-corelibs-foundation```? 

2) I had a hell of a time getting the CircleCI tests to pass for about an hour. It kept failing to build the AfterEachTests.o file because of the ```#endif``` symbol in the ContextTests.swift file. 

Finally, I looked at the code again, and noticed that there was no newline at the end of the ContextTests.swift file, and that it ended with an ```#endif```. So, I thought, "Hey, could we need a newline after the endif? Why not try it?".

Suddenly, a green CircleCI build appears.

I can sort of reason through why you cannot end a file with an ```#endif``` without a newline, but I'd be lying if I said that I totally understood. @modocache, @briancroom - do either of you find this surprising? Have you seen this before? 

Anyways, thanks again @modocache for giving this its latest look. Let me know if you have any feedback for today's iteration!

Sources/Quick/DSL/World+DSL.swift
```diff
@@ -19,6 +19,9 @@ extension World {
     }
 
     internal func describe(description: String, flags: FilterFlags, closure: () -> ()) {
+        guard currentExampleMetadata == nil else {
+            raiseError("'describe' cannot be used inside 'it', 'describe' may only be used inside 'context' or 'describe'. ")
```
briancroom: My only concern here is that cases like the following won't print the correct diagnostic:
```swift
beforeEach {
    describe("Uh oh this doesn't belong") {
    }
}
```
and the same for `afterEach`.

Perhaps we could add a method to look at the newly-introduced properties on the ExampleGroup and produce a string giving the name of the currently-executing block. Does that make any sense?

Sources/Quick/ExampleGroup.swift
```diff
@@ -8,6 +8,11 @@ import Foundation
 final public class ExampleGroup: NSObject {
     weak internal var parent: ExampleGroup?
     internal let hooks = ExampleHooks()
+    
+    internal var beforesStartedExecuting = false
```
briancroom: Modelling this state with 4 `bool`s creates the theoretical possibility of a bunch of invalid states. I know we were trying to move away from the large enum in the first iteration (sorry!) but I'd suggest using a smaller enum (5 cases, if I'm thinking about this correctly) here as the "source of truth". These particular booleans can then be computed properties on the enum, as necessary.

briancroom:
Thanks for sticking with us @jwfriese! I think we're going to land at a good spot with this. I've left just a few more in-line comments.

I am also a big fan of the tests you've got in here now :+1: 

As to your questions:

1) You've done the conditionals just as I would have. We really don't have a way to test these sorts of errors using the swift.org toolchain just yet. There's some discussion and links around that [here](https://github.com/Quick/Nimble/issues/234) already.

2) That sure feels like a compiler bug to me. Sorry you ended up wasting so many cycles on that!

Sources/Quick/DSL/World+DSL.swift
jwfriese: > My only concern here is that cases like the following won't print the correct diagnostic:

You're totally right. That's a detail that I missed. Thanks for pointing it out!

> Perhaps we could add a method to look at the newly-introduced properties on the ExampleGroup and produce a string giving the name of the currently-executing block. Does that make any sense?

Yes, it does. It would allow the code to do something *similar* to what that validator-type class in original PR tried to accomplish.

Sources/Quick/ExampleGroup.swift
jwfriese: > ... I'd suggest using a smaller enum ... These particular booleans can then be computed properties on the enum, as necessary.

We can definitely do this. I agree that the state machine has some danger built into as currently implemented. Let me see what I can come up with.

Sources/Quick/DSL/World+DSL.swift
```diff
@@ -134,4 +158,18 @@ extension World {
...
+            if beforesCurrentlyExecuting {
+                return "beforeEach"
+            } else if aftersCurrentlyExecuting {
+                return "afterEach"
+            }
+
+            return "it"
+        }
+
+        return ""
```
jwfriese: I'm curious how other people feel about this. From the new ```currentPhase``` function, it's technically possible to return an empty string if the function is misused. I don't feel super great about this, but I didn't feel super great about the alternatives either --

1) Return some legitimate value as a default value, like "it". 
2) Throw an error -- seems very heavyweight. 

Thoughts?

jwfriese:
@briancroom - 
Thanks again for the suggestions. I think that they have definitely led us to a better solution here. That crazy boolean-heavy state machine has been replaced by a much cleaner, simpler enum property. 

Additionally, I fixed the bug you pointed out by creating a function called ```currentPhase``` that infers the running execution phase. 

Let me know what you think of this iteration!

Sources/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -9,15 +9,67 @@ class FunctionalTests_ItSpec: QuickSpec {
     override func spec() {
         var exampleMetadata: ExampleMetadata?
         beforeEach { metadata in exampleMetadata = metadata }
-
+        
         it("") {
             expect(exampleMetadata!.example.name).to(equal(""))
         }
-
+
```
modocache: nit-pick: Extra whitespace :sweat_smile:

Sources/Quick/DSL/World+DSL.swift
modocache: This is only ever called from within `guard` statements that check for `currentExampleMetadata`; it's not meant to be used in any other situation. Furthermore, it doesn't use that value. So why not have it return `"it"` by default, as in:

```swift
private var currentPhase: String {
    if beforesCurrentlyExecuting {
        return "beforeEach"
    } else if aftersCurrentlyExecuting {
        return "afterEach"
    }
    return "it"
}
```

And lose the checks on `currentExampleMetadata`? You could include a comment to indicate when it should be used, or add a `precondition` that checks that `currentExampleMetadata != nil`.

modocache:
@jwfriese I had two comments, but neither seemed important enough to keep this unmerged. Feel free to submit additional pull requests to address my nit-picks--and thanks for all the hard work! :+1:

PR #452, opened 2016-01-10
[RFC] Workaround for behaviour when running a single spec in Xcode

### LINKED ISSUE #439: Proposal: A way to solve the strange behaviour with Xcode's test navigator
I found a _whole bunch_ of potentially related issues:

- https://github.com/Quick/Quick/issues/14
- https://github.com/Quick/Quick/issues/157
- https://github.com/Quick/Quick/issues/221
- https://github.com/Quick/Quick/issues/219
- https://github.com/Quick/Quick/issues/373
- https://github.com/Quick/Quick/issues/427
- https://github.com/Quick/Quick/issues/429

I did a bit of digging into what might be going on here.

When you press Cmd-U, XCTest seems to do roughly the following:

- For each bundle, builds the default test suite for that bundle
- For each bundle, creates a test suite named "All tests"
- For each test case class in the bundle, creates a test suite for that class by sending `+[XCTestSuite testSuiteForTestCaseClass:]`, and adds it to the "All tests" test suite
- Internally that method then sends `+testInvocations`, which is where Quick kicks in and builds the invocations from its example suite

When you click on a test suite in the test navigator, the behaviour is different:

- Only your test bundle is invoked
- A test suite named "Selected tests" is created
- XCTest sends `+[XCTestSuite testSuiteForTestCaseClassWithName:]`, where the name argument looks like this: `<spec-class>/<selector>`, e.g., `NewSpec/stuff__is_cool_pathtoNewSpecswift_6()`. (The trailing `()` might not be there for Objective-C, I haven't tested that.) **It sends this for each selector in the test navigator.** These test suites are added to the "Selected tests" test suite.
- `+testInvocations` is never sent.
- This method in turn sends the spec class the `+testCaseWithSelector:` message. It appears do do this once for each specific test invocation selector listed in the test navigator.

---

Here's the problem:

- You've run the tests once, and you can see everything in the test navigator.
- You run a specific spec, and everything looks fine.
- You change the description of one of the example group nodes, either a `describe()`, `context()` or `it()`.
- This changes the selector that will be generated at runtime. This is a different selector to the one that Xcode has cached in the test navigator.
- You re-run just that spec.
- When XCTest sends the test case class the `+testCaseWithSelector:` message, the `selector` that gets passed _doesn't exist anymore_.
- You get the weird `-[MyTests.MySpec (null)]` message.
- Stuff may even get weirder.

Semantically, when you try and run a specific spec, you really want to run all the examples in that spec. However, instead of asking our spec class to re-run all its examples, it instead is specifying a finite set of the test case methods it already knows about.

The only thing I've been able to come up with so far to fix this is to swizzle `+[XCTestSuite testSuiteForTestCaseWithName:]` and override it with our own semantics. But the selector we're asked for _might not even exist_.

So that means that we'd probably have to go up a level:

- Swizzle `-[XCTestSuite initWithName:]`
- Look for a test suite being created with the specific name "Selected tests"
- Swizzle this instance's class to be an instance of our own `XCTestSuite` subclass
- Build our own "Selected tests" test suite, and add it.
- Ignore any other tests that Xcode tries to add to this suite.

---

There are other factors to consider, and a risk is that we get even more coupled to Xcode. This implementation would be a start, but it's not ideal.

And with a version of XCTest going open-source, perhaps it would be better to adopt a wait-and-see approach?

sharplet:
Fixes #439.

The main fix this change introduces:

- When running a single spec from the Xcode test navigator, every example in that spec will always be run.

However, the "missing test case" behaviour is unfortunately still there:

- Whenever the "Selected tests" suite completes, **a single spec goes missing from the test navigator**. Any test failures in the missing test case will also be hidden.

But thankfully the issue where each subsequent test run would kill one more test case is resolved. Our tests are saved! :relieved: 

I'm open to any and all feedback on the implementation! I'm not how well I've kept to the Quick code style. I've tried to limit both the amount of method swizzling, and the amount of Objective-C.


Sources/Quick/NSBundle+CurrentTestBundle.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension NSBundle {
+
+    /**
+     Locates the first bundle with a '.xctest' file extension.
+     */
+    internal static var currentTestBundle: NSBundle? {
+        return allBundles().lazy
```
modocache: Oh cool, `CollectionType.lazy` is new to me. Good to know about--thanks!

Sources/Quick/NSBundle+CurrentTestBundle.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension NSBundle {
+
+    /**
+     Locates the first bundle with a '.xctest' file extension.
```
modocache: Hmm, is it possible to have multiple test bundles? What happens then? Based on the code below that uses `NSBundle.currentTestBundle`, I'd bet that if a test case class isn't in the first test bundle, this method would return `nil`, even if the test case exists in another bundle.

At first I thought Quick's own tests may have two test bundles:

screen shot 2016-01-16 at 11 01 55 am

But inspecting at runtime from within `QuickFocused-iOSTests` shows that they're not both available:

```
(lldb) po NSBundle.allBundles()
▿ 4 elements
  - [0] : NSBundle </Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/lib> (not yet loaded)
  - [1] : NSBundle </Users/bgesiak/Library/Developer/Xcode/DerivedData/Quick-druxpcbupaazldhgnyterfnxgtal/Build/Products/Debug-iphonesimulator/QuickFocused-iOSTests.xctest> (loaded)
  - [2] : NSBundle </Users/bgesiak/Library/Developer/Xcode/DerivedData/Quick-druxpcbupaazldhgnyterfnxgtal/Build/Products/Debug-iphonesimulator/QuickFocused-iOSTests.xctest/Frameworks> (not yet loaded)
  - [3] : NSBundle </Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/Library/Xcode/Agents> (loaded)
```

So I'm assuming this will work fine. Still, it would be nice to know if it's possible to have two test bundles at once. Any ideas?

modocache:
This seems like an improvement to me--thanks so much for working on it! :100: 

> Whenever the "Selected tests" suite completes, a single spec goes missing from the test navigator. Any test failures in the missing test case will also be hidden.

Do you mean hidden in the UI? Or is the "missing test case" never run, and the XCTest process returns a successful exit code?

Sources/Quick/NSBundle+CurrentTestBundle.swift
briancroom: When I looked into that question recently I discovered that Xcode launches the entire test host from scratch for each test bundle in the scheme, so I don't know of any situation where multiple test bundles would be loaded into a single process.

modocache: Good to know! Thanks :smile:

sharplet:
@modocache The test case is purely hidden in the UI. And on subsequent runs of the same spec, all test cases are executed and a single test case gets hidden again.

Since running a single spec from Xcode actually means _run the exact set of test cases listed in the navigator for this spec,_ before this change, every test run would execute one test case fewer than the previous run, and then another would go missing, until you had no test cases left to run.

Sources/Quick/QuickSelectedTestSuiteBuilder.swift
```diff
@@ -0,0 +1,69 @@
...
+     The test spec class to run.
+     */
+    let testCaseClass: AnyClass!
+
+    /**
+     For Objective-C classes, returns the class name. For Swift classes without,
+     an explicit Objective-C name, returns a module-namespaced class name
+     (e.g., "FooTests.FooSpec").
+     */
+    var namespacedClassName: String {
```
sharplet: I'm considering renaming this to just `className`, given that classes with Objective-C names aren't actually namespaced. What do you think?

modocache: At first glance this reminded me of the [`-[XCTestCase languageAgnosticTestMethodName]`](https://github.com/modocache/XCTest-Public-Headers/blob/4b638808d0a742f03d4613f71e2eb03f3118498d/Developer/Platforms/iPhoneOS.platform/Developer/Library/Frameworks/XCTest.framework/XCTest/XCTestCase.h#L66) I came across back in the day. It seemed like a distracting point to bring up in my first review of this pull request, but now that you mention changing this method name... :wink: 

I think `className` is less descriptive. How would you feel about `moduleNamespacedClassName`, or something like that? I know it's long-winded, but I think it's pretty clear.

modocache:
> The test case is purely hidden in the UI. And on subsequent runs of the same spec, all test cases are executed and a single test case gets hidden again.

Oh man, we should get this in right away! :clap:

Unfortunately I merged #444 just now, and that seems to have caused conflicts here. Let me know if you'd like to resolve them, or whether you'd prefer I merge manually. :+1:

Sources/Quick/QuickSelectedTestSuiteBuilder.swift
sharplet: Yeah, it's just that for Objective-C the class name isn't module namespaced! 😄 I suppose I was worrying whether this names is descriptive to the point of describing something that's not necessarily true 😉

What about `testCaseClassName` or `testSuiteClassName`?

modocache: `languageAgnosticNamespacedMethodName`? :grin: 

I'd be fine with `testSuiteClassName` as well.

AnthonyMDev:
@sharplet Awesome job here! Hope to see this pulled in soon!

sharplet:
Ok I've made the change to `testSuiteClassName` and pulled master. Hopefully if the build passes should be good to go! :v:

modocache:
@sharplet Looks like the Linux build failed :disappointed: You'll need to add conditionals to ensure that your additions only run on systems where the Objective-C runtime is available.

Still, super exciting! Almost there! :heart_eyes:

sharplet:
Ok @modocache looks like the linux build is fixed now, I went with this:

```swift
#if os(OSX) || os(iOS) ...
```

which I copied from another file. I also considered `#if os(Darwin)`, but I don't have time to verify whether it returns true for all Apple platforms.

Quick.xcodeproj/project.pbxproj
```diff
@@ -109,6 +109,21 @@
...
+		CE57CEDE1C430BD200D63004 /* QuickSelectedTestSuiteBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = CE57CED91C430BD200D63004 /* QuickSelectedTestSuiteBuilder.swift */; };
+		CE57CEDF1C430BD200D63004 /* QuickTestSuite.swift in Sources */ = {isa = PBXBuildFile; fileRef = CE57CEDA1C430BD200D63004 /* QuickTestSuite.swift */; };
+		CE57CEE01C430BD200D63004 /* String+FileName.swift in Sources */ = {isa = PBXBuildFile; fileRef = CE57CEDB1C430BD200D63004 /* String+FileName.swift */; };
+		CE57CEE11C430BD200D63004 /* XCTestSuite+QuickTestSuiteBuilder.m in Sources */ = {isa = PBXBuildFile; fileRef = CE57CEDC1C430BD200D63004 /* XCTestSuite+QuickTestSuiteBuilder.m */; };
+		CE590E1A1C431FE300253D19 /* QuickTestSuite.swift in Sources */ = {isa = PBXBuildFile; fileRef = CE57CEDA1C430BD200D63004 /* QuickTestSuite.swift */; };
+		CE590E1B1C431FE300253D19 /* QuickSelectedTestSuiteBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = CE57CED91C430BD200D63004 /* QuickSelectedTestSuiteBuilder.swift */; };
```
KinaMarie: Electromagnetic Signatures of R-process Nucleosynthesis in Neutron Star Binary Mergers (INT-17-2b)
 July 24 - August 18, 2017
 R. Fernández, D. Kasen, G. Martínez-Pinedo, B.D. Metzger

Spatial and Momentum Tomography of Hadrons and Nuclei (INT-17-3)
 August 28 - September 29, 2017
 I. Cloët, K. Hafidi, Z.-E. Meziani, B. PasquiniLeave a line note.... Don't you know I get vertigo when you do these things? Also, stop shining that light on my back just to see me split by a beam to pa through YOUR SLOTS LOLOL. Think outside the sandbox Joe. My best friend, foo, says he's willing to work with ya but if you hurt me anymore with your experiments, he will destroy you quick quick.

modocache:
Fantastic! Thanks, @sharplet! :bouquet:

sharplet:
:tada:

eimantas:
This one is a huge show stopper for me. Any chance you could make a release just with this fix?

modocache:
@eimantas Absolutely, I'll make a release this week. @Quick/contributors, feel free to release yourselves as well, if you have some spare time. There's a release script (thanks @jeffh!).

modocache:
@eimantas I've published a new release that includes this fix: https://github.com/Quick/Quick/releases/tag/v0.9.0

eimantas:
@modocache thanks a heap! 🎉

Issue #392, opened 2015-09-24
Prettier test method names in output


modocache:
#295 added file names and line numbers to test method names, which I consider to be aesthetically unpleasant. It would be great to make sure tests with identical names do not clash, but produce test output that doesn't affix file names or line numbers.

Perhaps only append file names or line numbers where clashes occur, but leave the test names alone otherwise?

jeffh:
I think only post fixing a number if a conflict occurred would be the least intrusive change. Otherwise I agree that simplifying the test names

brentleyjones:
When I noticed the file name and line numbers I thought it was a bug. They are so ugly now :(

abbeycode:
Agreed. For me, it's more about the absurd length (on top of already verbose names), and the end being all you can often see in the test navigator. Look at this from an Xcode Bot build in-progress:

<img width="1421" alt="Screenshot of bot test name" src="https://cloud.githubusercontent.com/assets/1349578/12075292/9f8a74e8-b149-11e5-976f-2ee198a4ca5f.png">

modocache:
A pull request would be much appreciated here! I believe a solution could be achieved by storing the name of each test selector in a set (stored on the shared `World` instance). If a selector already exists in the set, add an incrementing `1`, `2`... to its name.




On Sat, Jan 2, 2016 at 9:09 AM -0800, "Dov Frankel" <notifications@github.com> wrote:












Agreed. For me, it's more about the absurd length (on top of already verbose names), and the end being all you can often see in the test navigator. Look at this from an Xcode Bot build in-progress:

jeffh:
I don't think there needs to be any global state. We can check when creating all the test methods for each `QuickSpec` subclass.

modocache:
@abbeycode fixed this in #462. I'll release a new version soon, then close this issue. :wave:

NachoSoto:
I'm so excited for this :cl

abbeycode:
Should we close this issue out?

modocache:
Yup! Thanks for the reminder, @abbeycode!

Issue #401, opened 2015-10-08
Quick tests throw exceptions when running with xctool.


ScottRobbins:
I think this was an issue in the past. When I try to run my tests using xctool and quick, I get this error logged:
```
2015-10-08 10:41:29.423 xctool[33355:32011903] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Test should not have started yet.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff89c2703c __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x00007fff8c83f76e objc_exception_throw + 43
	2   CoreFoundation                      0x00007fff89c26e1a +[NSException raise:format:arguments:] + 106
	3   Foundation                          0x00007fff897e599b -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 195
	4   xctool                              0x0000000101a55714 -[OCTestEventState stateBeginTest] + 197
	5   xctool                              0x0000000101a1d614 -[TestRunState beginTest:] + 512
	6   xctool                              0x0000000101a1caba -[Reporter handleEvent:] + 1160
	7   xctool                              0x0000000101a1c612 -[Reporter parseAndHandleEvent:] + 151
	8   xctool                              0x0000000101a2a23e -[LineReader processBuffer] + 85
	9   xctool                              0x0000000101a2a3a7 -[LineReader dataAvailableNotification:] + 100
	10  CoreFoundation                      0x00007fff89be345c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12
	11  CoreFoundation                      0x00007fff89ad3634 _CFXNotificationPost + 3140
	12  Foundation                          0x00007fff89724aa1 -[NSNotificationCenter postNotificationName:object:userInfo:] + 66
	13  Foundation                          0x00007fff89859b66 _performFileHandleSource + 1622
	14  CoreFoundation                      0x00007fff89b43a01 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
	15  CoreFoundation                      0x00007fff89b35b8d __CFRunLoopDoSources0 + 269
	16  CoreFoundation                      0x00007fff89b351bf __CFRunLoopRun + 927
	17  CoreFoundation                      0x00007fff89b34bd8 CFRunLoopRunSpecific + 296
[...]

PR #462, opened 2016-01-20
Shorten selector names

abbeycode:
I followed the approach suggested in issue #392's thread, using an `NSSet` to track the existing unique selector names and append a number onto the end of additional ones to keep them unique.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,6 +21,7 @@ namespace "test" do
   task xctool: %w[test:xctool:ios test:xctool:osx]
   namespace :xctool do
     desc "Run unit tests for all iOS targets using xctool"
+    run "echo Using xctool v`xctool -v`..."
     task :ios do |t|
       run "xctool -workspace Quick.xcworkspace -scheme Quick-iOS -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6' clean test"
     end
--- a/Rakefile
+++ b/Rakefile
@@ -18,17 +18,23 @@ namespace "test" do
   end
 
   desc "Run unit tests for all iOS and OS X targets using xctool"
-  task xctool: %w[test:xctool:ios test:xctool:osx]
+  task xctool: %w[test:xctool:version test:xctool:ios test:xctool:osx]
   namespace :xctool do
     desc "Run unit tests for all iOS targets using xctool"
-    run "echo Using xctool v`xctool -v`..."
     task :ios do |t|
+      Rake::Task["test:xctool:version"].invoke
       run "xctool -workspace Quick.xcworkspace -scheme Quick-iOS -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6' clean test"
     end
 
     desc "Run unit tests for all OS X targets using xctool"
     task :osx do |t|
+      Rake::Task["test:xctool:version"].invoke
       run "xctool -workspace Quick.xcworkspace -scheme Quick-OSX clean test"
+    end
+
+    desc "Print the version of xctool being used"
+    task :version do
+      run "echo Using xctool v`xctool -v`..."
     end
   end
 
--- a/Sources/Quick/NSString+QCKSelectorName.m
+++ b/Sources/Quick/NSString+QCKSelectorName.m
@@ -27,7 +27,11 @@
 
     NSArray *validComponents = [self componentsSeparatedByCharactersInSet:invalidCharacters];
 
-    return [validComponents componentsJoinedByString:@"_"];
+    NSString *result = [validComponents componentsJoinedByString:@"_"];
+    
+    return ([result length] == 0
+            ? @"_"
+            : result);
 }
 
 @end
--- a/Sources/Quick/NSString+QCKSelectorName.m
+++ b/Sources/Quick/NSString+QCKSelectorName.m
@@ -4,12 +4,8 @@
 
 - (NSString *)qck_selectorName {
     static NSMutableCharacterSet *invalidCharacters = nil;
-    static NSString *separator = @"_";
-    static NSCharacterSet *separatorSet = nil;
     static dispatch_once_t onceToken;
-    
     dispatch_once(&onceToken, ^{
-        separatorSet = [NSCharacterSet characterSetWithCharactersInString:separator];
         invalidCharacters = [NSMutableCharacterSet new];
 
         NSCharacterSet *whitespaceCharacterSet = [NSCharacterSet whitespaceCharacterSet];
@@ -31,13 +27,7 @@
 
     NSArray *validComponents = [self componentsSeparatedByCharactersInSet:invalidCharacters];
 
-    NSString *selector = [validComponents componentsJoinedByString:separator];
-    selector = [selector stringByTrimmingCharactersInSet:separatorSet];
-    if ([selector length] == 0) {
-        selector = @"blank";
-    }
-
-    return selector;
+    return [validComponents componentsJoinedByString:@"_"];
 }
 
 @end
--- a/Sources/Quick/NSString+QCKSelectorName.m
+++ b/Sources/Quick/NSString+QCKSelectorName.m
@@ -4,8 +4,12 @@
 
 - (NSString *)qck_selectorName {
     static NSMutableCharacterSet *invalidCharacters = nil;
+    static NSString *separator = @"_";
+    static NSCharacterSet *separatorSet = nil;
     static dispatch_once_t onceToken;
+    
     dispatch_once(&onceToken, ^{
+        separatorSet = [NSCharacterSet characterSetWithCharactersInString:separator];
         invalidCharacters = [NSMutableCharacterSet new];
 
         NSCharacterSet *whitespaceCharacterSet = [NSCharacterSet whitespaceCharacterSet];
@@ -27,7 +31,13 @@
 
     NSArray *validComponents = [self componentsSeparatedByCharactersInSet:invalidCharacters];
 
-    return [validComponents componentsJoinedByString:@"_"];
+    NSString *selector = [validComponents componentsJoinedByString:separator];
+    selector = [selector stringByTrimmingCharactersInSet:separatorSet];
+    if ([selector length] == 0) {
+        selector = @"blank";
+    }
+
+    return selector;
 }
 
 @end
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -122,7 +122,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
     NSUInteger i = 2;
     
     while ([selectorNames containsObject:selectorName]) {
-        selectorName = [NSString stringWithFormat:@"%@_%li", example.name.qck_selectorName, i++];
+        selectorName = [NSString stringWithFormat:@"%@_%lu", example.name.qck_selectorName, i++];
     }
     
     [selectorNames addObject:selectorName];
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -5,6 +5,7 @@
 #import <objc/runtime.h>
 
 static QuickSpec *currentSpec = nil;
+static NSMutableSet<NSString *> *selectorNames;
 
 const void * const QCKExampleKey = &QCKExampleKey;
 
@@ -31,6 +32,8 @@ const void * const QCKExampleKey = &QCKExampleKey;
     world.currentExampleGroup = [world rootExampleGroupForSpecClass:[self class]];
     QuickSpec *spec = [self new];
 
+    selectorNames = [[NSMutableSet alloc] init];
+    
     @try {
         [spec spec];
     }
@@ -115,10 +118,15 @@ const void * const QCKExampleKey = &QCKExampleKey;
     }
 
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
-    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld",
-                              example.name.qck_selectorName,
-                              sanitizedFileName,
-                              (long)example.callsite.line];
+    NSString *selectorName = example.name.qck_selectorName;
+    NSUInteger i = 2;
+    
+    while ([selectorNames containsObject:selectorName]) {
+        selectorName = [NSString stringWithFormat:@"%@_%li", example.name.qck_selectorName, i++];
+    }
+    
+    [selectorNames addObject:selectorName];
+    
     SEL selector = NSSelectorFromString(selectorName);
     class_addMethod(self, selector, implementation, types);
 
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -119,7 +119,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
 
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
     NSString *selectorName = example.name.qck_selectorName;
-    NSUInteger i = 2;
+    unsigned long i = 2;
     
     while ([selectorNames containsObject:selectorName]) {
         selectorName = [NSString stringWithFormat:@"%@_%lu", example.name.qck_selectorName, i++];
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -118,7 +118,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
     
     NSString *originalName = example.name.qck_selectorName;
     NSString *selectorName = originalName;
-    unsigned int i = 2;
+    NSUInteger i = 2;
     
     static NSMutableSet<NSString *> *selectorNames;
     static dispatch_once_t onceToken;
@@ -127,7 +127,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
     });
     
     while ([selectorNames containsObject:selectorName]) {
-        selectorName = [NSString stringWithFormat:@"%@_%u", originalName, i++];
+        selectorName = [NSString stringWithFormat:@"%@_%lu", originalName, (unsigned long)i++];
     }
     
     [selectorNames addObject:selectorName];
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -58,8 +58,11 @@ const void * const QCKExampleKey = &QCKExampleKey;
 + (NSArray *)testInvocations {
     NSArray *examples = [[World sharedWorld] examplesForSpecClass:[self class]];
     NSMutableArray *invocations = [NSMutableArray arrayWithCapacity:[examples count]];
+    
+    NSMutableSet<NSString*> *selectorNames = [NSMutableSet set];
+    
     for (Example *example in examples) {
-        SEL selector = [self addInstanceMethodForExample:example];
+        SEL selector = [self addInstanceMethodForExample:example classSelectorNames:selectorNames];
         NSInvocation *invocation = [self invocationForInstanceMethodWithSelector:selector
                                                                          example:example];
         [invocations addObject:invocation];
@@ -100,7 +103,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
 
  @return The selector of the newly defined instance method.
  */
-+ (SEL)addInstanceMethodForExample:(Example *)example {
++ (SEL)addInstanceMethodForExample:(Example *)example classSelectorNames:(NSMutableSet<NSString*> *)selectorNames {
     IMP implementation = imp_implementationWithBlock(^(QuickSpec *self){
         currentSpec = self;
         [example run];
@@ -119,12 +122,6 @@ const void * const QCKExampleKey = &QCKExampleKey;
     NSString *originalName = example.name.qck_selectorName;
     NSString *selectorName = originalName;
     NSUInteger i = 2;
-    
-    static NSMutableSet<NSString *> *selectorNames;
-    static dispatch_once_t onceToken;
-    dispatch_once(&onceToken, ^{
-        selectorNames = [[NSMutableSet alloc] init];
-    });
     
     while ([selectorNames containsObject:selectorName]) {
         selectorName = [NSString stringWithFormat:@"%@_%tu", originalName, i++];
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -5,7 +5,6 @@
 #import <objc/runtime.h>
 
 static QuickSpec *currentSpec = nil;
-static NSMutableSet<NSString *> *selectorNames;
 
 const void * const QCKExampleKey = &QCKExampleKey;
 
@@ -32,8 +31,6 @@ const void * const QCKExampleKey = &QCKExampleKey;
     world.currentExampleGroup = [world rootExampleGroupForSpecClass:[self class]];
     QuickSpec *spec = [self new];
 
-    selectorNames = [[NSMutableSet alloc] init];
-    
     @try {
         [spec spec];
     }
@@ -122,6 +119,12 @@ const void * const QCKExampleKey = &QCKExampleKey;
     NSString *originalName = example.name.qck_selectorName;
     NSString *selectorName = originalName;
     unsigned int i = 2;
+    
+    static NSMutableSet<NSString *> *selectorNames;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        selectorNames = [[NSMutableSet alloc] init];
+    });
     
     while ([selectorNames containsObject:selectorName]) {
         selectorName = [NSString stringWithFormat:@"%@_%u", originalName, i++];
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -127,7 +127,7 @@ const void * const QCKExampleKey = &QCKExampleKey;
     });
     
     while ([selectorNames containsObject:selectorName]) {
-        selectorName = [NSString stringWithFormat:@"%@_%lu", originalName, (unsigned long)i++];
+        selectorName = [NSString stringWithFormat:@"%@_%tu", originalName, i++];
     }
     
     [selectorNames addObject:selectorName];
--- a/Sources/Quick/QuickSpec.m
+++ b/Sources/Quick/QuickSpec.m
@@ -118,11 +118,13 @@ const void * const QCKExampleKey = &QCKExampleKey;
     }
 
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
-    NSString *selectorName = example.name.qck_selectorName;
-    unsigned long i = 2;
+    
+    NSString *originalName = example.name.qck_selectorName;
+    NSString *selectorName = originalName;
+    unsigned int i = 2;
     
     while ([selectorNames containsObject:selectorName]) {
-        selectorName = [NSString stringWithFormat:@"%@_%lu", example.name.qck_selectorName, i++];
+        selectorName = [NSString stringWithFormat:@"%@_%u", originalName, i++];
     }
     
     [selectorNames addObject:selectorName];
--- a/Sources/QuickTests/FunctionalTests/ItTests.swift
+++ b/Sources/QuickTests/FunctionalTests/ItTests.swift
@@ -14,8 +14,8 @@ class FunctionalTests_ItSpec: QuickSpec {
             expect(exampleMetadata!.example.name).to(equal(""))
         }
 
-        it("has a description with セレクター名に使えない文字が入っている ��") {
-            let name = "has a description with セレクター名に使えない文字が入っている ��"
+        it("has a description with セレクター名に使えない文字が入っている 👊💥") {
+            let name = "has a description with セレクター名に使えない文字が入っている 👊💥"
             expect(exampleMetadata!.example.name).to(equal(name))
         }
 
--- a/Sources/QuickTests/FunctionalTests/ItTests.swift
+++ b/Sources/QuickTests/FunctionalTests/ItTests.swift
@@ -18,6 +18,13 @@ class FunctionalTests_ItSpec: QuickSpec {
             let name = "has a description with セレクター名に使えない文字が入っている 👊💥"
             expect(exampleMetadata!.example.name).to(equal(name))
         }
+        
+        it("is a test with a unique name") {
+            let allSelectors: Set<String> = FunctionalTests_ItSpec.allSelectors();
+            
+            expect(allSelectors).to(contain("is_a_test_with_a_unique_name"));
+            expect(allSelectors).toNot(contain("is_a_test_with_a_unique_name_2"));
+        }
     }
 }
 
@@ -30,6 +37,6 @@ class ItTests: XCTestCase, XCTestCaseProvider {
 
     func testAllExamplesAreExecuted() {
         let result = qck_runSpec(FunctionalTests_ItSpec.self)
-        XCTAssertEqual(result.executionCount, 2 as UInt)
+        XCTAssertEqual(result.executionCount, 3 as UInt)
     }
 }
--- a/Sources/QuickTests/FunctionalTests/ItTests.swift
+++ b/Sources/QuickTests/FunctionalTests/ItTests.swift
@@ -20,11 +20,42 @@ class FunctionalTests_ItSpec: QuickSpec {
         }
 
 #if _runtime(_ObjC)
-        it("is a test with a unique name") {
-            let allSelectors: Set<String> = FunctionalTests_ItSpec.allSelectors();
+
+        describe("when an example has a unique name"){
+            it("has a unique name") {}
             
-            expect(allSelectors).to(contain("is_a_test_with_a_unique_name"));
-            expect(allSelectors).toNot(contain("is_a_test_with_a_unique_name_2"));
+            it("doesn't add multiple selectors for it") {
+                let allSelectors = [String](
+                    FunctionalTests_ItSpec.allSelectors()
+                        .filter { $0.hasPrefix("when_an_example_has_a_unique_name__") }
+                    )
+                    .sort(<)
+                
+                expect(allSelectors) == [
+                    "when_an_example_has_a_unique_name__doesn_t_add_multiple_selectors_for_it",
+                    "when_an_example_has_a_unique_name__has_a_unique_name"
+                ]
+            }
+        }
+    
+        describe("when two examples have the exact name") {
+            it("has exactly the same name") {}
+            it("has exactly the same name") {}
+            
+            it("makes a unique name for each of the above") {
+                let allSelectors = [String](
+                    FunctionalTests_ItSpec.allSelectors()
+                        .filter { $0.hasPrefix("when_two_examples_have_the_exact_name__") }
+                    )
+                    .sort(<)
+                
+                expect(allSelectors) == [
+                    "when_two_examples_have_the_exact_name__has_exactly_the_same_name",
+                    "when_two_examples_have_the_exact_name__has_exactly_the_same_name_2",
+                    "when_two_examples_have_the_exact_name__makes_a_unique_name_for_each_of_the_above"
+                ]
+            }
+            
         }
 
         describe("error handling when misusing ordering") {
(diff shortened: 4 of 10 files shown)


abbeycode:
I'm not sure why the integrations are failing. It doesn't look like it's due to my changes. There are build warnings, but the tests all pass. What am I missing?

modocache:
Sorry, @abbeycode, it looks like your changes don't play nicely with `xctool`. Could you try running the [`xctool` rake task](https://github.com/Quick/Quick/blob/11167ab8d6db66ced958a3ea094e4f6440cf995e/Rakefile#L20-L32)?

```
$ which xctool || brew install xctool
$ rake test:xctool
```

modocache:
Oh, and before I forget: this is an awesome improvement. Thanks so much for taking it on!! :100:

abbeycode:
You're welcome. I see the rake failure locally, but I also see it when I build from `master`...

abbeycode:
I'm on Xcode 7.2, and xctool 0.2.7. Should I be using older versions?

modocache:
> I'm on Xcode 7.2, and xctool 0.2.7. Should I be using older versions?

Oh, good point. I'm not sure off the top of my head which version of `xctool` Travis uses. I'd recommend adding a line to the Rake task that prints out the current `xctool` version. We should have the CI use the latest version possible.

abbeycode:
Okay, Circle and Travis are both using 0.2.7, so that's ruled out...

When I build the `Quick-iOS` scheme in Xcode, I see a couple of warnings. For the test bundles:

> ld: warning: directory not found for option '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.2.sdk/Developer/Library/Frameworks'

Indeed, I see that the Frameworks directory doesn't exist there. I cleared out the `Framework Search Paths` for the iOS and tvOS test targets, and those warnings go away (though still the rake fails).

That leaves a warning for the iOS framework targets:

> ld: warning: embedded dylibs/frameworks only run on iOS 8 or later

I updated the Quick and Nimble iOS framework targets for a deployment target of iOS 8.0, and cleared those warnings. So now, in Xcode, I get _no_ visible warnings, and all unit tests pass. Still, the rake fails.

I combed through the xctool output, and it looks like there's an exception with this message (again, this is in `master`):

```
[2m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━[0m
[2mFailed to query the list of test cases in the test bundle: Error while parsing JSON: The data is not in the correct format.: {
    stderr = "";
    stdout = "Pending: an example that will not run\nPending: an example that will not run\nPending: an example that will not run\nPending: an example that will not run\nPending: an example that will not run\n[\"AfterEachTests\\/testAfterEachIsExecutedInTheCorrectOrder\",\"AfterEachTests_ObjC\\/testAfterEachIsExecutedInTheCorrectOrder\",\"AfterSuiteTests\\/testAfterSuiteIsNotExecutedBeforeAnyExamples\",\"AfterSuiteTests_ObjC\\/testAfterSuiteIsNotExecutedBeforeAnyExamples\",\"BeforeEachTests\\/testBeforeEachIsExecutedInTheCorrectOrder\",\"BeforeEachTests_ObjC\\/testBeforeEachIsExecutedInTheCorrectOrder\",\"BeforeSuiteTests\\/testBeforeSuiteIsExecutedBeforeAnyExamples\",\"BeforeSuiteTests_ObjC\\/testBeforeSuiteIsExecutedBeforeAnyExamples\",\"Configuration_AfterEachSpec\\/is_executed_before_the_configuration_afterEach_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsConfigurationAfterEachConfigurationAfterEachTestsswift_13\",\"Configuration_AfterEachTests\\/testExampleIsRunAfterTheConfigurationBeforeEachIsExecuted\",\"Configuration_BeforeEachSpec\\/is_executed_after_the_configuration_beforeEach_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsConfigurationBeforeEachConfigurationBeforeEachTestsswift_10\",\"Configuration_BeforeEachTests\\/testExampleIsRunAfterTheConfigurationBeforeEachIsExecuted\",\"DescribeTests\\/testDescribeThrowsIfUsedOutsideOfQuickSpec\",\"FailureTests_ObjC\\/testFailureSpecExecutedAllExamples\",\"FailureTests_ObjC\\/testFailureSpecFailureCountIsEqualToTheNumberOfFailingExamples\",\"FailureTests_ObjC\\/testFailureSpecHasSucceededIsFalse\",\"FailureUsingXCTAssertTests_ObjC\\/testFailureUsingXCTAssertSpecExecutedAllExamples\",\"FailureUsingXCTAssertTests_ObjC\\/testFailureUsingXCTAssertSpecFailureCountIsEqualToTheNumberOfFailingExamples\",\"FailureUsingXCTAssertTests_ObjC\\/testFailureUsingXCTAssertSpecHasSucceededIsFalse\",\"FunctionalTests_AfterEachSpec\\/executes_the_outer_afterEach_closures_a_second_time__but_not_before_this_closure__2__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterEachTestsswift_31\",\"FunctionalTests_AfterEachSpec\\/executes_the_outer_afterEach_closures_once__but_not_before_this_closure__1__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterEachTestsswift_25\",\"FunctionalTests_AfterEachSpec\\/when_there_are_nested_afterEach__executes_the_outer_and_inner_afterEach_closures__but_not_before_this_closure__3__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterEachTestsswift_41\",\"FunctionalTests_AfterEachSpec_ObjC\\/executes_the_outer_afterEach_closures_a_second_time__but_not_before_this_closure__2__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterEachTestsObjCm_28\",\"FunctionalTests_AfterEachSpec_ObjC\\/executes_the_outer_afterEach_closures_once__but_not_before_this_closure__1__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterEachTestsObjCm_24\",\"FunctionalTests_AfterEachSpec_ObjC\\/when_there_are_nested_afterEach__executes_the_outer_and_inner_afterEach_closures__but_not_before_this_closure__3__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterEachTestsObjCm_36\",\"FunctionalTests_AfterSuite_Spec\\/is_executed_before_afterSuite_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterSuiteTestsswift_20\",\"FunctionalTests_AfterSuite_Spec_ObjC\\/is_executed_before_afterSuite_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsAfterSuiteTestsObjCm_19\",\"FunctionalTests_BeforeEachSpec\\/executes_the_outer_beforeEach_closures_a_second_time__2__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeEachTestsswift_25\",\"FunctionalTests_BeforeEachSpec\\/executes_the_outer_beforeEach_closures_once__1__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeEachTestsswift_24\",\"FunctionalTests_BeforeEachSpec\\/when_there_are_nested_beforeEach__executes_the_outer_and_inner_beforeEach_closures__3__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeEachTestsswift_32\",\"FunctionalTests_BeforeEachSpec_ObjC\\/executes_the_outer_beforeEach_closures_a_second_time__2__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeEachTestsObjCm_23\",\"FunctionalTests_BeforeEachSpec_ObjC\\/executes_the_outer_beforeEach_closures_once__1__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeEachTestsObjCm_22\",\"FunctionalTests_BeforeEachSpec_ObjC\\/when_there_are_nested_beforeEach__executes_the_outer_and_inner_beforeEach_closures__3__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeEachTestsObjCm_30\",\"FunctionalTests_BeforeSuite_Spec\\/is_executed_after_beforeSuite_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeSuiteTestsswift_20\",\"FunctionalTests_BeforeSuite_Spec_ObjC\\/is_executed_after_beforeSuite_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsBeforeSuiteTestsObjCm_19\",\"FunctionalTests_FailureSpec_ObjC\\/a_group_of_failing_examples__fails__but_only_when_running_the_functional_tests__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsFailureTestsObjCm_19\",\"FunctionalTests_FailureSpec_ObjC\\/a_group_of_failing_examples__fails_again__but_only_when_running_the_functional_tests__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsFailureTestsObjCm_23\",\"FunctionalTests_FailureSpec_ObjC\\/a_group_of_failing_examples__passes_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsFailureTestsObjCm_15\",\"FunctionalTests_FailureUsingXCTAssertSpec_ObjC\\/fails_again_using_an_XCTAssert__but_only_when_running_the_functional_tests__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsFailureUsingXCTAssertTestsObjCm_14\",\"FunctionalTests_FailureUsingXCTAssertSpec_ObjC\\/fails_using_an_XCTAssert__but_only_when_running_the_functional_tests__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsFailureUsingXCTAssertTestsObjCm_10\",\"FunctionalTests_FailureUsingXCTAssertSpec_ObjC\\/succeeds_using_an_XCTAssert_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsFailureUsingXCTAssertTestsObjCm_18\",\"FunctionalTests_ItSpec\\/_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsItTestsswift_13\",\"FunctionalTests_ItSpec\\/has_a_description_with_\U30bb\U30ec\U30af\U30bf\U30fc\U540d\U306b\U4f7f\U3048\U306a\U3044\U6587\U5b57\U304c\U5165\U3063\U3066\U3044\U308b____UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsItTestsswift_17\",\"FunctionalTests_ItSpec_ObjC\\/__UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsItTestsObjCm_14\",\"FunctionalTests_ItSpec_ObjC\\/has_a_description_with_\U30bb\U30ec\U30af\U30bf\U30fc\U540d\U306b\U4f7f\U3048\U306a\U3044\U6587\U5b57\U304c\U5165\U3063\U3066\U3044\U308b____UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsItTestsObjCm_18\",\"FunctionalTests_PendingSpec\\/a_describe_block_containing_only_one_enabled_example__an_example_that_will_run_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsPendingTestsswift_19\",\"FunctionalTests_PendingSpec_ObjC\\/a_describe_block_containing_only_one_enabled_example__an_example_that_will_run_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsPendingTestsObjCm_18\",\"FunctionalTests_SharedExamples_BeforeEachSpec\\/a_group_of_three_shared_examples_with_a_beforeEach__passes_once_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsswift_26\",\"FunctionalTests_SharedExamples_BeforeEachSpec\\/a_group_of_three_shared_examples_with_a_beforeEach__passes_three_times_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsswift_26\",\"FunctionalTests_SharedExamples_BeforeEachSpec\\/a_group_of_three_shared_examples_with_a_beforeEach__passes_twice_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsswift_26\",\"FunctionalTests_SharedExamples_BeforeEachSpec\\/executes_the_spec_beforeEach_once_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsswift_25\",\"FunctionalTests_SharedExamples_BeforeEachSpec_ObjC\\/a_group_of_three_shared_examples_with_a_beforeEach_in_Obj_C__passes_once_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsObjCm_28\",\"FunctionalTests_SharedExamples_BeforeEachSpec_ObjC\\/a_group_of_three_shared_examples_with_a_beforeEach_in_Obj_C__passes_three_times_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsObjCm_28\",\"FunctionalTests_SharedExamples_BeforeEachSpec_ObjC\\/a_group_of_three_shared_examples_with_a_beforeEach_in_Obj_C__passes_twice_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsObjCm_28\",\"FunctionalTests_SharedExamples_BeforeEachSpec_ObjC\\/executes_the_spec_beforeEach_once_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesBeforeEachTestsObjCm_27\",\"FunctionalTests_SharedExamples_ContextSpec\\/shared_examples_that_take_a_context__is_passed_the_correct_parameters_via_the_context_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsswift_17\",\"FunctionalTests_SharedExamples_ContextSpec_ObjC\\/shared_examples_that_take_a_context__is_passed_the_correct_parameters_via_the_context_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsObjCm_15\",\"FunctionalTests_SharedExamples_SameContextSpec_ObjC\\/gets_called_with_a_different_context_from_within_the_same_spec_file__tracks_correctly_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsObjCm_39\",\"FunctionalTests_SharedExamples_SameContextSpec_ObjC\\/gets_called_with_a_different_context_from_within_the_same_spec_file__tracks_correctly_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsObjCm_43\",\"FunctionalTests_SharedExamples_Spec\\/a_group_of_three_shared_examples__passes_once_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsswift_11\",\"FunctionalTests_SharedExamples_Spec\\/a_group_of_three_shared_examples__passes_three_times_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsswift_11\",\"FunctionalTests_SharedExamples_Spec\\/a_group_of_three_shared_examples__passes_twice_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsswift_11\",\"FunctionalTests_SharedExamples_Spec_ObjC\\/a_group_of_three_shared_examples__passes_once_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsObjCm_9\",\"FunctionalTests_SharedExamples_Spec_ObjC\\/a_group_of_three_shared_examples__passes_three_times_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsObjCm_9\",\"FunctionalTests_SharedExamples_Spec_ObjC\\/a_group_of_three_shared_examples__passes_twice_UsersDovSourceCodeQuickSourcesQuickTestsFunctionalTestsSharedExamplesTestsObjCm_9\",\"ItTests\\/testAllExamplesAreExecuted\",\"ItTests_ObjC\\/testAllExamplesAreExecuted\",\"PendingTests\\/testAnOtherwiseFailingExampleWhenMarkedPendingDoesNotCauseTheSuiteToFail\",\"PendingTests\\/testBeforeEachDoesNotRunForContextsWithOnlyPendingExamples\",\"PendingTests\\/testBeforeEachOnlyRunForEnabledExamples\",\"PendingTests_ObjC\\/testAnOtherwiseFailingExampleWhenMarkedPendingDoesNotCauseTheSuiteToFail\",\"PendingTests_ObjC\\/testBeforeEachDoesNotRunForContextsWithOnlyPendingExamples\",\"PendingTests_ObjC\\/testBeforeEachOnlyRunForEnabledExamples\",\"QuickConfigurationTests\\/testInitThrows\",\"SharedExamplesTests\\/testAGroupOfThreeSharedExamplesExecutesThreeExamples\",\"SharedExamplesTests\\/testSharedExamplesWithContextPassContextToExamples\",\"SharedExamplesTests_ObjC\\/testAGroupOfThreeSharedExamplesExecutesThreeExamples\",\"SharedExamplesTests_ObjC\\/testSharedExamplesWithContextPassContextToExamples\",\"SharedExamples_BeforeEachTests\\/testBeforeEachInSharedExamplesExecutedOnceBeforeEachSharedExample\",\"SharedExamples_BeforeEachTests\\/testBeforeEachOutsideOfSharedExamplesExecutedOnceBeforeEachExample\",\"SharedExamples_BeforeEachTests_ObjC\\/testBeforeEachInSharedExamplesExecutedOnceBeforeEachSharedExample\",\"SharedExamples_BeforeEachTests_ObjC\\/testBeforeEachOutsideOfSharedExamplesExecutedOnceBeforeEachExample\"]";
} [0m
[2m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━[0m
```

modocache:
Funny. I suppose `xctool` doesn't know how to parse our `pending` example messages. Still, this isn't a result of your changes. The culprit must be hiding somewhere else...

You may need to search the `xctool` codebase for the messages `Can't find test state for` and `Test name seems to be malformed:`.

abbeycode:
Alright, well let's put aside the errors I'm seeing locally for OS X tests, for the time being, and look at what I'm currently seeing in the CI environments.

I'm getting the same `xctool` exception for Circle, Travis, and locally, but while running different test cases: 

> 2016-01-21 12:54:26.438 xctool[1945:9678] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Can't find test state for '-[FunctionalTests_PendingSpec_ObjC a_describe_block_containing_only_one_enabled_example__an_example_that_will_run]', check senTestList'

A bunch of tests run, but then it fails for one, and not even necessarily one that my new code appended to for uniqueness (as shown above). Have you seen this before while developing Quick? It's not so easy to trace through what `xctool` is doing here.

I think what I'm missing, in trying to diagnose this, is how exactly Quick generates metadata to communicate to Xcode or `xctool` what tests and test classes exist. I seem to have broken that.

briancroom:
@abbeycode I don't have much to contribute here about the xctool failure, but I did notice this issue which seems like it could be connected somehow: #401

abbeycode:
@briancroom Thanks, it definitely is because of duplicate selectors, as in the issue you linked. I verified by appending a global increasing int to the end of each selector, and the tests succeeded. Now I need to figure out how my code is generating non-unique names.

abbeycode:
Yay! This PR is ready for review, when someone has a chance.

Rakefile
```diff
@@ -21,6 +21,7 @@ namespace "test" do
   task xctool: %w[test:xctool:ios test:xctool:osx]
   namespace :xctool do
     desc "Run unit tests for all iOS targets using xctool"
+    run "echo Using xctool v`xctool -v`..."
```
modocache: This'll be printed out every time the Rakefile is used, since it's not in a `do ... end` closure.

This could either be copy-pasted into the `test:xctool:ios` and `test:xctool:osx` tasks, or you could put this into its own task, like `test:xctool:version`, and have the `test:xctool:ios` and `test:xctool:osx` tasks dependent upon that task.

Sources/Quick/QuickSpec.m
```diff
@@ -115,10 +115,23 @@ + (SEL)addInstanceMethodForExample:(Example *)example {
...
 
     const char *types = [[NSString stringWithFormat:@"%s%s%s", @encode(id), @encode(id), @encode(SEL)] UTF8String];
-    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld",
-                              example.name.qck_selectorName,
-                              sanitizedFileName,
-                              (long)example.callsite.line];
+    
+    NSString *originalName = example.name.qck_selectorName;
+    NSString *selectorName = originalName;
+    unsigned int i = 2;
```
modocache: nit-pick: Very silly, but I'd prefer to use `NSUInteger` here. :stuck_out_tongue_closed_eyes: 

I prefer to always use `NSInteger` and `NSUInteger` over `int` and `unsigned int`. [There are some good reasons to do so](http://stackoverflow.com/a/4445199/679254), although those are probably not relevant here.

Sources/Quick/QuickSpec.m
```diff
@@ -115,10 +115,23 @@ + (SEL)addInstanceMethodForExample:(Example *)example {
...
+    
+    static NSMutableSet<NSString *> *selectorNames;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        selectorNames = [[NSMutableSet alloc] init];
+    });
+    
+    while ([selectorNames containsObject:selectorName]) {
+        selectorName = [NSString stringWithFormat:@"%@_%u", originalName, i++];
+    }
```
modocache: This should probably check whether a number needs to be appended to the end of the selector at all. Check this out:

screen shot 2016-01-23 at 9 59 03 pm

Ideally if there are no duplicate names, then the selector stays the same. So in the above example, the selector name should remain `is_a_test`.

Sources/Quick/QuickSpec.m
```diff
@@ -115,10 +115,23 @@ + (SEL)addInstanceMethodForExample:(Example *)example {
...
-    NSString *selectorName = [NSString stringWithFormat:@"%@_%@_%ld",
-                              example.name.qck_selectorName,
-                              sanitizedFileName,
-                              (long)example.callsite.line];
+    
+    NSString *originalName = example.name.qck_selectorName;
+    NSString *selectorName = originalName;
+    unsigned int i = 2;
+    
+    static NSMutableSet<NSString *> *selectorNames;
```
modocache: I'd like to make this more prominent, so it's obvious that we're maintaining global state. Is it possible to move this up to where `currentSpec` is defined, at the top of the file? Alternatively, we could move this into it's own singleton class--think `[QCKSelectorNameStore sharedStore]`, or something.

modocache:
Awesome, thanks! I had a few comments, but looks good so far!

Sources/Quick/QuickSpec.m
abbeycode: I'd used it in my first commit, but changed the variable type to silence a compiler warning. I updated to instead cast to `unsigned long` in the `stringWithFormat:` call.

modocache: If you're referring to warnings when converting `NSUInteger` to `%lu`, I think the best solution is using `%tu` (for `NSUInteger`) or `%zu` (for `NSInteger`). That way you can use string formatting without a cast! :+1:

Sources/Quick/QuickSpec.m
abbeycode: Hmm, I'll need to investigate what's going on there. How can I create a test within Quick to test the generated selector name?

Sources/Quick/QuickSpec.m
```diff
@@ -115,10 +115,23 @@ + (SEL)addInstanceMethodForExample:(Example *)example {
...
+    NSUInteger i = 2;
+    
+    static NSMutableSet<NSString *> *selectorNames;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        selectorNames = [[NSMutableSet alloc] init];
+    });
+    
+    while ([selectorNames containsObject:selectorName]) {
+        selectorName = [NSString stringWithFormat:@"%@_%lu", originalName, (unsigned long)i++];
```
modocache: You can use `%tu` here, and avoid the cast! :sunglasses:

abbeycode: Nice, I hadn't seen that before. Strange that Xcode doesn't suggest that as a Fix-It.

modocache: Yeah! I bet that'd be a nice patch to send to Clang... :wink:

Sources/Quick/QuickSpec.m
abbeycode: Bump, just in case you missed my comment above amidst the others.

modocache: Thanks for the bump, I indeed hadn't noticed. You may be able to test the selector name using example metadata, as in this example: https://github.com/Quick/Quick/blob/662950f3e14329ce7569f940e4cce17b93928e4a/Sources/QuickTests/FunctionalTests/ItTests.swift#L17-L20

If that doesn't work, perhaps we should think about testing the test output, as in https://github.com/apple/swift-corelibs-xctest/pull/20. It doesn't have to be in this pull request, though--if we can't test the selector in an automated way, so be it.

abbeycode: 1. The Example class doesn't presently track the selector used for the Example. I tried adding this capability, but was having trouble assigning the selector. I declared the `var` with type `Selector` in the Example class, but it doesn't come through into the QuickSpec.m file where I would have assigned it.

2. Yeah, it would be great to test the generated selector names directly. For now, I'm addressing this with an internal `+[QuickSpec allSelectors]` method that returns an immutable copy of the global selector set, and test expectations against that.

3. Using the test described above, I verified that I'm getting two selectors for a single example. It looks like this is happening because `+[QuickSpec testInvocations]` is called twice: once in `+initialize` and again in `-[XCTestSuite testSuiteForTestCaseClass:]`. I tried removing the call in `+initialize` and didn't noticed any ill-effects (the tests passed). Is it there for backwards-compatibility with an older Xcode version, or something like that?

abbeycode:
Alright, I believe I have addressed all open comments. I changed the list of selector names so it's no longer global, but instead generated for each class as it iterates through its examples. I also added a rudimentary form of selector checking (I added a category on `QuickSpec` that lists all of its selectors). The test cases I added verify that there are no `_2` selectors generated.

abbeycode:
Nuts. And then the build fails for Linux. I'm not sure what to do about that - the tests are failing to build now because I'm calling a selector through the bridging header.

abbeycode:
Fixed it!

abbeycode:
@modocache Can you take a look at my updates when you have a chance?

modocache:
Yup, I'm on it. Busy with some other stuff, but sometime this week for sure. :+1:

Sources/Quick/QuickSpec.m
```diff
@@ -115,10 +118,17 @@ + (SEL)addInstanceMethodForExample:(Example *)example {
...
+    
+    NSString *originalName = example.name.qck_selectorName;
+    NSString *selectorName = originalName;
+    NSUInteger i = 2;
+    
+    while ([selectorNames containsObject:selectorName]) {
+        selectorName = [NSString stringWithFormat:@"%@_%tu", originalName, i++];
+    }
+    
+    [selectorNames addObject:selectorName];
```
modocache: Awesome! This is great. :+1:

Sources/Quick/NSString+QCKSelectorName.m
```diff
@@ -27,7 +27,11 @@ - (NSString *)qck_selectorName {
 
     NSArray *validComponents = [self componentsSeparatedByCharactersInSet:invalidCharacters];
 
-    return [validComponents componentsJoinedByString:@"_"];
+    NSString *result = [validComponents componentsJoinedByString:@"_"];
+    
+    return ([result length] == 0
+            ? @"_"
+            : result);
```
modocache: Interesting... I guess without this additional code, `it("")` would have a selector name of `""`, is that right? What sort of behavior was that causing? Would `it("")` examples never be run?

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
```diff
@@ -0,0 +1,23 @@
...
+
+@implementation QuickSpec (QuickSpec_MethodList)
+
++ (NSSet<NSString*> *)allSelectors {
+    id t = [[[self class] alloc] init];
+    NSMutableSet<NSString*> *allSelectors = [NSMutableSet set];
+    
+    unsigned int methodCount = 0;
+    Method * mlist = class_copyMethodList(object_getClass(t), &methodCount);
+    NSLog(@"%d methods", methodCount);
```
modocache: Please remove this `NSLog`, and the one below. If the information is relevant, we should have an automated test for it--which we do, in this case! :+1:

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
```diff
@@ -0,0 +1,23 @@
+#import "QuickSpec+QuickSpec_MethodList.h"
+#import <objc/runtime.h>
+
+
+@implementation QuickSpec (QuickSpec_MethodList)
+
++ (NSSet<NSString*> *)allSelectors {
+    id t = [[[self class] alloc] init];
```
modocache: Two nit-picks: Let's come up with a better variable name than `t` here. How about `specInstance`? Also, we know that this will be an instance of a subclass of `QuickSpec *`, so let's type it as such. In other words, this line should be `QuickSpec *specInstance = ...`.

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
```diff
@@ -0,0 +1,23 @@
...
+    NSMutableSet<NSString*> *allSelectors = [NSMutableSet set];
+    
+    unsigned int methodCount = 0;
+    Method * mlist = class_copyMethodList(object_getClass(t), &methodCount);
```
modocache: Nit pick: Remove the space between `Method *` and `mlist`. It's more consistent with the spacing in the rest of this file, and the rest of the codebase.

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
```diff
@@ -0,0 +1,23 @@
...
+    NSLog(@"%d methods", methodCount);
+    for(int i = 0; i<methodCount; i++) {
```
modocache: Nit pick: How about an `unsigned int` for the index here? It matches the type of `methodCount` that way. Also, add a space in between `i`, `<`, and `methodCount` -- as in `i < methodCount`.

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.h
```diff
@@ -0,0 +1,7 @@
+#import <Quick/Quick.h>
+
+@interface QuickSpec (QuickSpec_MethodList)
+
++ (NSSet<NSString*> *)allSelectors;
```
modocache: Add documentation to this method. The documentation should make it clear that:

1. An instance of the class will be instantiated in order to get a method list.
2. This method should only be used for test assertions.

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.h
```diff
@@ -0,0 +1,7 @@
...
+
+@end
```
modocache: Wrap this header in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`, to improve its bridging to Swift.

Sources/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -14,12 +14,19 @@ class FunctionalTests_ItSpec: QuickSpec {
             expect(exampleMetadata!.example.name).to(equal(""))
         }
 
-        it("has a description with セレクター名に使えない文字が入っている 👊💥") {
-            let name = "has a description with セレクター名に使えない文字が入っている 👊💥"
+        it("has a description with セレクター名に使えない文字が入っている ��") {
+            let name = "has a description with セレクター名に使えない文字が入っている ��"
```
modocache: Why are these emoji no longer displaying here? Could you revert this change, assuming you made one?

Sources/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -14,12 +14,19 @@ class FunctionalTests_ItSpec: QuickSpec {
...
         }
 
 #if _runtime(_ObjC)
+        it("is a test with a unique name") {
+            let allSelectors: Set<String> = FunctionalTests_ItSpec.allSelectors();
+            
+            expect(allSelectors).to(contain("is_a_test_with_a_unique_name"));
+            expect(allSelectors).toNot(contain("is_a_test_with_a_unique_name_2"));
+        }
+
```
modocache: Hmm, speaking of which: does this problem also exist on Quick for Linux? @briancroom, do you happen to know off the top of your head what happens when two examples have the same name when running `QCKMain()`?

Sources/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -14,12 +14,19 @@ class FunctionalTests_ItSpec: QuickSpec {
...
+            let name = "has a description with セレクター名に使えない文字が入っている ��"
             expect(exampleMetadata!.example.name).to(equal(name))
         }
 
 #if _runtime(_ObjC)
+        it("is a test with a unique name") {
+            let allSelectors: Set<String> = FunctionalTests_ItSpec.allSelectors();
+            
+            expect(allSelectors).to(contain("is_a_test_with_a_unique_name"));
+            expect(allSelectors).toNot(contain("is_a_test_with_a_unique_name_2"));
```
modocache: This is very close, but I don't think this test is enough. I'd like a test like the following:

```swift
describe("when two examples have the exact name name") {
  it("exact same name") {}
  it("exact same name") {}

  it("uses a unique name for each") {
    expect(allSelectors.allObjects).to(equal("exact_same_name", "exact_same_name_2"));
  }
}
```

modocache:
This is looking great, @abbeycode! I had a request for the test, and a bunch of minor nit picks.

Sources/Quick/NSString+QCKSelectorName.m
abbeycode: It was causing xctool errors because the selector was like `[TestClass ]`

Sources/QuickTests/FunctionalTests/ItTests.swift
briancroom: As of now (and still as of the latest version of https://github.com/apple/swift-corelibs-xctest/pull/40), the tests are presented as an array, and the string component is only actually used for display purposes, so a duplicate name shouldn't cause any issues.

abbeycode:
Alright, I believe I've addressed all open concerns, and everything's still up-to-date with `master`.

Sources/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
```diff
@@ -0,0 +1,30 @@
...
+ *  @warning Only intended to be used in test assertions!
+ *
+ *  @return a set of NSStrings representing the list of selectors it contains
+ */
++ (NSSet<NSString*> *)allSelectors {
+    QuickSpec *specInstance = [[[self class] alloc] init];
+    NSMutableSet<NSString*> *allSelectors = [NSMutableSet set];
+    
+    unsigned int methodCount = 0;
+    Method *mlist = class_copyMethodList(object_getClass(specInstance), &methodCount);
```
jeffh: Memory leak: `mlist` needs to be freed.

abbeycode: D'oh! ARC is starting to make me lazy. Thanks.

modocache:
Excellent! Thanks for all the hard work, @abbeycode. I'm very excited to see this improvement; I'll release a new version soon.

Issue #463, opened 2016-01-21
Fix podspec


modocache:
Perhaps broken in #434...?

modocache:
Fixed in #464.

PR #464, opened 2016-01-21
Update podspec for the new source paths

briancroom:
This addresses #463.

Note that I haven't had a chance to properly verify this yet, so that needs to happen before merging please!


modocache:
Thanks, @briancroom! :+1:

PR #469, opened 2016-01-26
Return only valid phase strings within currentPhase helper function

jwfriese:
@modocache 
This is a PR to address the concerns you raised when you merged #448. Thanks as always for the feedback.


Sources/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -9,16 +9,16 @@ class FunctionalTests_ItSpec: QuickSpec {
     override func spec() {
         var exampleMetadata: ExampleMetadata?
         beforeEach { metadata in exampleMetadata = metadata }
-
```
jwfriese: My bad on all the whitespace in this file... :sweat_smile: 

I fixed it though! :smile:

modocache:
Great! Thanks for following up, @jwfriese! :bouquet:

PR #474, opened 2016-02-04
Linux update

briancroom:
Make updates to support the newest Swift development snapshot toolchain.


Sources/Quick/QuickMain.swift
```diff
@@ -3,6 +3,13 @@ import XCTest
 // NOTE: This file is not intended to be included in the Xcode project or CocoaPods.
 //       It is picked up by the Swift Package Manager during its build process.
 
+/// When using Quick with swift-corelibs-xctest, automatic discovery of specs and
+/// configurations is not available. Instead, you should create a standalone
+/// executable and call this function from its main.swift file. This will execute
+/// the specs and then terminate the process with an exit code of 0 if the tests
+/// passed, or 1 if there were any failures.
+///
+/// Quick is known to work with the DEVELOPMENT-SNAPSHOT-2016-02-03-a Swift toolchain
```
modocache: Interesting!

1. nit-pick: This sentence should end with a period.
2. I'm on the fence when it comes to including this as documentation here; we'll have to update it constantly. It is very useful to end-users, though!

modocache:
Awesome, thanks @briancroom! :+1: 

I had a few nit-picks, but feel free to push commits addressing them directly to master.

Sources/Quick/QuickMain.swift
briancroom: Yeah..I'm not 100% convinced either. On an earlier PR you had suggested adding documentation about the Swift version somewhere because just the `.swift-version` file isn't terribly discoverable.

I'm inclined to include it here for a little while at least, while we know that further breaking changes are [in the works](https://github.com/apple/swift-corelibs-xctest/pull/40) and we have no intention of branching while we are targeting these fast-moving development versions. Once things have settled down a bit then we can remove this note, I'd think

Issue #467, opened 2016-01-24
Turn `script/release` into a Rake task


modocache:
Just a personal preference. I find shellscripts harder to debug.

ashfurrow:
We should consider doing this for Nimble, too, since they currently have similar-ish release scripts anyway.

phatblat:
Agreed. It should be easy to copy over to Nimble once #475 is complete.

ikesyo:
Closing this as part of issue cleanup.

Issue #422, opened 2015-11-13
Failed to use carthage (0.10) build Quick framework for tvOS


unchartedworks:
carthage update

*** Checking out Quick at "a98f8e37867462c8f00fea1a7b255d12f5a4b611"

ld: warning: Auto-Linking supplied '/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVOS.platform/Developer/Library/Frameworks/XCTest.framework/XCTest', '/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVOS.platform/Developer/Library/Frameworks/XCTest.framework/XCTest' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE) or obtain an updated library from the vendor
clang: error: linker command failed with exit code 1 (use -v to see invocation)

yoichitgy:
I think this is the same issue as the following issue of Nimble. A workaround is to add `--platform iOS,Mac,watchOS` option to the Carthage command if you don't use tvOS.

https://github.com/Quick/Nimble/issues/213

dearprakash:
:+1: @yoichitgy

phatblat:
[Carthage 0.17](https://github.com/Carthage/Carthage/releases/tag/0.17) is out now which addresses this.

briancroom:
Amazing! Well done @phatblat

c0diq:
The problem still persists with Carthage 0.17.2 and Quick 0.9.2
```

The following build commands failed:
	Ld /Users/srebaud/Library/Developer/Xcode/DerivedData/Quick-ceyoloigrcwecohaakvfitbkjqel/Build/Products/Release-appletvos/Quick.framework/Quick normal arm64
	GenerateDSYMFile /Users/srebaud/Library/Developer/Xcode/DerivedData/Quick-ceyoloigrcwecohaakvfitbkjqel/Build/Products/Release-appletvos/Quick.framework.dSYM /Users/srebaud/Library/Developer/Xcode/DerivedData/Quick-ceyoloigrcwecohaakvfitbkjqel/Build/Products/Release-appletvos/Quick.framework/Quick
(2 failures)
ld: warning: embedded dylibs/frameworks only run on iOS 8 or later
ld: warning: embedded dylibs/frameworks only run on iOS 8 or later
ld: warning: embedded dylibs/frameworks only run on iOS 8 or later
ld: warning: embedded dylibs/frameworks only run on iOS 8 or later
[...]

PR #475, opened 2016-02-06
Create rake task to do releases

jwfriese:
This is an initial step towards #467, calling for a rake task to replace the current bash script to do releases. I say "initial step" because, while this is a big step forward, there is still quite a bit of work to be done, and I'd like to collaborate on it. I will bring attention to the biggest points inline, but please feel free to comment on anything at all that you think could use improvement and I'll follow up quickly. 

At a high level, the task as written worked to push a release build to the `convert_release_script_to_rake_task` branch on my personal fork, with a couple of changes:

1) All the cocoapods trunk pushing code is commented out. I don't know much about how pushing to cocoapods works, but it seems as though my personal cocoapods session doesn't grant me access to update the Quick podspec. I was trying to set up the trunk stuff using this site: https://guides.cocoapods.org/making/getting-setup-with-trunk.html, which has a part about adding maintainers for a podspec. Maybe this is the problem here? 

2) When trying to run the Carthage build process for tvOS, I was running into the linking problems outlined in #422. I got around this by building only iOS and OSX. I'm not sure this is what we want even for the short-term. Guidance here by an experienced Quick maintainer would be much appreciated. 

Again, these are the major things, but please comment on anything that sticks out to you, even if it's a basic ruby best practice.

@modocache, I'd especially love to hear your feedback, as this was your request, and because this conversion will be a good chance for us to make any other changes to the release process that we feel are appropriate. 

Thanks all, looking forward to working this out.

### DIFF
--- a/release.rb
+++ b/release.rb
@@ -1,0 +1,215 @@
+class ReleaseScript
+
+  def self.run(version, release_notes_file_path)
+    run_internal(version, release_notes_file_path, false)
+  end
+
+  def self.run_force(version, release_notes_file_path)
+    run_internal(version, release_notes_file_path, true)
+  end
+
+  private
+
+  def self.run_internal(version, release_notes_file_path, force)
+    remote_branch = 'convert_release_script_to_rake_task'
+    pod_name = 'Quick'
+    pod_spec = 'Quick.podspec'
+    github_tags_url = 'https://github.com/jwfriese/Quick/tags'
+    carthage_framework_name = 'Quick'
+    carthage = ENV["CARTHAGE"] || 'carthage'
+    pod = ENV["COCOAPODS"] || 'pod'
+    version_tag = "v#{version}"
+
+    puts 'Verifying Local Directory for Release'
+    carthage_available = verify_availability(carthage) 
+    die("Carthage is required to produce a release. Aborting.") unless carthage_available
+    puts ' > Carthage is installed'
+
+    pod_available = verify_availability(pod)
+    die(" Cocoapods is required to produce a release. Aborting.") unless pod_available
+    puts ' > Cocoapods is installed'
+
+    puts ' > Is this a reasonable tag?'
+
+    verify_version(version_tag, force)
+
+    release_notes = gather_release_notes(release_notes_file_path, version, pod_name)
+    puts " > Release notes: #{release_notes}"
+
+    die("Cannot find podspec: #{pod_spec}. Aborting.") unless File.file?("#{pod_spec}")
+
+    puts ' > Podspec exists'
+    
+    pgp_found = system("git config --get user.signingkey > /dev/null")
+    unless pgp_found
+      puts '[ERROR] No PGP found to sign tag. Aborting.'
+      puts
+      puts '  Creating a release requires signing the tag for security purposes. This allows users to verify the git cloned tree is from a trusted source.'
+      puts '  From a security perspective, it is not considered safe to trust the commits (including Author & Signed-off fields). It is easy for any'
+      puts '  intermediate between you and the end-users to modify the git repository.'
+      puts
+      puts "  While not all users may choose to verify the PGP key for tagged releases. It is a good measure to ensure 'this is an official release'"
+      puts '  from the official maintainers.'
+      puts
+      puts "  If you're creating your PGP key for the first time, use RSA with at least 4096 bits."
+      puts
+      puts 'Related resources:'
+      puts ' - Configuring your system for PGP: https://git-scm.com/book/tr/v2/Git-Tools-Signing-Your-Work'
+      puts ' - Why: http://programmers.stackexchange.com/questions/212192/what-are-the-advantages-and-disadvantages-of-cryptographically-signing-commits-a'
+      puts
+      return
+    end
+
+    puts " > Found PGP key for git"
+
+    #verify_cocoapods_trunk_ownership(pod_name)
+    #puts " > Verified ownership to #{pod_name} pod"
+
+    puts "--- Releasing version #{version} (tag: #{version_tag})..."
+
+    puts "-> Ensuring no differences to origin/#{remote_branch}"
+    fetched_origin = system("git fetch origin")
+    die("Failed to fetch origin. Aborting") unless fetched_origin
+
+    head_aligned_with_origin = system("git diff --quiet HEAD \"origin/#{remote_branch}\"")
+    die("HEAD is not aligned to origin/#{remote_branch}. Cannot update version safely. Aborting.") unless head_aligned_with_origin
+
+    puts "-> Building Carthage release"
+    built_carthage_framework = system("#{carthage} build --no-skip-current --platform iOS,Mac")
+    die("Failed to build framework for carthage. Aborting.") unless built_carthage_framework
+
+    puts "-> Setting podspec version"
+    set_podspec_version = system("cat #{pod_spec} | grep 's.version' | grep -q \"#{version}\"")
+    if !set_podspec_version
+      updated_podspec_version = system("sed -i.backup \"s/s.version *= *\".*\"/s.version     = \"#{version}\"/g\" #{pod_spec}")
+      unless updated_podspec_version 
+        restore_podspec
+        die("Failed to update version in podspec. Aborting.")
+      end
+
+      added_podspec = system("git add #{pod_spec}")
+      unless added_podspec
+        restore_podspec
+        die("Failed to add #{pod_spec} to index. Aborting.")
+      end
+
+      pushed_updated_version = system("git commit -m \"Bumping version to #{version}\"")
+      unless pushed_updated_version
+        restore_podspec
+        die("Failed to push updated version. Aborting.")
+      end
+    else
+      puts " > Podspec already set to #{version}. Skipping."
+    end
+
+    if force
+      puts "-> Tagging version"
+      tagged_version = system("git tag -s #{version_tag} -F #{release_notes}")
+      
+      die("Failed to tag version. Aborting.") unless tagged_version
+      puts "-> Pushing tag to origin"
+      pushed_tag = system("git push origin #{version_tag}")
+      die("Failed to push tag '#{version_tag}' to origin") unless pushed_tag
+    else
+       puts "-> Tagging version (force)"
+       tagged_version = system("git tag -f -s #{version_tag} -F #{release_notes}")
+       die("Failed to tag version. Aborting.") unless tagged_version
+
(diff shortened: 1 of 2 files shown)


release.rb
```diff
@@ -0,0 +1,215 @@
...
+  end
+
+  def self.run_force(version, release_notes_file_path)
+    run_internal(version, release_notes_file_path, true)
+  end
+
+  private
+
+  def self.run_internal(version, release_notes_file_path, force)
+    remote_branch = 'convert_release_script_to_rake_task'
```
jwfriese: We'll obviously want to change this to ```master``` once we're happy with the task.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+    pod_name = 'Quick'
+    pod_spec = 'Quick.podspec'
+    github_tags_url = 'https://github.com/jwfriese/Quick/tags'
```
jwfriese: Ditto for this -- we'll point to the proper repo's tags once we're finished.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+      puts ' - Configuring your system for PGP: https://git-scm.com/book/tr/v2/Git-Tools-Signing-Your-Work'
+      puts ' - Why: http://programmers.stackexchange.com/questions/212192/what-are-the-advantages-and-disadvantages-of-cryptographically-signing-commits-a'
+      puts
+      return
+    end
+
+    puts " > Found PGP key for git"
+
+    #verify_cocoapods_trunk_ownership(pod_name)
+    #puts " > Verified ownership to #{pod_name} pod"
```
jwfriese: As noted in the leading comment, these validation steps were always failing for me because the ```pod trunk me``` command did not pass Quick back as one of my pods.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+    puts
+    puts "Archiving carthage release..."
+    
+    archived_carthage_release = system("#{carthage} archive #{carthage_framework_name}")
+    die("Failed to archive framework for carthage. Aborting.") unless archived_carthage_release
+
+    puts
+    #puts "Pushing to pod trunk..."
+
+    #`#{pod} trunk push #{pod_spec}`
```
jwfriese: Same deal as the comments for lines 65-66.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+    puts " - Go to #{github_tags_url} and mark this as a release."
+    puts "   - Paste the contents of #{release_notes} into the release notes. Tweak for Github styling."
+    puts "   - Attach #{carthage_framework_name}.framework.zip to it."
+    puts " - Announce!"
+
+    `rm #{pod_spec}.backup`
+  end
+
+  def self.die(error_message)
+    raise("[ERROR]: #{error_message}")
```
jwfriese: I'm not certain that raising an error is the best way to kill the task early. I'm interested to hear if anyone agrees and has a more elegant killswitch.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+  end
+
+  def self.verify_availability(command) 
+    system("which #{command}")
+  end
+
+  def self.verify_version(version, force) 
+    die("This tag (#{version}) is an incorrect format. You should remove the 'v' prefix.") unless /^[vv]/.match(version)
+    die("This tag (#{version}) is an incorrect format. It should be in 'v{MAJOR}.{MINOR}.{PATCH}(-{PRERELEASE_NAME}.{PRERELEASE_VERSION})' form.") unless /^v(\d+\.)(\d+\.)(\d+)(-\w+(\.\d+)?)?$/.match(version)
```
jwfriese: I'm an intermediate at best at regex, so a little proof here from a regex master would be fantastic.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+      release_notes_file_path = '.release-changes'
+      latest_tag = `git for-each-ref refs/tags --sort=-refname --format="%(refname:short)" | grep -E "^v\\d+\\.\\d+\\.\\d+(-\\w+(\\.\\d)?)?\\$" | ruby -e 'puts STDIN.read.split("\n").sort { |a,b| Gem::Version.new(a.gsub(/^v/, "")) <=> Gem::Version.new(b.gsub(/^v/, "")) }.last'`
+      latest_tag = latest_tag.strip
+      puts " > Latest tag #{latest_tag}"
+      `echo "#{pod_name} v#{version}" > #{release_notes_file_path}`
+      `echo "================" >> #{release_notes_file_path}`
+      `echo >> #{release_notes_file_path}`
+      `echo "# Changelog from #{latest_tag}..HEAD" >> #{release_notes_file_path}`
+      `git log #{latest_tag}..HEAD | sed -e 's/^/# /' >> #{release_notes_file_path}`
+      system("#{ENV["EDITOR"]} #{release_notes_file_path}")
```
jwfriese: I used ```system``` here to spin up a subshell without trying to eat vim's stdout, which would cause vim to fail. Backticking the rest of the commands made sense to me in the moment.

release.rb
```diff
@@ -0,0 +1,215 @@
...
+  end
+  
+  def self.gather_release_notes(release_notes_file_path, version, pod_name) 
+    unless !release_notes_file_path.nil? && File.file?(release_notes_file_path)
+      puts " > Failed to find #{release_notes_file_path}. Prompting editor"
+      if ENV["EDITOR"].nil?
+        die("No default editor set. Assign an editor to the $EDITOR environment var. Aborting...")
+      end
+      release_notes_file_path = '.release-changes'
+      latest_tag = `git for-each-ref refs/tags --sort=-refname --format="%(refname:short)" | grep -E "^v\\d+\\.\\d+\\.\\d+(-\\w+(\\.\\d)?)?\\$" | ruby -e 'puts STDIN.read.split("\n").sort { |a,b| Gem::Version.new(a.gsub(/^v/, "")) <=> Gem::Version.new(b.gsub(/^v/, "")) }.last'`
```
jwfriese: This line is ludicrously long -- I'll address it in the next iteration. I forgot to do so in this one :sweat_smile:

release.rb
```diff
@@ -0,0 +1,215 @@
...
+
+    puts "-> Ensuring no differences to origin/#{remote_branch}"
+    fetched_origin = system("git fetch origin")
+    die("Failed to fetch origin. Aborting") unless fetched_origin
+
+    head_aligned_with_origin = system("git diff --quiet HEAD \"origin/#{remote_branch}\"")
+    die("HEAD is not aligned to origin/#{remote_branch}. Cannot update version safely. Aborting.") unless head_aligned_with_origin
+
+    puts "-> Building Carthage release"
+    built_carthage_framework = system("#{carthage} build --no-skip-current --platform iOS,Mac")
```
jwfriese: Notice the ```--platform iOS,Mac```. This is just a workaround that I sourced from the conversation around #422. Let me know if there is a better solution.

phatblat:
@jwfriese Would you mind incorporating the changes from #522 into your new ruby script? The general changes are:

1. Open GitHub in browser to new tag page
1. Remove Carthage (due to Swift binary incompatibility issues)

I'll look this script over more closely when I some some more time.

istx25:
Any updates on this? @jwfriese @phatblat

modocache:
This would be *great* to have. Cutting a Quick release is pretty easy right now, but it could definitely be easier.

istx25:
Is the PR ready for merging? @jwfriese @phatblat @modocache

jwfriese:
@istx25 - It's definitely not ready for merging in its current state. 

There are a number of things that I outlined in the initial post as items I needed some help working out. I don't know how many of those things might have been incidentally addressed since I initially made the PR. 

I'll be able to put some time into this over the weekend.

release.rb
modocache: Yeah, let's change this to `master` before merging this, and after you've tested that it all works.

release.rb
modocache: Gotcha. Please uncomment these before landing. If you'd like, I could add you to the Quick organization and as a CocoaPod owner, so `pod trunk me` no longer fails for you. 😉

release.rb
modocache: Yup, please uncomment these and the `puts` above.

release.rb
modocache: Error seems fine to me. What did you have in mind?

release.rb
modocache: Your regex seems absolutely correct to me! 💯 

My two cents: we could simplify this regex and our lives by just dumping the concept of a "prerelease version". I know Nimble has done prereleases in the past ([v2.0.0-rc.2](https://github.com/Quick/Nimble/releases/tag/v2.0.0-rc.2)), but personally I think Quick org projects aren't big/important enough to warrant this. If we push out a bad v1.0, we'll release a v2.0 -- not the end of the world.

Thoughts, @Quick/core?

release.rb
modocache: Sounds good to me!

release.rb
modocache: Cool, multiple lines here would definitely be nice!

release.rb
modocache: Sounds good.

I can't remember, though: didn't we at one point attach Carthage binaries to all of our releases? Looking at past Quick and Nimble releases now, though, I can't find any. GitHub's audit logs don't go back far enough for me to see all the activity that's ever been done around our releases, but I assume that at one point we deleted the binaries. Perhaps because Swift is not ABI stable? If anyone in @Quick/core knows for sure, feel free to chime in.

Anyway, since it appears we don't attach Carthage binaries, maybe we can delete this code for now?

release.rb
ashfurrow: Yeah, Quick is a larger project than Moya but we've never had a pre-release or RC or anything. We're also up to major version 8 though – doesn't bother me 😄

release.rb
briancroom: Yes, that's right - prebuilt frameworks containing Swift code really aren't viable yet, because they can only work with application code built using the exact same compiler toolchain, due to the lack of a stable Swift ABI. I remember going back and deleting frameworks attached to old releases to prevent Carthage from grabbing them which was causing headaches for some users.

modocache: Awesome, thanks for the clarification @briancroom. So let's either remove the Carthage code for now, or add it as a completely separate Rake task (outside of the release task).

istx25:
>I'll be able to put some time into this over the weekend.

Did that ever happen? 😸 

Would love an update.

istx25:
I feel that this PR has been abandoned. Any against closing this? 👍 from me for doing so.

jeffh:
Agreed, closing.

PR #478, opened 2016-02-09
Add Ja docs

takecian:
Hi, i also translated two docs.
Please review again.


modocache:
Thanks so much for all the translations, @takecian!!

PR #480, opened 2016-02-10
translate ja docs

takecian:
hi, i also translated docs.
please review.

Regards,
takecian.


modocache:
Thanks for being so careful around your translations of "edible"'! Very much appreciated!!

PR #486, opened 2016-02-16
Adjust the way that test observation suspension works.

briancroom:
When I updated the Nimble submodule to fix compilation on Xcode 7.3, I discovered that the changes made in https://github.com/Quick/Nimble/pull/244 break the Quick tests because of the way that we use `+[XCTestObservationCenter _suspendObservationForBlock:]` to enable running `XCTestSuite` instances during the middle of an existing run. The problem is that Nimble now relies on a test observer of its own in order to be able to report failures to XCTest (see the CI failure [here](https://travis-ci.org/Quick/Quick/jobs/109524574)). 

This PR changes the `QCKSpecRunner` behavior so that, instead of using private API on XCTestObservationCenter, it swizzles public methods to keep track of XCTest's own observers, then temporarily removes them from the center to implement the suspension behavior. I'm not convinced that this is necessarily any more future-proof than the old mechanism, because it relies on being able to add and remove observers at arbitrary times during an ongoing test run, but it does work with current XCTest versions, at least.

Any concerns, @Quick/contributors?

### DIFF
--- a/Sources/QuickTests/Helpers/XCTestObservationCenter+QCKSuspendObservation.m
+++ b/Sources/QuickTests/Helpers/XCTestObservationCenter+QCKSuspendObservation.m
@@ -1,0 +1,23 @@
+@import XCTest;
+#import <objc/runtime.h>
+
+@interface XCTestObservationCenter (Redeclaration)
+- (NSMutableSet *)observers;
+@end
+
+@implementation XCTestObservationCenter (QCKSuspendObservation)
+
+static BOOL (^isFromApple)(id, BOOL *) = ^BOOL(id observer, BOOL *stop){
+    return [[NSBundle bundleForClass:[observer class]].bundleIdentifier containsString:@"com.apple.dt.XCTest"];
+};
+
+- (void)qck_suspendObservationForBlock:(void (^)(void))block {
+    NSSet *observersToSuspend = [[self observers] objectsPassingTest:isFromApple];
+    [[self observers] minusSet:observersToSuspend];
+
+    block();
+
+    [[self observers] unionSet:observersToSuspend];
+}
+
+@end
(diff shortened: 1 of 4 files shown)


briancroom:
Well, I take back that part about avoiding private APIs...

Unfortunately it seems that `xctool` does swizzling on `XCTestLog` which is a "legacy test observer" and uses a different registration mechanism, with the result that it threw exceptions about getting into an invalid state when the nested test suite started running. I've now pushed an alternative implementation which manipulates the `XCTestObservationCenter`'s `observers` property to remove any observers (legacy or standard) which come from XCTest, while leaving custom ones (like Nimble's).

modocache:
Sorry to keep you waiting, @briancroom! This looks great. I'm also not thrilled about all the hoops we jump through to support `xctool`... but c'est la vie, I guess :sweat_smile:

Sources/QuickTests/Helpers/XCTestObservationCenter+QCKSuspendObservation.m
```diff
@@ -0,0 +1,23 @@
...
+#import <objc/runtime.h>
+
+@interface XCTestObservationCenter (Redeclaration)
+- (NSMutableSet *)observers;
+@end
+
+@implementation XCTestObservationCenter (QCKSuspendObservation)
+
+static BOOL (^isFromApple)(id, BOOL *) = ^BOOL(id observer, BOOL *stop){
+    return [[NSBundle bundleForClass:[observer class]].bundleIdentifier containsString:@"com.apple.dt.XCTest"];
```
modocache: nit-pick: As a maintainer I would love an inline comment here describing why this `isApple` predicate is necessary. I understand I can `git blame` and find this pull request, but \*hand waves\* code locality something something.

Sources/QuickTests/Helpers/XCTestObservationCenter+QCKSuspendObservation.m
```diff
@@ -0,0 +1,23 @@
...
+};
+
+- (void)qck_suspendObservationForBlock:(void (^)(void))block {
+    NSSet *observersToSuspend = [[self observers] objectsPassingTest:isFromApple];
+    [[self observers] minusSet:observersToSuspend];
+
+    block();
+
+    [[self observers] unionSet:observersToSuspend];
```
sharplet: Just a thought: it might be technically possible that if `block` raises an exception this line won't run, leaving the `observers` set in an unexpected state. Probably a real edge case, even if it's possible.

Just wondering whether it would be worth wrapping this in a try/finally, and restoring the observers before rethrowing the exception?

Sorry for the drive-by comment!

modocache: No apologies necessary! Thanks for the input, @sharplet!

I'm in favor of a try/finally here, but how is it possible for the block to raise? Does `+[NSBundle bundleForClass:]` raise if passed a `nil` value?

sharplet: Not the `isFromApple` block, the `block` parameter—which is running the nested test suite. I figure any test suite that's being run here could possibly raise?

sharplet: Although if this is just a test helper it's probably not a big deal!

modocache: Ah yes, my bad! Yes, this is just a test helper, but I suppose one could never be too careful! A try/finally might be nice.

briancroom:
Thanks for the review and good points @modocache and @sharplet! I've pushed a commit addressing these things to master.

Issue #485, opened 2016-02-16
Documentation Mock


takecian:
Add document which describes Mock and how to use Mock in Swift.

modocache:
Awesome! :+1:

istx25:
I guess this issue can be closed now since #487 has been merged?

modocache:
Yup, thanks for the reminder!

deveras:
Hi guys,

I was reading https://github.com/Quick/Quick/blob/master/Documentation/en-us/TestUsingTestDoubles.md and wanted to note that the following:

Mock object: Used for receiving output from a test class.
Stub object: Used for providing input to a test class.
Fake object: Behaves similarly to the original class, but in a simplified way.

They do not follow the definitions of ISO/IEC/IEEE 24765:2017 (Systems and software engineering --Vocabulary) where:

Mock object is a temporary dummy object created to aid testing until the real objects become available
Stub is a computer program statement substituting for the body of a software module that is or will be defined elsewhere.

For consistency, it might be good idea to update the documentation and follow standards to avoid confusion and ease understanding unless the intention is to deviate from what common, and if so, what are the advantages introduced? Do we want to be a class apart?

PR #487, opened 2016-02-17
Add a document about mock

takecian:
I wrote a document about mock(#485), as of now, Japanese edition only.
Please review.

Thanks!


takecian:
I also added document in English.

takecian:
Hi @jeffh , thank you for review.
I fixed typo and added language specifier.

modocache:
Thanks @takecian! Planning on reading/merging this over the weekend. :boat:

Documentation/ja/TestUsingMock.md
```diff
@@ -0,0 +1,153 @@
+# モックを使ったテスト
+
+## モックとは
+
+モックとはテスト対象のオブジェクト(クラス、構造体)が呼び出し先のオブジェクトと意図したとおりに協調動作するかどうかをテストするために使うオブジェクトのことです。
```
modocache: I prefer the documentation to focus foremost on the benefits of testing applications, and secondly on techniques on how to do so. This explanation of a mock is very technical. It focuses on what a mock is, before explaining what problems a mock would solve. Instead, I would try something like this:

1. Explain a problem that comes up often when people write tests. Here's one example: `BTests` tests `B`, which uses `A`. Now bugs in `A` cause `BTests` to fail, even though `B` is fine! All the failures are confusing--it's hard to answer the question "what's broken?"
2. Explain that we can avoid this problem by using a stand-in for `A` in `BTests`.
3. Show an example of using using a stand-in.
4. Mention that these sorts of "stand-in" objects are referred to as "test doubles", "mocks", and other things. We only introduce these technical details once the user understands the problem.

I think the structure of @jeffh's [Testing in Swift](https://realm.io/news/testing-in-swift/) talk ([Japanese here](https://realm.io/jp/news/testing-in-swift/)) is really good. We should also provide a link to that talk!

Documentation/ja/TestUsingMock.md
```diff
@@ -0,0 +1,153 @@
...
+
+```swift
+// Swift
+protocol ArticleProviderProtocol: UITableViewDataSource {
+    var articles: [Article] { get }
+    weak var tableView: UITableView! { get set }
+    func setup()
+    func fetch()
+}
+```
```
modocache: Would it be possible to demonstrate a problem that can be solved with mocks *without* using UIKit? I'm concerned about readers who may not be 100% familiar with UIKit and `UITableViewDataSource`. Just wondering whether that would make this easier to read--thoughts? :raised_hands:

modocache:
(Still reading--I'll post more thoughts soon!)

Documentation/ja/TestUsingMock.md
takecian: Sure, i will update sample code to more simple, it would be easier:ok_hand:

Documentation/ja/TestUsingMock.md
takecian: Thanks for feedback! I will update document.

takecian:
I updated document about probem when writing test and test doubles.
I made sample code more easier.
Any feedbacks are welcome!

jeffh:
@modocache: Did you want to review this more? Otherwise I'll merge this in a couple days.

modocache:
Oops, sorry! Yes I'll read this in the next few days. Feel free to merge in the meantime, though, and I'll just review post-commit.

istx25:
Are there any plans to merge this?

takecian:
@modocache hi, any feedbacks about this? :)

Documentation/en-us/README.md
```diff
@@ -27,6 +27,8 @@ testing, consider reading them in the order they're introduced below:
   *examples* and *example groups*.
 - **[Testing OS X and iOS Applications](TestingApps.md)**:
   Read this to learn more about testing code that uses the AppKit and UIKit frameworks.
+- **[Testing with test doubles](TestUsingTestDoubles.md)**:
+  Read this to learn what test doubles is and how to use it.
```
modocache: Grammar: "are and how to use them".

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
+# Testing with Mock
+
+## Test doubles
+
+Following problem comes up often when writing tests. Here, `B` Class uses(depends on) `A Class`.
```
modocache: Grammar: "The following problem"

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+
+![](https://github.com/Quick/Assets/blob/master/Screenshots/TestUsingMock_BusesA.png)
+
+Here's one example: BTests tests B, which uses A. Now bugs in A cause BTests to fail, even though B is fine! 
+All the failures are confusing--it's hard to answer the question "what's broken?"
+
+To avoid this problem, we can use a stand-in for A in Btests. In this case, we call stand-in for A `A' class`.
```
modocache: I find `A` and `A'` hard to distinguish visually. Can we name the "stand-in for `A`" something else? Maybe `StandInA`?

In general, I think I prefer using more fun examples. 😅  Rather than `class A` and `class A'`, these could be `class Airplane` and `class PaperAirplane`, or something. 😛

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+
+![](https://github.com/Quick/Assets/blob/master/Screenshots/TestUsingMock_BusesAmock.png)
+
+`A' class` has same methods--however implementation might be different, and properties as `A Class`. So we can replace `A Class` with `A' Class` in BTests.
```
modocache: This sentence is a little confusing I think. Perhaps better as "`A' class` has the same methods and properties as `A class` (although its implementations of those methods and properties might be different). So we can..."

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+
+Objects like `A' Class` are called 'test doubles'. 'test doubles' are used as stand-in for tests.
+There are some kinds of 'test doubles'.
```
modocache: ~~some~~ several

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+
+- Mock object: used for receiving output from test class
+- Stub object: used for providing input to test class
+- Fake object: behaves similar as original class
+
+Here, we explain using `Mock` next section.
+
+## Mock
+
+Mock is object(class, struct) used for verifying test object works with other objects from point of view of test target's output.
```
modocache: "A mock is a class or struct used for"

takecian:
Thanks for feedback! I'll fix it!

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Writing test with Mock in Swift
+
+#### Sample app
+
+For example, we show an app which retrieves data from internet.
```
modocache: I would put this example, or any example, first in this document. That's because, as a reader, I'm very impatient. I don't think I would pay attention to explanations of what "test doubles" or "mocks" were unless I knew that I wanted to use them. So I think it'd be best to motivate the reader first, before explaining any technical jargon like "test double" or "mock".

Put another way: I think this document should answer the question "why should you, the reader, care about what test doubles are?" in as few sentences as possible at the very beginning. Then it should proceed to explain what test doubles are and how to use them.

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+
+Here's DataProvider which implements DataProviderProtocol.
+
+```swift
+// Swift
+class DataProvider: NSObject, DataProviderProtocol {
+    func fetch(callback: (data: String) -> Void) {
+        let url  = NSURL(string: "http://example.com/")!
+        let session = NSURLSession(configuration: NSURLSessionConfiguration.defaultSessionConfiguration())
+        let task    = session.dataTaskWithURL(url, completionHandler: {
```
modocache: nit-pick: No need to align `=` here. I prefer a single space before and after the `=`.

Documentation/en-us/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,131 @@
...
+            self.resultLabel.text = data
+        })
+    }
+}
+```
+
+#### Writing test with Mock of DataProviderProtocol
+
+ArticleViewController depends on DataProviderProtocol.
+Here, create Mock which inherits DataProviderProtocol in Test Targets in order to test ViewController.
```
modocache: Not to keep mentioning the same point, but I think this line is a good example of what I mean. If I were a reader who didn't know what mocks were, I would ask "why should I create a `MockDataProvider`? Why are you asking me to write this code?"

The actual reasons to write the mock are compelling:

1. Your tests will run much faster.
2. They will run correctly whether you're connected to the Internet or not.
3. They will not fail because of errors in your actual `DataProvider` (i.e.: test isolation).

Documentation/ja/TestUsingTestDoubles.md
```diff
@@ -0,0 +1,135 @@
...
+            (data, resp, err) in
+            let string = NSString(data:data!, encoding:NSUTF8StringEncoding) as! String
+            callback(data: string)
+        })
+        task.resume()
+    }
+}
+```
+
+ViewController の viewDidLoad 中にデータの取得(fetch()の呼び出し)を行います。
```
modocache: \`viewDidLoad\` and \`fetch()\`

modocache:
@takecian My pleasure! Sorry for letting this sit for so long.

To make things quicker, I'll merge this now. Feel free to make commits to modify these documents directly -- or use pull requests to send the changes, and then merge them yourself.

takecian:
Sure!

istx25:
Hurray 🎉

PR #488, opened 2016-02-18
Fix the test counting used to determine when to execute `afterSuite` hooks

### LINKED ISSUE #482: Using xit or fit causes your custom Configuration beforeSuite/afterSuite not to run
If you create a custom Configuration class and add your own custom closures for:
* beforeSuite,
* afterSuite,
* beforeEach
* afterEach

and then use fit (focused it) or xit (exclude it) on any of your Quick tests, the configuration closures for beforeEach and afterEach are called, however the beforeSuite and afterSuite closures are not.

If there are no focused/excluded tests in the suite, then the beforeSuite and afterSuite closures are invoked.

briancroom:
Fixes #482.

I had really been hoping to introduce test coverage around this (cf. [here](https://github.com/Quick/Quick/blob/master/Sources/QuickTests/FunctionalTests/AfterSuiteTests.swift#L41)), but came to the conclusion that it would require more extra infrastructure in the test suite (because the assertion would have to run *after* Quick's test suite finishes, clearly) than I have time to build right now.

### DIFF
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -84,7 +84,7 @@ final public class Example: NSObject {
 
         numberOfExamplesRun += 1
 
-        if !world.isRunningAdditionalSuites && numberOfExamplesRun >= world.exampleCount {
+        if !world.isRunningAdditionalSuites && numberOfExamplesRun >= world.includedExampleCount {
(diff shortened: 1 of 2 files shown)


jeffh:
LGTM. Thanks @briancroom!

PR #493, opened 2016-02-29
Fix typo

pixyzehn:
This is a trivial thing. Fixed typo in Japanese. 
Thank you for good Japanese documents:)


modocache:
Thanks, @pixyzehn! :100:

PR #500, opened 2016-03-11
Updated Nimble to 3.1.0 in Podfile example

tkuichooseyou:
Also is there a reason why the example points to a specific Nimble version instead of ~>


jeffh:
Thanks @tkuichooseyou!

Issue #494, opened 2016-03-01
Xcode 7.3 support


raven:
I see [this revert](https://github.com/Quick/Quick/commit/bc67084995ba3c27a52cf19036f9585f09cc48f1) https://github.com/Quick/Quick/commit/bc67084995ba3c27a52cf19036f9585f09cc48f1 for updating to the latest Nimble, however I can't find any further reasons as to why it failed and was reverted. I also can't find an associated issue on the Nimble project. What can I do to help?

briancroom:
Thanks for raising this again @raven. I opened https://github.com/Quick/Quick/pull/486 a couple of weeks ago with the necessary changes to allow using the newer Nimble versions and still have a functional test suite for Quick. It's been pretty quiet around that PR so I'd welcome any feedback! I realize that the changes introduced there aren't the most pleasant...

briancroom:
@ravan Can you check if things seem to be working for you now that #486 has been merged?

raven:
I have since updated both Quick and Nimble to latest commits on master for both, and our project is reporting number of tests 0, failures 0 when run via fastlane's scan

```
+--------------------+-------+
|        Test Results        |
+--------------------+-------+
| Number of tests    | 0     |
| Number of failures | 0     |
+--------------------+-------+
```

Also running the unit tests straight from within Xcode does appear to be running the tests(seen from the Test navigator), but the console output doesn't report any success or failure, or any mention of individual tests passing / failing.

Now our project has a bunch of moving parts, and updating Quick/Nimble was the first thing to bite off after updating to Xcode7.3 so I'll try and spend some time to discern what is failing where and report back.

gapl:
@raven Have you figured out how to solve this issue? We are also experiencing this issue now.

paweldudek:
Same issue here. @raven the problem with `scan` is that it uses `xcpretty` under the hood. `xcpretty` parses XCTest output to determine what test succeeded/failed. Since there's no output from XCTest right now the tool itself is broken. 

The fix would be to restore output or find a way to let `xcpretty` know about test successes/failures, however I'm not skilled enough with XCTest to know/propose a solution.

raven:
Havent looked at this recently. But it was observable outside of using scan / xcpretty. Just a straight up cmd+u from within Xcode only put some basic logs into the console and then 0 tests. Im away from the project until tomorrow, Il provide more info then. 

Get Outlook




[...]

PR #505, opened 2016-03-24
Nimble++

phatblat:
to bring in Quick/Nimble#269

PR #507, opened 2016-03-25
Updates Nimble submdule commit.

ashfurrow:
[As discussed](https://github.com/Quick/Nimble/pull/271#issuecomment-201082777), we need to update this to get the latest fix for #494.


modocache:
Thanks!! Hopefully peeps aren't using the version of Nimble Quick's tests depend on as their Nimble dependency, but I have a feeling some do.

Feel free to merge once CI passes.

ashfurrow:
Hmm, both builds stalled out. Should I just try to restart?

ashfurrow:
Ha, looks like this happened the last time we upped the Nimble submodule commit: https://github.com/Quick/Quick/pull/505 I'll try restarting.

ashfurrow:
The [same test](https://github.com/Quick/Quick/blob/3a1d88fb81476001c62f086dbd94bb0f5354f1a3/Sources/QuickTests/FunctionalTests/FailureTests%2BObjC.m#L56-L59) is failing on both CI providers, probably not a coincidence... Since the Nimble change we're trying to bring in has to do with _when_ test observation starts, I'm willing to bet it's a valid failure. I'll look into it and report back.

ashfurrow:
Tests didn't fail when running within Xcode. [Invoking](https://github.com/ashfurrow/Quick/blob/3a1d88fb81476001c62f086dbd94bb0f5354f1a3/Rakefile#L32) `xcodebuild -workspace Quick.xcworkspace -scheme Quick-OSX clean test` led to :green_apple: tests. Looking into possible `xctool` issues.

ashfurrow:
Hmm, might be a trickier issue than expected. Kif ran into [similar problems](https://github.com/kif-framework/KIF/pull/819) that https://github.com/Quick/Nimble/pull/271 was supposed to fix, but decided not to use the solution we did `dispatch_async`. However, their solution appears to [remove any call to `addTestObserver:`](https://github.com/kif-framework/KIF/pull/819/files#diff-d869217c7fbf012a7678c6ab49fcdbc1L46). I don't want to ask for clarification there without a better understanding of what `XCTestObservationCenter` is and why we need it.

Suggestions on how to proceed would be awesome :bow: I feel a little out of my depth.

phatblat:
Looks like KIF's solution was to move their test observation functionality into the `KIFTestCase.setUp` method to start before each test. Nimble doesn't have the an `XCTestCase` subclass to move its logic to.

The Nimble test observation logic appears to only be used when [`_runtime(_ObjC)`](https://github.com/Quick/Nimble/blob/master/Sources/Nimble/Adapters/NimbleXCTestHandler.swift#L40-L51) so that it can [testCase.recordFailureWithDescription](https://github.com/Quick/Nimble/blob/master/Sources/Nimble/Adapters/NimbleXCTestHandler.swift#L68) instead of `XCTFail` on the swift side.

ashfurrow:
Hmm. You reckon its worth opening an issue on Nimble to discuss options? Maybe there's another hook we could use, and `dispatch_once` to add the observer.

phatblat:
Yes, let's open a Nimble issue to discuss. I've been doing some debugging and have a theory.

phatblat:
@ashfurrow can you update your branch to Quick/Nimble@1a9e53e2d74363e1d6320a87a5028bc234c9a4d4?

ashfurrow:
Done :+1: Let's see how CI goes :smile:

phatblat:
Aw yeah! :rocket:

briancroom:
Awesome! Thanks guys 🙌

phatblat:
Time for a release?

modocache:
Sure! @phatblat, I've added you as a `pod trunk` owner, so you should be able to tag a release. Use `script/release` to do so 👍

briancroom:
There's also the nice section in the Contributing document about making releases: https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md#creating-a-release

PR #512, opened 2016-04-06
Fix example description in QuickSpec.h

muukii:
QuickSpec.spec() is instance method.

### DIFF
--- a/Sources/Quick/QuickSpec.h
+++ b/Sources/Quick/QuickSpec.h
@@ -33,7 +33,7 @@
  Override this method in your spec to define a set of example groups
  and examples.
 
-     override class func spec() {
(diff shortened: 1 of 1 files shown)


phatblat:
True, the `spec` method is not a class method. However, that example is Swift in an Objective-C header. Would you mind changing the example to Objective-C?

muukii:
@phatblat 
Sure.
But, original example was written by Swift.
Still do you change it to Objective-C?

phatblat:
Never mind. This example shows up in the docs for both objc and swift.

screen_shot_2016-04-06_at_8_15_49_pm

muukii:
Thank you!

PR #513, opened 2016-04-07
Format example spec in doc comment

phatblat:
Wrapped example spec with `@code` and `@endcode` to preserve formatting in help callout.

# Before

screen_shot_2016-04-06_at_8_15_49_pm

# After

screen shot 2016-04-06 at 8 41 18 pm

### DIFF
--- a/Sources/Quick/QuickSpec.h
+++ b/Sources/Quick/QuickSpec.h
@@ -33,13 +33,15 @@
  Override this method in your spec to define a set of example groups
  and examples.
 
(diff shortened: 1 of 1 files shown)


briancroom:
Nice catch! I'm surprised to see that those annotations are required - I guess because it is an Obj-C source file rather than Swift?

phatblat:
Yeah, the objc doc comments don't support markdown and need some help knowing where the code block is.

PR #516, opened 2016-04-13
Adds instructions for putting the QUICK_DISABLE_SHORT_SYNTAX macro in project config

twobitlabs:
We have a long-lived project with both Objective-C and Swift tests. The Ojective-C specs use Specta, which conflicts with Quick. For some reason no matter what header I add the `QUICK_DISABLE_SHORT_SYNTAX` macro to, it does not apply. The only way I was able to get it to work was by creating the macro in the test target's build configuration.

This PR just adds instructions for how to do that.


modocache:
Awesome, thanks! 👍

Issue #515, opened 2016-04-11
Quick was compiled with optimization - stepping may behave oddly; variables may not be available


onmyway133:
I got a breakpoint EXE_BAD_INSTRUCTION at

```swift
internal func it(description: String, flags: FilterFlags, file: String, line: UInt, closure: () -> ()) {
        if beforesCurrentlyExecuting {
            raiseError("'it' cannot be used inside 'beforeEach', 'it' may only be used inside 'context' or 'describe'. ")
        }
        if aftersCurrentlyExecuting {
            raiseError("'it' cannot be used inside 'afterEach', 'it' may only be used inside 'context' or 'describe'. ")
        }
        guard currentExampleMetadata == nil else {
            raiseError("'it' cannot be used inside 'it', 'it' may only be used inside 'context' or 'describe'. ")
        }
        let callsite = Callsite(file: file, line: line)
        let example = Example(description: description, callsite: callsite, flags: flags, closure: closure)
        currentExampleGroup.appendExample(example)  // GOT BREAKPOINT HERE
    }
```

This happens in my test targets, sometimes I get it, so don't know how to reproduce. Any help is appreciated 😢

onmyway133:
I'm using code like this to read fixture files

```swift
class DummySpec: QuickSpec {}

struct Helper {
  static func read(fileName: String) -> NSData {
    let path = NSBundle(forClass: DummySpec().dynamicType).pathForResource(fileName, ofType: "json")!
    return NSData(contentsOfFile: path)!
  }
  
  static func readJSON(fileName: String) -> [String: AnyObject] {
    let data = Helper.read(fileName)
    return try! NSJSONSerialization.JSONObjectWithData(data, options: .AllowFragments) as! [String: AnyObject]
  }
}
```

after changing it to

```swift
static func readJSON(fileName: String) -> [String: AnyObject] {
    let data = Helper.read(fileName)
    if let json = try? NSJSONSerialization.JSONObjectWithData(data, options: .AllowFragments),
      jsonDictionary = json as? [String: AnyObject] {
      return jsonDictionary
    } else {
      return [:]
    }
  }
```

I don't get that problem, it was probably the optional force unwrapping issue, but I check that the data is ALWAYS read successfully 🐼

morganchen12:
I don't think this is a bug on your end; this looks like it may be related to or a duplicate of #506.

onmyway133:
@morganchen12 you're right. I still get this problem, sometimes

[...]

PR #517, opened 2016-04-21
Fix crash caused by currentExampleGroup being unexpectedly nil

### LINKED ISSUE #506: Intermittent fatal error due to currentExampleGroup being nil
I'm intermittently getting a `fatal error: unexpectedly found nil while unwrapping an Optional value` in [World+DSL.swift:58](https://github.com/Quick/Quick/blob/master/Sources/Quick/DSL/World+DSL.swift#L58) whilst running my unit tests. The value which is nil is `currentExampleGroup`.

It's only one one spec, so I'm wondering if I'm doing something wrong there. The fact it's intermittent makes it feel a bit odd though.

I was wondering if anyone has seen this before?

briancroom:
This could occur because `+[QuickSpec initialize]` was being called recursively when a new `QuickSpec` subclass was being referenced for the first time while another spec was in the middle of gathering examples.

Fixes #506, #515


briancroom:
It looks like we are getting bitten by an `xcodebuild` hang on CI here, as documented at https://github.com/facebook/xctool/issues/689. I'm going to merge this since the Linux and non-xctool tests are consistently passing and I don't want to leave this crasher unpatched any longer than necessary.

PR #518, opened 2016-04-22
Bumps Nimble version to 4.0.1.

ashfurrow:
See the [release notes](https://github.com/Quick/Nimble/releases/tag/v4.0.1). Note that one of these changes (the race condition) was already included in the commit Quick was using.


ashfurrow:
The CI failures are strange, I've restarted them. Weird to get the same timeout on both PR and push builds. 

I should have included the changes involved in the Nimble commit. You can browse them at https://github.com/Quick/Nimble/compare/1a9e53e2d74363e1d6320a87a5028bc234c9a4d4...0cf12916a276ba7bc7a66ded95bd8f7e77ca7f6b

ashfurrow:
Weird, same timeout issue:

```
  [Info] Collecting info for testables...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated

```

Looks like it's happening on [other builds](https://travis-ci.org/Quick/Quick/jobs/122862108) too. Any suggestions?

briancroom:
Indeed.. I observed that here as well: https://github.com/Quick/Quick/pull/517#issuecomment-213200717

It looks like we are getting bit by an Apple bug in `xcodebuild` which causes it to hang intermittently on certain types of invocations that `xctool` (also Carthage) uses. I'll open a new issue to discuss the best course of action here.

PR #522, opened 2016-04-22
Release process update

ashfurrow:
Similar work to https://github.com/Quick/Nimble/pull/284 . ~~Waiting to hear back about attaching binary releases before removing that from the script, too.~~


phatblat:
Very nice. Is this ready to merge? CI failures don't apply to this change.

phatblat:
Just noticed #475 and #467. These changes should be incorporated into the if we continue with converting the release script.

briancroom:
Let's do it! 👍

briancroom:
Ah yes, I forgot about that work. It looks like we've let that get pretty stale, unfortunately. I think it's reasonable to merge this in now since it's ready, and include doc updates in #475 when that one is ready to go.

phatblat:
Yep, agree completely.

Issue #520, opened 2016-04-22
CI is flaky due to buggy xcodebuild


briancroom:
Our OS X build on Travis is incredibly flaky right now. See e.g. [these](https://travis-ci.org/Quick/Quick/builds/122863200) [builds](https://travis-ci.org/Quick/Quick/builds/124912462). The underlying issue appears to be a bug in Xcode 7.3's `xcodebuild` which results in `xctool` hanging just before it starts to execute tests. There is some discussion of the issue [here](https://github.com/facebook/xctool/issues/689) and the consensus currently seems to be that there's not much to be done besides wait for a fix.

I'm thinking that we should either
1. Disable the `xctool` invocations for the Travis build altogether for now, or
2. Split the `xctool` piece out to a separate job, so it's easy to verify if the `xcodebuild` test run succeeded, even if the overall build fails.

Thoughts @Quick/contributors?

modocache:
I'm in favor of option 2. Some people use xctool out there, and I think we should try to support that use case. I don't think we'll succeed at that unless we integrate it into our CI somehow.

(Warning: Backseat programmer alert!) For https://github.com/facebook/xctool/issues/689, I wonder if it would be possible to avoid using `xcodebuild -showBuildSettings`. The argument is that that's the point of failure, and it's up to Apple to fix it? I wonder if `xctool` could avoid invoking that in the meantime...?

briancroom:
It may indeed be possible, but I am definitely *not* the one to start messing with the `xctool` source. :shudder:

To be clear, I wasn't suggesting permanently disabling the `xctool` testing with option 1.. but it's hard to say how long it will be before there's a stable Xcode release that fixes the issue.

modocache:
Oh yeah, I understand. But option 2 seems like it would help us at least be aware of how well `xctool` works with Quick (or in this case, even on its own).

ashfurrow:
I think option 2 works, open to suggestions on an implementation. We could add a new entry to our CI matrix to text with xcodebuild vs xctool, or maybe some more cleverer?

briancroom:
Taken care of by #525.

PR #523, opened 2016-04-23
Split out a separate Travis job for testing xctool

briancroom:
This is my attempt at implementing option 2 as discussed at #520. [Here](https://travis-ci.org/briancroom/Quick/builds/125160996) is the build Travis made against this commit already, where it can be seen how the `xctool` job failed while the Linux and basic OS X jobs passed, as expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,26 @@
-os:
-  - osx
-  - linux
-language: generic
 osx_image: xcode7.3
-sudo: required
-dist: trusty
+language: generic
+matrix:
+  include:
+    - os: osx
+      env:
+        - PLATFORM=osx
+    - os: osx
+      env:
+        - PLATFORM=ios
+    - os: osx
+      env: 
+        - XCTOOL=1
+        - PLATFORM=osx
+    - os: osx
+      env: 
+        - XCTOOL=1
+        - PLATFORM=ios
+    - os: linux
+      env:
+        - PLATFORM=linux
+      sudo: required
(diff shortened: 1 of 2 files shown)


phatblat:
Could you add an environment variable such as `PLATFORM=iOS` to these? That will make the travis menu more informative so we know which build is which.

screen shot 2016-04-24 at 10 22 14 am

briancroom:
Currently each job builds and tests on both OS X and iOS in sequence. Do you think we should split those out further?

phatblat:
My mistake, the two "no environment set" are xcodebuild on OS X (running both iOS and OS X) and linux builds. I would like to see the target platforms separate and it can help make the build faster, but my main point is to have some sort of label on each entry we see on travis so we know at a glance more about what failed.

briancroom:
I've updated this now to split out the iOS and OSX jobs. How does this look to you @phatblat?

screen shot 2016-04-27 at 1 41 37 pm

phatblat:
😍 What about adding `PLATFORM=linux` on the last one?

briancroom:
Sure! Seems useful as a label.

I've pushed and squashed now.

phatblat:
screen shot 2016-04-27 at 12 14 35 pm

This is a thing of beauty. We can tell at a glance what's failing and what's not.

phatblat:
Merged as the xctool CI issues are known

modocache:
It is known.

PR #524, opened 2016-04-24
Better handling of unsupported tvOS devices

phatblat:
Quick cannot currently be used on tvOS devices since bitcode is required but the XCTest framework does not contain bitcode. By changing the `SDKROOT` to `appletvsimulator` on the Quick-tvOS target, this has no effect on simulator builds, but provides an immediate and much more informative error on device builds instead of linker errors.

## Before

screen shot 2016-04-24 at 10 11 07 am

## After

screen shot 2016-04-23 at 5 48 43 pm

# Carthage

Carthage currently cannot build Quick since it attempts to build for both the simulator and device (#422). I've proposed a fix for this in Carthage/Carthage#1268 which keys off of the `ENABLE_BITCODE` setting being `NO` to skip building for the device since bitcode is required to build for tvOS devices. This setting has no effect on simulator builds where bitcode is not used.

### DIFF
--- a/Quick.xcodeproj/project.pbxproj
+++ b/Quick.xcodeproj/project.pbxproj
@@ -1584,6 +1584,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(PLATFORM_DIR)/Developer/Library/Frameworks",
 					"$(inherited)",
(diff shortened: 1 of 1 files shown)


phatblat:
I starting to read "CI" as "sigh" because of annoying failures like [this](https://travis-ci.org/Quick/Quick/jobs/125393606)

briancroom:
Yeah..this is the `xctool` issue again. Regardless of that, this looks good to me. Especially now that https://github.com/Carthage/Carthage/issues/1280 has made it in 😄 Thanks @phatblat!

phatblat:
Sweet! Now I can use Quick to verify that tvOS builds of test frameworks continue working in Carthage.

briancroom:
👍 really happy to see tvOS getting some love.

phatblat:
I'll comment on and close the related Quick & Nimble issues once there's a Carthage release out with the fix.

phatblat:
My ultimate plan is to add watchOS support to these framework as soon as we can actually unit test on that platform. (June? 🙏🏻)

briancroom:
Yeah, I'm really curious to see if watchOS test bundles were a priority for the team this year! ⌚

phatblat:
I'm betting on it as it's already halfway there for the watchOS simulator. I was able to [hack together a watchOS framework test bundle](https://github.com/phatblat/WatchApp). Linking failed when trying a watch app target.

briancroom:
Oh wow, that's some neat hacking! Cool to see that mostly working. That linker failure makes sense, since native watchOS executables use a different architecture than iOS ones do.

PR #529, opened 2016-05-12
Adds podspec lint on CI

### LINKED ISSUE #521: Add `pod lib lint` to CI
As discussed in Quick/Nimble#284, the Quick pod should be linted regularly to ensure that releases can be pushed to CocoaPods Trunk. This step should be part of the regular CI process to catch any issues early.

pedrovereza:
* Created a new build process just for lint (similar approach used in https://github.com/Quick/Nimble/pull/288)
* ~~Travis claims that [cocoapods is already installed on OS X instances](https://docs.travis-ci.com/user/osx-ci-environment#Gems-in-the-global-gem-set), so I didn't bother using `gem` or `bundle` to install it.~~
* Added a task `podspec:lint` on Rakefile
* Changed `travis-script-osx` to run either `podspec:lint` or `tests` (not sure if that's the best approach tho)

This fixes #521

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ matrix:
       env:
         - PLATFORM=osx
     - os: osx
+      sudo: required
+      env:
+        - PODSPEC=1
+    - os: osx
       env:
         - PLATFORM=ios
     - os: osx
@@ -24,6 +28,7 @@ matrix:
 install:
   - if [[ "$TRAVIS_OS_NAME" == "osx"   ]]; then ./script/travis-install-osx;   fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./script/travis-install-linux; fi
+  - if [[ "$PODSPEC" ]]; then sudo gem install bundler; bundle install; fi
 script:
   - if [[ "$TRAVIS_OS_NAME" == "osx"   ]]; then ./script/travis-script-osx;   fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./script/travis-script-linux; fi
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,3 @@
+source "https://rubygems.org"
+
+gem 'cocoapods', '1.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,67 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    activesupport (4.2.6)
+      i18n (~> 0.7)
+      json (~> 1.7, >= 1.7.7)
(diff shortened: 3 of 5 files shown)


pedrovereza:
Seems like it's https://github.com/Quick/Nimble/pull/286 all over again 😞 
````
/Users/travis/.rvm/gems/ruby-2.0.0-p648/gems/psych-2.0.17/lib/psych.bundle: [BUG] Segmentation fault
````

I've read https://github.com/Quick/Nimble/pull/286 several times and still not sure what magic was done by @ashfurrow to make it work. I see that some cases passed for him on the push check but not on the PR one, not sure if that's what I'm facing here.
image

phatblat:
I'm not sure if it crashed while installing bundler or when it tried to run bundle install. According to the [travis docs](https://docs.travis-ci.com/user/languages/ruby#Travis-CI-uses-Bundler) both are redundant, but the Gemfile may need to be specified since this project is using a matrix.

phatblat:
@ashfurrow got it working in Quick/Nimble#288

pedrovereza:
@phatblat Yes, I tried to follow most of what he did on https://github.com/Quick/Nimble/pull/288. Still, his PR failed in the PR check with a seg fault: https://travis-ci.org/Quick/Nimble/builds/125416685 but passed the push check.

ashfurrow:
Ahh yeah, this problem. So I never really figured out what was going on, I assumed it was a Travis issue involving changes to `.travis.yml`'s build matrix on a pr build. After I got all the builds passing for the pr build _except_ the podspec lint, we merged and it worked. 

screen shot 2016-05-13 at 9 06 06 am

It's not a satisfying answer, but I'm willing to bet it would work here. I'm also not as familiar with Quick's build system as Nimble's.

It looks like the [most recent master build](https://travis-ci.org/Quick/Quick/builds/128508913) has the same two other failures as this PR, the ones using xctool, so I'd wager these failures are unrelated to this PR. In that case, I think we should merge this PR and verify that the master `podspec` test passes. If not, we can revert.

pedrovereza:
@ashfurrow thanks for jumping in!

I squashed the commits to remove noise and to make reverting easier if needed. Let me know if more changes are needed from me.

briancroom:
I agree with @ashfurrow. Let's get this in and see how CI behaves when it's on master. 

Thanks @pedrovereza!

briancroom:
Unfortunately it looks like the same error is happening on master as well, twice in a row. 

Do you have any further ideas about how to work around these issues @ashfurrow?

ashfurrow:
Looks familiar, I'll take a look when I can.

ashfurrow:
Opened #530 to address.

PR #530, opened 2016-05-15
Adds ruby specifier.

ashfurrow:
Continued from #529. Trying to fix the CI podspec linting job.

Comparing it to https://github.com/Quick/Nimble/pull/288 , the only thing that jumped out at me was that on Nimble's PR, we specified system ruby. I remember there being weirdness with generic os Travis builds. We'll see.


ashfurrow:
Looks like it got passed the `bundle install` step on https://travis-ci.org/Quick/Quick/jobs/130391984 which is where it's failing in master.

ashfurrow:
Well, the actual _linting_ failed, but the previous error of failing bundle installation (the problem this PR aims to fix) is gone. Maybe merge this, and open a new issue to address the failing lint? FWIW, the listing fails locally too.

briancroom:
Sounds like the right action plan @ashfurrow. Thanks for taking the next steps here!

ashfurrow:
Here's the local error:

> ld: framework not found XCTest for architecture x86_64

briancroom:
(Oops wrong button 😆)

Yeah I thought that might be the error. The linter is attempting to build an app that links to Quick which in turn links to XCTest, however that framework is located at a non-standard search path, so apps can't link to it by default. I imagine we will need to modify the podspec to add the necessary Framework a Search Path to Quick's build settings to get that working.

ashfurrow:
Cool, let's fix in another PR?

briancroom:
Definitely. Merging this now as it moves us in the right direction!

PR #531, opened 2016-05-16
Adds search path in podspec for XCTest

pedrovereza:
Continued from #529 and #530 

As pointed out by @briancroom in #530, podspec needs a Search Path in order to locate XCTest.

### DIFF
--- a/Quick.podspec
+++ b/Quick.podspec
@@ -33,5 +33,6 @@ Pod::Spec.new do |s|
 
   s.framework = "XCTest"
   s.requires_arc = true
+  s.xcconfig = { 'FRAMEWORK_SEARCH_PATHS' => '$(PLATFORM_DIR)/Developer/Library/Frameworks' }
   s.pod_target_xcconfig = { 'ENABLE_BITCODE' => 'NO' }
 end


pedrovereza:
@ashfurrow @briancroom Now we got it! :wink: https://travis-ci.org/Quick/Quick/jobs/130557759

modocache:
Looks like all the right CI builds pass. Thanks, @pedrovereza! 👍

phatblat:
Nice work!

briancroom:
🎉

There *may* be a case to make that this is a Cocoapods bug and the search path should be handled automatically. What do you think @ashfurrow @orta?

ashfurrow:
Hmm, seems strange that's not included, but I've not heard of it affecting other, non-testing frameworks like UIKit. @mrackwitz the expert on frameworks – Marius, do you know if this is a bug we should open on CocoaPods?

briancroom:
This old PR seems relevant to the discussion https://github.com/CocoaPods/CocoaPods/pull/2332

mrackwitz:
There is an issue currently, which prevents testing pods to lint with CocoaPods 1.0. See https://github.com/CocoaPods/CocoaPods/issues/5321.
What you came up here with seems like a workaround for this problem, or where you trying to fix any other "real world" issues with that?
Please note though that `xcconfig` is deprecated and shouldn't be used anymore. It was split in `pod_target_xcconfig` and `user_target_xcconfig`. You would want to use the former instead, because the test targets have already XCTest in their framework search paths by default.

pedrovereza:
> Please note though that xcconfig is deprecated and shouldn't be used anymore. It was split in pod_target_xcconfig and user_target_xcconfig. You would want to use the former instead, because the test targets have already XCTest in their framework search paths by default.

Interesting, I tried to use `pod_target_xcconfig` [as used in Nimble](https://github.com/Quick/Nimble/blob/master/Nimble.podspec#L21) but `pod lib lint` still failed on trying to find XCTest (not sure if it relates to https://github.com/CocoaPods/CocoaPods/issues/5321)
`ld: framework not found XCTest for architecture x86_64`.

I only managed to get it running locally using `xcconfig`, but as I don't use cocoapods that often I may have missed something.

mrackwitz:
My fault. You need only `user_target_xcconfig` meanwhile to workaround the problem.

pedrovereza:
Oh, got it working with `user_target_xcconfig`:

````
s.user_target_xcconfig = { 'FRAMEWORK_SEARCH_PATHS' => '$(PLATFORM_DIR)/Developer/Library/Frameworks' }
s.pod_target_xcconfig = { 'ENABLE_BITCODE' => 'NO' }
````
Quick question (pun intended): is it safe to combine both options in `user_target_xcconfig` or can this cause any side effect?

````
s.user_target_xcconfig = { 'ENABLE_BITCODE' => 'NO', 'FRAMEWORK_SEARCH_PATHS' => '$(PLATFORM_DIR)/Developer/Library/Frameworks' }
````

mrackwitz:
No, I don't think you want to combine these both. The build settings  added to `user_target_xcconfig` are made available only to the test target of the user. The build settings in `pod_target_xcconfig` are added to the generated pod target for Quick.

pedrovereza:
I had a strong feeling combining was wrong, just wanted to confirm.

Thanks for the help, @mrackwitz 👍

PR #533, opened 2016-05-18
Replaces deprecated property in podspec (changes xcconfig to user_target_xcconfig)

pedrovereza:
As called out in #531, `xcconfig` is deprecated and `user_target_xcconfig` should be used instead.


phatblat:
💯

PR #540, opened 2016-05-31
Add file headers to the Xcode templates.

bojan:


istx25:
Thank you for the contribution, @bojan.

PR #542, opened 2016-06-07
Update ArrangeActAssert.md

brunodlz:
I think that not necessary to use 'else' this example.


modocache:
Thanks @brunodlz! This is probably a matter of preference. I personally like the `else`, it makes it clear to me that it's one or the other... so I'll opt not to take this change for now. Thanks anyway! 👍

PR #543, opened 2016-06-13
Adjust a format to the Markdown syntax

JPMartha:
Adjusted a format to the Markdown syntax by removing an extra whitespace.


modocache:
Thanks, @JPMartha!

JPMartha:
@modocache My pleasure! :)

PR #547, opened 2016-06-14
Compile under Xcode 8 beta 1

sharplet:
See https://github.com/Quick/Nimble/pull/302. Details are also in the commit message!

Currently the tests are failing in Xcode 8. I haven't bothered to investigate (seems related to XCTestCase / XCTestSuite integration), so far this is more about getting Quick to compile under Xcode 8.


sharplet:
The tests appear to be passing now. Currently the changes to the test target aren't backwards-compatible with Xcode 7.3. However the rest of the changes should be backwards-compatible (barring any issues with the deprecated API being removed).

modocache:
Awesome work, @sharplet!! Non-backwards compatible changes, huh? Do you think we should branch somehow? Perhaps an Xcode 8 branch and separate releases until Xcode 8 is officially released?

Sources/QuickTests/Helpers/QCKSpecRunner.m
```diff
@@ -12,11 +12,6 @@
         if ([suite respondsToSelector:@selector(runTest)]) {
             [suite runTest];
             result = suite.testRun;
-        } else {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-            result = [suite run];
-#pragma clang diagnostic pop
```
modocache: Yup, looks like Apple XCTest removed this API; see https://github.com/apple/swift/pull/2645. But don't we need to call the newer API, `-[XCTest runTest]` here? I think we can then grab the result off of the test case object.

modocache:
My suggestion: create a `xcode-7` branch. We'll make v0.9.* releases from that branch. `master` will be used for Xcode 8 and up.

Looks like Travis CI already supports Xcode 8: https://docs.travis-ci.com/user/osx-ci-environment/#Xcode-8. Let's use this to upgrade our CI.

Thoughts? 🙌

sharplet:
@modocache I'm actually thinking that at this stage, the amount of effort to support _both_ Xcode 8 and Xcode 7.3 (on Swift 2.x) is small enough that it might be worth it to cut a patch release for Xcode 8 support with no other changes.

I'm saying this mainly from the point of view of someone who wants to try and maintain source compatibility with both 2.x and 3.0 if possible, in a [project of mine](sharplet/Regex).

modocache:
Makes sense to me! We can always figure out some sort of branching strategy later.

Now, should we update the `.travis.yml` in this pull request to use Xcode 8?

sharplet:
I just got around to testing out some workarounds that were suggested on Twitter: https://github.com/Quick/Nimble/pull/302#issuecomment-226376805. tl;dr: one of the workarounds did work, but it requires raising all our deployment targets which isn't really viable.

So I think we should be able to go ahead with the workaround as implemented, however I just realised that Xcode 8 + Swift 2.3 + tvOS appears broken: https://github.com/Quick/Nimble/pull/302#issuecomment-226375515.

sharplet:
@modocache I've fixed up the handling of the old `-[XCTest run]` API by just redeclaring it for Xcode 8. This is harmless on new deployment targets which will continue to use `runTest`/`testRun`, but will allow continued support for the 7.0 deployment target (which btw should probably be revisited for the next breaking release, say alongside Swift 3).

I've also added runtime checks for the test observation category to allow it to work for both Xcode 7.3 and Xcode 8.

sharplet:
I haven't been able to reproduce the CI failures so far on my machine. I've kicked off the build again, so hopefully better luck this time 🙏

sharplet:
So here's a summary of the failing jobs on CI:

- The Xcode 8 tests involving simulators. They seem to hang waiting for the simulator to launch, which seems like a Travis bug not an Us bug.
- All the xctool ones.

The Xcode 8 iOS and tvOS tests run fine locally, so I'm personally in favour of merging and tagging a new release. @Quick/contributors?

modocache:
Yes, let's merge first and fix CI later. Thanks for all the hard work here, @sharplet! ❤️

ikesyo:
I opened https://github.com/sharplet/Quick/pull/1 for beta 2 support.

sharplet:
Ok, so this is looking good to me now. The current failures on CI appear to be related to the `xcode8` image not yet being updated to beta 2; all the tests pass under beta 2 on my machine.

If someone else could verify that this all passes on their machine, I think we're ready to go!

sharplet:
Thanks @ikesyo for your help with this!

ikesyo:
Travis now supports beta 2, so let's re-run CI!

https://twitter.com/travisci/status/752704816116293632

sharplet:
Ok, so I've re-run CI. The iOS and tvOS tests are failing on Xcode 8, but they appear to be timeouts. The xctool tvOS build is also failing, but my guess is that's xctool-specific and can be ignored.

I think we might be good to merge 👍

modocache:
Looks good to me! Merge when ready 💯

ikesyo:
Opened https://github.com/sharplet/Quick/pull/3.

ikesyo:
One more thing: https://github.com/sharplet/Quick/pull/4

sharplet:
Closing in favour of #565.

PR #549, opened 2016-06-15
Use `appletvos` for `SDKROOT` build setting in Quick-tvOS

### LINKED ISSUE #548: There is no `appletvsimulator` SDK in Xcode 8
We cannot build Quick-tvOS at the branch of #547 (that was introduced in #524).

> warning: target specifies SWIFT_VERSION = '2.3', but it is overridden by TOOLCHAINS = 'com.apple.dt.toolchain.AppleTVOS10_0'

`SWIFT_VERSION=2.3` seems to be overridden. Changing to `appletvos` (and adding missing `-fobjc-arc` flag to the `.m` files) should fix the build problem.

ikesyo:
Xcode 8 claims that `appletvsimulator` is not found for the setting. Fixes #548.

appletvsimulator


modocache:
Thanks, @ikesyo! 💯

Issue #545, opened 2016-06-13
Remove SwiftPM dependency on Nimble?


modocache:
![screen shot 2016-06-13 at 3 38 32 pm](https://cloud.githubusercontent.com/assets/552921/16020656/0d232b54-317d-11e6-8501-13e6fcc98290.png)

It's unfortunate, but most [Swift package indexes list Quick as having a dependency](https://swiftpkgs.ng.bluemix.net/package/Quick/Quick): Nimble. This is not true, of course -- Quick isn't dependent upon Nimble, it simply uses Nimble as part of the test suite that tests Quick.

I'd prefer Quick be listed without any dependencies, but as far as I know the only way to do so (that we can do ourselves) is  by modifying `Package.swift` somehow, or by not using Nimble as part of the test suite. However, I like testing Quick and Nimble together, because I think many people use them together.

Is it possible to define a `Package.swift` that does not list Nimble as a dependency?

takecian:
I saw comment in Package.swift and found following,
```
// TODO: Once the `test` command has been implemented in the Swift Package Manager, this should be changed to
    // be `testDependencies:` instead. For now it has to be done like this for the library to get linked with the test targets.
    // See: https://github.com/apple/swift-evolution/blob/master/proposals/0019-package-manager-testing.md
    dependencies: [ 
        .Package(url: "https://github.com/briancroom/Nimble", majorVersion: 3)
    ]
```

If my understanding is correct, we need to use future functionality to remove dependency on Nimble.

https://github.com/apple/swift-evolution/blob/master/proposals/0019-package-manager-testing.md#test-only-dependencies

modocache:
Until that functionality is present, though, can we remove the `dependencies` altogether? Does SwiftPM build our test suite as well? Can we disable that?

briancroom:
SwiftPM has matured tremendously since I was last actively working on this and wrote that note. There's a pretty good chance that `testDependencies` is functional now!

takecian:
I tried using `testDependencies` on Travis CI, but it doesn't work.
I think it depends on Swift version on Travis CI. I'll check.

takecian:
I updated swift version, and following error occured.

```
t-fujiki$ swift build
error: mixedSources("/Users/t-fujiki/Development/iOS/Quick/Sources/Quick")
```

I think that we have to support Swift3 in order to use `testDependencies `.
What do you think?

ikesyo:
[...]

PR #510, opened 2016-04-02
[WIP] Swift 3.0 Compatibility

norio-nomura:
- Uses wrapper that converting from Swift 3.0 API to Swift 2.2 API.
- Includes #497
- Uses swift-DEVELOPMENT-SNAPSHOT-2016-03-24-a
- Depends on https://github.com/Quick/Nimble/pull/277. `Package.swift` uses my fork until that merged.
- [x] Re-organize for `swift test`. Can we do that?

PR #553, opened 2016-06-16
[WIP]Update dependency on Nimble with testDependencies

takecian:
Discussed in #545 , I replaced dependencies with testDependencies.


takecian:
I got new error, `error: the module at Sources/QuickTests has an invalid name ('QuickTests'): the name of a non-test module has a ‘Tests’ suffix`.

I think this error occured due to [Proposal: SE-0019](https://github.com/apple/swift-evolution/blob/master/proposals/0019-package-manager-testing.md), we have to change directory name of test code.

ikesyo:
> we have to change directory name of test code

That is already done in #510.

ikesyo:
Unfortunately, `testDependencies` are removed in recent development snapshots 😢 : https://github.com/apple/swift-package-manager/pull/597

takecian:
Ok..I'll close this..

PR #563, opened 2016-07-15
[Feature] Add a Chinese version document

LinShiwei:
A Chinese version of [BehavioralTesting.md](https://github.com/Quick/Quick/blob/master/Documentation/en-us/BehavioralTesting.md).


istx25:
@modocache: Do we have anyone in the Quick community who understands/reads Chinese? It would be nice to get a few eyes on all documentation before it gets merged in.

modocache:
Yeah, my Chinese isn't up to par to read this. I'd be fine merging and having others review. If we could find someone to "own" the Chinese guides (and give them commit access to make that easier), that'd be awesome.

istx25:
Good idea. I'll send a tweet out, but it'd be awesome if you could too. Especially since we plan to put up a separate website for the documentation, it would be nice to have a few people to maintain the documentation to ensure it's always up to date and nice to read.

istx25:
🎉

PR #576, opened 2016-08-24
Add a test to Quick's test suite that verifies afterSuite is called

### LINKED ISSUE #561: Add a test to Quick's test suite that verifies afterSuite is called
Reproducing steps:
1. Checkout 5fb172b
2. Set break point at `AfterSuiteTests.swift:10` and `AfterSuiteTests+ObjC.m:12`
3. Run test

Expected result:
4. Debugger stopped at break points set on step 2

Actual result:
4. Debugger did not stop at break points set on step 2

takecian:
Fixes #561 
## What this PR do
- Add new test target `QuickAfterSuiteTests`
  - `QuickTests` target uses their own TestSuite, in this case, afterSuite not called. So I add new test target.
- Verify that `afterSuite` called. To do this, I use `class func tearDown()`, this method is called after all test in a class executed.
- Remove afterSuite test in `QuickTests`, as currently this code does not verify `afterSuite` called.


istx25:
Hi @takecian! Sorry for the delay in getting back to you about this. Would you be able to fix the merge conflicts and then I will review this PR asap?

takecian:
ok, i will try to resolve conflict..

Package.swift
```diff
@@ -3,6 +3,7 @@ import PackageDescription
 let package = Package(
     name: "Quick",
     exclude: [
+      "Sources/QuickAfterSuiteTests",
```
istx25: Any reason why? Let's remove that and then we can get this merged!

takecian: @istx25 
I cannot remember why i excluded this test..
Anyway, @ikesyo san fixed CI failure at #639 , could you review that PR?

PR #594, opened 2016-09-01
[Documentation] Add Troubleshooting.md in ja

takecian:


takecian:
Self merged, as only documents changed.

PR #595, opened 2016-09-01
[Documentation] Add trouble shooting link in README

takecian:


takecian:
Self merged, as only documents changed.

PR #603, opened 2016-09-04
Use Carthage for easier submodule management

ikesyo:


modocache:
Could you add to the commit message what is easier about using Carthage? I am a bit skeptical, probably because I'm a big fan of Git submodules. Is Carthage really easier? 🤔

This commit also needs updates to the instructions on how to contribute to the project, probably [in `CONTRIBUTING.md`](https://github.com/Quick/Quick/blob/580148d8a869dbef57fd664a14281d41ec1d3cff/CONTRIBUTING.md#building-the-project) or somewhere.

ikesyo:
> Could you add to the commit message what is easier about using Carthage? I am a bit skeptical, probably because I'm a big fan of Git submodules. Is Carthage really easier? 🤔

Switching to a forked repository is tedious work with plain git commands for me. Additionally, it isn't so easy to get what version of the dependency is used. Makes sense?

> This commit also needs updates to the instructions on how to contribute to the project, probably in CONTRIBUTING.md or somewhere.

This still manages the dependency with git submodule, so the `Building the Project` section is not affected I think (thanks to `--use-submodules` option of `carthage update/bootstrap/checkout` commands. And also, if there are existing submodules in `Carthage/Checkouts` directory, Carthage automatically checkout dependencies using git submodule without the option).

modocache:
> Switching to a forked repository is tedious work with plain git commands for me.

Ah, yes, very true. This is especially important because of how much volatility there is around Swift versions, etc. If Carthage makes this easier, that's great!

> the Building the Project section is not affected I think

I guess what I mean is, since I haven't used Carthage in a few years, I don't know how these changes affect the project. I also wouldn't know how to troubleshoot any issues. Adding documentation somewhere would help me (and maybe others) fix issues with the new setup, should they arise.

ikesyo:
Reconsider this and this introduces a new barrier for new contributors which is essentially not needed, so I withdraw this. Sorry for the confusion and thanks for your thoughts! :pray:

PR #606, opened 2016-09-08
Good bye to Circle CI, keeping Travis only

### LINKED ISSUE #586: Remove one of the continuous integration platforms
I don't see the point of having multiple CI platforms on the Quick project. I think we should remove one of them and fix the remaining one to work as effectively as possible.

1. Do people agree with me?
2. Which one(s) should we kick and which one should we keep?

pedrovereza:
This fixes #586


modocache:
Thanks!! ❤️

istx25:
Thanks!!

PR #615, opened 2016-10-03
[Documentation] Add Swift version in README

takecian:
Describe Swift version and that supports Quick and Nimble release version.

PR #616, opened 2016-10-03
[README] Fix Swift version descriptions

ikesyo:
Improvements to #615.


modocache:
Looks great, thanks! I also think it'd be nice to include a paragraph above, explaining the table a bit. Something like this:

> Certain versions of Quick and Nimble only support certain versions of Swift. Depending on which version of Swift your project uses, you should use specific versions of Quick and Nimble. Use the table below to determine which version of Quick and Nimble are compatible with your project.

ikesyo:
Thank you @modocache! I added it as is: 5ef789e

modocache:
Awesome, thanks! :)

PR #617, opened 2016-10-04
Adds link to Swift version in installation doc

pedrovereza:
There is also a section on using Swift 1.2 which I'm not sure if it's worth keeping (maybe the versions needed for 1.2 should be added on README?)

Thoughts?


wongzigii:
+1 for dropping this outdated doc.

takecian:
+1

PR #628, opened 2016-10-21
Update function prototype in docs for shared examples

rastersize:
It was not fully up-to date for Swift 3 GM and as such would fail to compile if compiled as-is.


istx25:
Thank you for making this change!

PR #629, opened 2016-10-26
[Package.swift] Don't lock Nimble version to exact v5.0.0

ikesyo:
See also https://github.com/ReactiveCocoa/ReactiveSwift/pull/68.


istx25:
LGTM

PR #637, opened 2016-11-06
[SwiftPM] Remove test dependencies from `Package.swift`

ikesyo:
We still run `swift test` on CI using separated `.Package.test.swift`.

This is the same as https://github.com/Carthage/Commandant/pull/83.


wongzigii:
Look good to me. But will defer to other to merge. 😃

istx25:
Looks good to me.

PR #638, opened 2016-11-10
Fix resolve module name when test bundle path contains a space (#557)

### LINKED ISSUE #557: Quick can't recognize a moduleName for Unit test targets containing a space in the name
Hi! I had an issue trying to run a single test from the test navigator: 0 tests were performed.  
I searched for existing issues and found #534 (after debugging seems unrelated - no space in the file path).  
I tried to debug the issue and looks like in `testCaseClassForTestCaseWithName(name:)` Quick fails to resolve the module name and returns nil. This is caused by my test target having a space in the name (Kakapo iOSTests), the path of the bundle contains a space: "/path/Kakapo iOSTests.xctest" while the module name (created by Xcode) is `Kakapo_iOSTest`. NOTE: `String+Filename` fails because of the space but the module name wouldn't be correct even if the space was escaped.

I can create a PR but I'm not sure what's the proper behavior:
1. Space in the target name (actually in the bundle path) should be avoided -> then we might just add an assert
2. Since Xcode use underscore `_` for the module name instead of spaces we might just replace it (kind of bad)
3. Another proper way to identify the module name? The module name might be completely different than the bundle filePath (if you specify a different module name than the default it won't work)

I know the preferred way to run a single test is `fit` but I would prefer a crash or failure if something goes wrong when running the tests from the navigator, running 0 tests successfully is a misleading result.

MP0w:
Closes #557 

I couldn't find any test that cover this stuff but looks like hard to test.
Let me know if something is wrong.


istx25:
Looks good to me. Thanks @MP0w.

PR #644, opened 2016-11-22
Updates README

orta:
Originally I just wanted to update the README to CocoaPods 1.0 syntax, but I figured you could use CP numbers too

When you run `pod init` on a project with unit + integration tests, it would look like this:

``` ruby
# Uncomment the next line to define a global platform for your project
# platform :ios, '9.0'

target 'MyApp' do
  # Comment the next line if you're not using Swift and don't want to use dynamic frameworks
  use_frameworks!

  # Pods for MyApp

  target 'MyAppTests' do
    inherit! :search_paths
    # Pods for testing
  end

  target 'MyAppUITests' do
    inherit! :search_paths
    # Pods for testing
  end

end
```

But we can go a bit better,

```ruby
# Podfile

use_frameworks!

target "MyApp" do
  # Normal libraries

  abstract_target 'Tests' do
    inherit! :search_paths
    target "MyAppTests"
    target "MyAppUITests"

    pod 'Quick'
    pod 'Nimble'
  end
end
```

Which fixes the DRY problem that the method was fixing in the current README.

---

I checked how many downloads you have too, both Quick and Nimble have over 10k unique installs on CocoaPods - so I set that as your baseline.

Quick
screen shot 2016-11-22 at 10 01 34

Nimble
screen shot 2016-11-22 at 10 01 26

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,23 +53,23 @@ All documentation can be found in the [Documentation folder](./Documentation), i
 
 use_frameworks!
 
-def testing_pods
+target "MyApp" do
+  # Normal libraries
+
+  abstract_target 'Tests' do
+    inherit! :search_paths
+    target "MyAppTests"
+    target "MyAppUITests"
+
     pod 'Quick'
     pod 'Nimble'
-end
-
-target 'MyTests' do
-    testing_pods
-end
-
-target 'MyUITests' do
-    testing_pods
+  end
 end
 ```
 
 ## Projects using Quick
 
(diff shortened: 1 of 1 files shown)


orta:
and congrats on the 1.0

istx25:
Thanks @orta for submitting this PR! We appreciate it. It's also amazing to see how many use Quick and Nimble with CocoaPods. We knew there were a lot (at least around 5,542 - GitHub stars) but it's cool to see the specific data. 💯 ❤️ I didn't even know that CocoaPods shared that data! Thanks so much for adding it.

modocache:
Wow, ten thousand! Super cool. Thanks for the info, @orta!

Issue #651, opened 2016-12-03
None of the tests are executed anymore


benediktreichert:
- [X] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Installed Quick and Nimble via Cocoapods.
Wrote some tests. Built and ran them.

### What did you expect to happen?

Every test has to be executed each time i run the tests

### What actually happened instead?

All tests (including standard XCTests) are simply skipped and ignored after their first run. The get only executed one again when I change something in the source code, so the file has to be recompiled. I need to execute the test twice to be noticed. After execution, the test is then ignored again. 

### Environment

List the software versions you're using:

 - Quick: 1.0.0
 - Nimble: 5.1.1
 - Xcode Version: *8.1* (Open Xcode; In menubar: Xcode > About Xcode)
 - Swift Version: Standard *3.0.1*
```
$ swift --version
Apple Swift version 3.0.1 (swiftlang-800.0.58.6 clang-800.0.42.1)
```

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - Cocoapods: *1.2.0.beta.1* (Use `pod --version` in Terminal)

### Project that demonstrates the issue
[My complete project](https://github.com/benediktreichert/moodleinfo_ios)

# Additional notes
My Quick tests get recognised by Xcode as "runtime Tests"
See here: http://stackoverflow.com/questions/19699730/xcode-5-testing-symbol-rt-means-what
I have also this purple `rT` symbol on my Quick tests and they got only recognized after their initial launch or when i change the source code of a test so it has to be recompiled.

I also deleted the `Derived Data` folder and `xcuserdata` folders.

danielsaidi:
I have the exact same problem. I have a large amount of tests in a project of mine, running Quick 0.10.0 and Nimble 5.0.0. The tests all run without problem.

Today, I created a new project and decided to use Quick 1.0.0 and Nimble 5.0.0. The initial test example ran just fine, but as soon as I added some general extensions and their tests from the old project to the new one, the test in the new project stopped running.

After commenting out all new tests, the initial test example started working again, so I started re-adding the new tests piece by piece.

For me, what made the tests fail was that one test file contained the QuickSpec class as well as some private classes used by that test. Once I moved the private classes to a separate file, the tests began to work once more.

Have a look at the two screenshots below - one is with the private classes inside the test class file, one is with the private classes moved out to a separate file:

[...]

Issue #534, opened 2016-05-20
Xcode says it has run the tests but not even one


shravankumarbj:
Hi,

Its strange that when I open Xcode fresh and run a test file whole unit tests in project will run with the file tests also. And if I focus or without focussing run a file again Xcode says it has run the test and passes, even though last run had failed test and it is still failing Xcode says it has passed, but it doesn't reach any break point in the tests written. I am using Xcode 7.3.1. I am using package managers but not with respect to any testing.

I really like Quick and nimble which helps me to write BDD tests but now I am in a situation that I am not able to fix the previous failing tests due to the respective main code change.

Can any one help me on these

shravankumarbj:
Any help on these.??? Or any link where I can rectify this????

modocache:
Oops, sorry for the slow response! Just to clarify: are you running the entire test suite (⌘U)? Or just a single test file via the Xcode test navigator?

Both should work, but I have seen reports of single test execution causing trouble in the past -- see #373. I would recommend always running the entire suite. To run only one test, use `fit`.

If you're running the entire suite, but still experiencing problems, could you provide more detail on your project? A sample project that demonstrates the problem would be best. As-is, the setup you describe works for me, so I'm not sure how to help.

dtrenz:
Same issue here. If I can't get it sorted, I will try to get an example project up to help reproduce.

dtrenz:
@modocache — Example project that reproduces the issue for me: https://github.com/dtrenz/QuickExample

dtrenz:
OK, this is getting strange. It's working this morning. I haven't restarted my machine, or even Xcode. It just started working again. Huh.

modocache:
If you can narrow it down, that would be very much appreciated. I have a feeling this is going to come down to some very interesting Apple Radars.

shravankumarbj:
Its happening same to me. I cannot create example project becoz of time constraint. If I restart XCode it will test it otherwise it says all tests passed but it will not even one test passed or it will hit any break point. If I run one QuickSpec file it runs all unit tests entire suite. I am pretty sure its Xcode problem but not able to identify whats the problem.

hirad:
I just started hitting this issue. Previously I was having the problem where using the Xcode test UI would not pick up on individual test cases but ⌘U would work. Now even ⌘U is not picking up on any of the tests. Usual voodoo magic of clean, delete derived data, restart Xcode did not work.

acecilia:
Hi,
[...]

PR #652, opened 2016-12-04
Fix Quick unable to find module name when bundle name contains `-` #534

MP0w:
help you if you're not sure:

 - What behavior was changed? Running test by selecting in the Xcode test navigator is now working for test bundles containing dashes in the name
 - What code was refactored / updated to support this change? -
 - What issues are related to this PR? Or why was this change introduced? #534 #557 #651 

 - [x] Does this have tests? Yes [See comment](https://github.com/Quick/Quick/pull/652#issuecomment-265172849)
 - [x] Does this have documentation? Yes
 - [x] Is this a new feature (Requires minor version bump)? No

Note: #557 fixed the issue with spaces but apparently there are other character causing troubles. Hopefully this are the only two commonly used in bundle names.


Sources/Quick/NSBundle+CurrentTestBundle.swift
```diff
@@ -11,6 +11,16 @@ extension Bundle {
...
+    /**
+     Return the module name of the bundle.
+     Uses the bundle filename and transform it to match Xcode's transformation.
+     Module name has to be a valid "C99 extended identifier".
+     */
+    internal var moduleName: String {
+        return bundleURL.fileName
+            .replacingOccurrences(of: " ", with: "_")
+            .replacingOccurrences(of: "-", with: "_")
+    }
```
modocache: Can you reuse [the same mechanism as the one we use for test names](https://github.com/Quick/Quick/blob/0fdf273c7c1463a5f25c3942587452b1e892918c/Sources/QuickObjectiveC/NSString%2BQCKSelectorName.m#L5-L34) here? Or is there a reason we can't?

MP0w: oh, I didn't know about that category.... looks like good for module name as well.
Should I rename it to `String.c99ExtendedIdentifier` (assuming it is)

MP0w:
@modocache adding another bundle can be a way but then we should do the same for space and we quickly end up with many targets I guess. Do we want this?
I wouldn't test if test were performed (I guess should be tested somewhere else?).
A unit test for this issue should check that a (swift) class defined in a bundle that contains a space/dash in the name can be retrieved at runtime (module name is correct).

Sources/Quick/NSBundle+CurrentTestBundle.swift
jeffh: Renaming + reusing that method sounds great.

jeffh:
I'm ok with adding another test bundle that has many invalid characters. I think we don't want to regress from it.

MP0w:
I pushed some changes to use `c99ExtendedIdentifier`. I converted it to swift because we can't use a riding header.
For the test I currently test the behaviour with the existing test bundle that contains a `-`.
I tried to add another test bundle but I'm not exactly sure how I should do it, If we want to add a test in this new test target which contains special characters then we would have to create 3 targets right? (iOS, OSX, tvOS).
Would it be enough to test it in the current targets + test that `c99ExtendedIdentifier` works also with spaces, etc ?

jeffh:
Thanks @MP0w, I've merged the current work as is. I would ok with just one extra test bundle on iOS that has that if you have time. Otherwise, it'll be part of the work again if a regression occurs.

PR #661, opened 2016-12-30
Remove unused code.

drodriguez:
Some unused code was present in the Objective-C QuickSpec class. The resulting `sanitizedFileName` was used, but then stop being used in 08b4bee51441ee64ab91794bb48076f26a9b0ef8.

The code should work exactly the same as before (maybe slightly faster). No tests nor documentation, but the change doesn’t modify the behaviour.


istx25:
LGTM. Thanks for filing this PR, @drodriguez.

PR #662, opened 2016-12-30
Change test method encoding to return void.

drodriguez:
The implementation above uses a block that return void.

Probably the code from XCTestCase doesn't try to inspect the return
value, but it is probably better to include the right type
information just in case.

No new tests or documentation. The change is internal and should not affect the current behaviour. Test pass.


istx25:
LGTM. Thanks for filing this PR, @drodriguez.

Issue #665, opened 2017-01-02
Advertise limitations of Quick Xcode integration


alper:
I've run into this in a project I'm entering into and it could be nice if Quick was up front about the problems with integrating with XCode. Testing in Xcode is bad enough as it is and a framework that does not integrate with the environment is pretty problematic.

This seems to be the blocked issue underneath: https://github.com/Quick/Quick/issues/219

modocache:
I've been pretty up-front about it when I discuss using Quick with people in person. I tend to focus on Quick's drawbacks, but many have told me they love using Quick, and don't care about some of its limitations. They use `fit` and `fdescribe` to run individual sets of tests, so they don't mind that Xcode doesn't produce an index of tests before they're run.

In any case, I'd be happy to review a pull request that has the exact wording you'd like to see. "Does not integrate with the environment" might be stronger language than Quick deserves, though. As I mention above, many people don't appear to see it that way. Perhaps something like "does not fully integrate", or "has limitations." It's not as if it *doesn't work*, it's more that Apple doesn't provide us with the tools we'd need in order to integrate 100%. (If you'd like to encourage them to provide us with these, please duplicate [this radar](http://openradar.appspot.com/radar?id=4974047628623872). Friends at Apple tell me that many people have duplicated this so far, and the Xcode teams are beginning to notice this issue.)

PR #667, opened 2017-01-05
Added a comment about Xcode limitations

alper:
As per this discussion: https://github.com/Quick/Quick/issues/665


istx25:
Thanks for making this change, @alper.

istx25:
Fair. @alper, mind filing another PR with that change?

alper:
Thanks for the merge. I think if people are annoyed enough by it, they will read the linked issue and take action?

jeffh:
Yep. That's the best way to get proper integration into Xcode.

PR #668, opened 2017-01-06
Update QuickConfiguration template

ikesyo:
This is missed for Swift 3 update.


istx25:
LGTM.

istx25:
Thanks for making that change, @ikesyo. :-)

Issue #641, opened 2016-11-15
Make Quick SwiftLint compatible


jeffh:
See [SwiftLint](https://github.com/realm/SwiftLint).

To get our code to be more consistent, we should add SwiftLint to CI. Before we can do that, we'll need to convert each source file inside Quick. We should also do this with Nimble.

To avoid causing conflicts with active PRs, it's probably safer fix the linting on a file and send that as a PR. Alternatively, we can just have a formatting pass as one massive PR to allow other PRs to resolve any conflicts all at once. Thoughts?

Rules left to resolve:
 - [x] line_length
 - [x] type_name
 - ~~valid_docs~~
 - [x] empty_count
 - [x] force_unwrapping
 - [x] function_body_length
 - [x] ~~variable_name~~ identifier_name

FranciscoAmado:
Hi,
I've seen some warnings related to `syntatic_sugar` and `trailling_comma`
Shouldn't these be also prevented or disabled? They tend to pile up.

Thanks for building Quick,
Francisco

lferro9000:
I only found 2 warnings by checking master (PR done). If more work required on this, i'm happy to contribute, but need the full file with all the offending rules set as desired by the project maintainers.

ikesyo:
Done ✅

PR #671, opened 2017-01-09
[SwiftLint] Enable some disabled rules

ikesyo:
There is no violations for the rules. Related to #641.


wongzigii:
Thanks @ikesyo !

PR #674, opened 2017-01-11
Change iOS Deployment Target to 7.0

### LINKED ISSUE #541: MinimumOSVersion invalid.
Just FYI that when we're downloading with Carthage and building and running w/ Fastlane or with Xcode Application Loader, BOTH Quick AND Nimble trigger an error because the MinimumOSVersion field is set to 7.0 w/ Xcode 7.3. With that version, the MinimumOSVersion field must be set to 8.0 or greater, which forces me to check in built and modified frameworks to my repo to avoid Carthage being executed by CI and breaking the frameworks.

jeffh:
Causes a warning in Xcode, but users still have minimum deployment targets at 7 that still work for tests.

Fixes #541.

---------------

 - [ ] Does this have tests? **No. (Ideally, we should have some verification to prevent regression since this is the 3rd-time).**
 - [x] Does this have documentation? No.
 - [x] Does this break the public API (Requires major version bump)? No.
 - [x] Is this a new feature (Requires minor version bump)? No.

### DIFF
--- a/Quick.xcodeproj/project.pbxproj
+++ b/Quick.xcodeproj/project.pbxproj
@@ -2005,7 +2005,7 @@
 				);
 				INFOPLIST_FILE = Sources/Quick/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 7.0;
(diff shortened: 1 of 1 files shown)


ikesyo:
It seems that what @cocotutch said in #541 is bumping deployment target to 8.0 to silence the Xcode warning and that was done in this commit: https://github.com/Quick/Quick/commit/83eb4897ebae25093a0c39a775424de0d797bd2f.

So this pull request does not fix #541 but actually reopen it.

jeffh:
Ah, hmm I totally misread that issue. Thanks @ikesyo. Closing.

PR #645, opened 2016-11-29
Expose current spec for XCTest expectations

pcantrell:
This provides a static `QuickSpec.current()` method which allows specs to get a reference to the currently executing `QuickSpec`, which is essential for using XCTest expectations.

This is an updated version of #331. It fixes #330 and #578.

There was a lot of high-minded discussion back then about different approaches to this problem, and indecision lead to nothing happening. This PR is an attempt to nudge Quick toward resolving the issue. Two arguments in favor of this solution:

- It’s simple.
- It works. [Siesta](https://github.com/bustoutsolutions/siesta) has happily been using this on a fork for over a year.

If the Quick team is willing to accept this change, I’m happy to add tests and/or docs to the PR. Let me know.

 - [x] Tests
 - [x] Docs
 - [x] Breaks API: no
 - [x] New feature: yes

PR #680, opened 2017-01-20
Initial support to implicit errors

### LINKED ISSUE #403: Catch thrown errors in tests
If you are writing a test in swift 2.0 and use `throws` the test should expect throwing functions and fail automatically if the function throws.

```
func aThrowingFunction() throws -> String {
  return "something"
}

class SomeTest: QuickSpec {
  override func spec() {
    describe("hanlding throwing functions") {
      it("should compile when having function that throws") {
        aThrowingFunction() // <- this should compile. test should fail if function throws
      }
    }
  }
}
```

marcelofabri:
This adds support to catching thrown errors inside `it` blocks. Will fix #403.

Currently, you'd have to use one of Nimble's matchers such as 

```swift
it("works") {
    expect { try foo }.toNot(throwError())
}
```

With this, you'd just have:
```swift
it("works") {
    try foo
}
```

This also is similar to `XCTest`, which supports functions being marked as `throws` (and the test fails if they do throw). 

This is still WIP because I don't know how to proceed on some places (will add comments).

 - [ ] Does this have tests? Not yet
 - [ ] Does this have documentation? Not yet

### DIFF
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -85,7 +85,7 @@ final public class Example: _ExampleBase {
             try closure()
         } catch {
             let description = "Test \(name) threw unexpected error: \(error.localizedDescription)"
-            QuickSpec.current.recordFailure(withDescription: description, inFile: callsite.file, atLine: callsite.line, expected: false)
+            QuickSpec.current.recordFailure(withDescription: description, inFile: callsite.file, atLine: Int(callsite.line), expected: false)
         }
 
         group!.phase = .aftersExecuting
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -85,7 +85,17 @@ final public class Example: _ExampleBase {
             try closure()
         } catch {
             let description = "Test \(name) threw unexpected error: \(error.localizedDescription)"
-            QuickSpec.current.recordFailure(withDescription: description, inFile: callsite.file, atLine: Int(callsite.line), expected: false)
+            #if SWIFT_PACKAGE
+            let file = callsite.file.description
+            #else
+            let file = callsite.file
+            #endif
+            QuickSpec.current.recordFailure(
+                withDescription: description,
+                inFile: file,
+                atLine: Int(callsite.line),
+                expected: false
+            )
         }
 
         group!.phase = .aftersExecuting
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -84,7 +84,8 @@ final public class Example: _ExampleBase {
         do {
             try closure()
         } catch {
-            print("Test \(name) threw unexpected error: \(error.localizedDescription)")
+            let description = "Test \(name) threw unexpected error: \(error.localizedDescription)"
+            QuickSpec.current.recordFailure(withDescription: description, inFile: callsite.file, atLine: callsite.line, expected: false)
         }
 
         group!.phase = .aftersExecuting
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -30,10 +30,10 @@ final public class Example: _ExampleBase {
     weak internal var group: ExampleGroup?
 
     private let internalDescription: String
-    private let closure: () -> Void
+    private let closure: () throws -> Void
     private let flags: FilterFlags
 
-    internal init(description: String, callsite: Callsite, flags: FilterFlags, closure: @escaping () -> Void) {
+    internal init(description: String, callsite: Callsite, flags: FilterFlags, closure: @escaping () throws -> Void) {
         self.internalDescription = description
         self.closure = closure
         self.callsite = callsite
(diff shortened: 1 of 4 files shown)


Sources/Quick/Example.swift
```diff
@@ -71,7 +71,11 @@ final public class Example: NSObject {
         }
         group!.phase = .beforesFinished
 
-        closure()
+        do {
+            try closure()
+        } catch {
+            print("Test \(name) threw unexpected error: \(error.localizedDescription)")
```
marcelofabri: How to property report an error? I've tried `XCTFail`, but can't use the `callsite` information because of types (`StaticString` x `String`).

I wasn't able to get a reference to the `XCTestCase` that contains this example, but I think that would solve the issue as I could call [`recordFailure(withDescription:inFile:atLine:expected:)`](https://developer.apple.com/reference/xctest/xctestcase/1496269-recordfailure)

Tests/QuickTests/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -101,6 +101,26 @@ class FunctionalTests_ItSpec: QuickSpec {
...
+
+                func throwingFunc() throws {
+                    throw ExampleError.error
+                }
+
+                it("supports calling functions marked as throws") {
+                    try nonThrowingFunc()
+                }
+
+                it("supports calling functions that actually throws") {
```
marcelofabri: I need a way to test this test as it'll fail once the implementation is correct. I've tried using the exception handling approach, but it doesn't seem like test failures create expections. I think I'd need something like `Nimble` uses to record assertions.

Sources/Quick/Example.swift
jeffh: Hey @marcelofabri, I think the `Callsite` class will probably need to be refactored to use `StaticString` to use `XCTFail` as probably the best way to go

Tests/QuickTests/QuickTests/FunctionalTests/ItTests.swift
jeffh: Would something like how [focused tests](https://github.com/Quick/Quick/blob/master/Tests/QuickTests/QuickFocusedTests/FocusedTests.swift) are executed work? I'm not as up to speed on the Quick side of the code for bootstrapping unfortunately, perhaps other @Quick/core can answer this better.

Sources/Quick/Example.swift
marcelofabri: I've tried that, but it seems impossible because of Objective-C compatibility

marcelofabri: Once/if #645 is merged this will be trivial, as we could just use `QuickSpec.current.recordFailure`.

gperdomor:
Any progress here?

marcelofabri:
i was awaiting for #645 to be merged first, so we can use `QuickSpec.current.recordFailure(...)`

Tests/QuickTests/QuickTests/FunctionalTests/ItTests.swift
```diff
@@ -106,10 +106,35 @@ class FunctionalTests_ItSpec: QuickSpec {
...
+            func throwingFunc() throws {
+                throw ExampleError.error
+            }
+
+            it("supports calling functions marked as throws") {
+                try nonThrowingFunc()
+            }
+
+            it("supports calling functions that actually throws") {
+                try throwingFunc()
```
marcelofabri: I couldn't find a way to do it without actually failing the suite. Any ideas?

jwelton:
Any updates on this?

ikesyo:
I'll take a look.

davuthsina:
Any updates on this?

forceunwrap:
Hey @ikesyo, as you self-assigned this issue, any update on this?

ikesyo:
This should be now ready thanks to the series of pull requests: #645, #848, #850, #854, #879, #882, #883, #884, #885.

I think this is a breaking API change so it will need a major version bump (Quick v3).

marcelofabri:
Thanks for picking this up! Excited that this is ready! ✨

ZevEisenberg:
❤️ I'd love to test this out when it's ready. I was literally just looking for something like this to exist

soaurabhtokopedia:
Hello @ikesyo @jeffh, Would love to see it getting merged. Looking for this functionality lately ❤️

ikesyo:
[v2.2.1](https://github.com/Quick/Quick/releases/tag/v2.2.1) has been released so now it's time for heading to Quick v3. Merging.

marcelofabri:
@ikesyo is there a list of things we want to land in v3? i'd love to see this released as soon as possible ❤️

ikesyo:
@marcelofabri Sorry for the delay, just published! 🚢 

https://github.com/Quick/Quick/releases/tag/v3.0.0

marcelofabri:
> @marcelofabri Sorry for the delay, just published! 🚢 
> 
> 
> 
> https://github.com/Quick/Quick/releases/tag/v3.0.0

Thanks so much! 😍🎉🍾🎊

PR #688, opened 2017-02-07
[QuickSpec] Track the current example within `imp_implementationWithBlock`

ikesyo:
This eliminates Objective-C's associated objects usage and overriding `-setInvocation:`.

There should be no functional changes.

### DIFF
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -1,12 +1,9 @@
 #import "QuickSpec.h"
 #import "QuickConfiguration.h"
 #import "World.h"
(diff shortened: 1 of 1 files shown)


modocache:
This looks great, thanks @ikesyo! 🙇 

> Need to add an unit test if you’re modifying swift source

I disagree with this lint warning. Here @ikesyo is modifying the internal implementation. The fact that the tests continue to pass means, if anything, that our existing tests are pretty great.

ikesyo:
Thanks @modocache!

wongzigii:
@modocache Currently this warning is caused by https://github.com/Quick/Quick/blob/master/Dangerfile#L12. We should listen to the voice of CI instead.

PR #690, opened 2017-02-08
[gardening][QuickSpec] Inline `+invocationForInstanceMethodWithSelector:example:`

ikesyo:
That is a very simple operation now.

Just a minor refactoring.


modocache:
I think that, at the time, I thought splitting them up was easier to read. I don't have an opinion either way anymore, so this LGTM.

ikesyo:
Thanks @modocache. Yes, before #688 there was the `objc_setAssociatedObject` part so splitting them up was nice.

modocache:
Aha, so _that's_ why I wrote that! 😆

Thanks again, @ikesyo!

PR #697, opened 2017-03-03
Cache the included example count on Example.run()

### LINKED ISSUE #696: Performace issue with a large number of tests

### What did you do?

Created a bunch of tests (approximately 7k) in our application.

That the performance of the execution would be stable (in the order of 1ms per simple test).

The average time to run a simple test got into the 50ms.

### Environment

 - Quick: *1.1.0*

 - Cocoapods: *1.2.0* (Use `pod --version` in Terminal)

I didn't upload one yet, but I can create if you need it.

### Investigation

After taking a look on the _Example.run()_, on the very end, it's doing [this comparison](https://github.com/Quick/Quick/blob/e4fa1e85c0305ba4e0866f25812d3fa398f3a048/Sources/Quick/Example.swift#L85). The problem is that `world.includedExamples` is a computed property which performs a `.filter` in an array with all the tests. Which is not scaling very well when the number of tests starts grow to thousands :)

I'll send a PR that caches the results of `world.includedExampleCount` and does not calculated it over and over again. Since when the tests are run, the number of tests is already defined, I hope that this is fine.

dbarden:
Due to the fact that world.includedExamples is a computed property that does  a filter on all the tests, performing `numberOfExamplesRun >= world.includedExampleCount` is not very efficient
with a very large number of tests.

This patch caches the result of world.includedExampleCount and uses the cached value to perform the comparison.

Fix #696

### DIFF
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -1,6 +1,7 @@
 import Foundation
 
 private var numberOfExamplesRun = 0
+private var numberOfIncludedExamples = 0
 
 /**
     Examples, defined with the `it` function, use assertions to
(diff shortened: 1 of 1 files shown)


jeffh:
Thanks @dbarden!

PR #701, opened 2017-04-04
Introduce `Behavior<Context>`

mosamer:
This introduces a base generic behavior definition which can be extended with a specified context type.

Currently, shared assertions implementation using "shared examples" suffers from two drawbacks:
- Defined in `QuickConfiguration` subclass. A file that initially intended for configuration how Quick behaves, "shared examples" are added here to allow their discovery during setup time.
- Receives a context of `[String: Any]`, which is string-typed and erase any type information of the context elements.

The proposed solution is to introduce `Behavior<Context>` which will address these issues by allowing user to define subclasses for each "shared example" with a specific context type and a similar structure to `QuickSpec`

The example below demonstrates how it might be used:

```
import Quick
import Nimble

struct EdibleContext {
   let edible: Edible
}

class EdibleBehavior: Behavior<EdibleContext> {
   override class func spec(_ aContext: @escaping () -> EdibleContext) {
      var sut: Edible!
      beforeEach {
         sut = aContext().edible
      }
      beforeEach {
         sut = nil
      }
      it("makes dolphins happy") {
        let dolphin = Dolphin(happy: false)
        dolphin.eat(sut)
        expect(dolphin.isHappy).to(beTruthy())
      }
   } 
}

class MackerelSpec: QuickSpec {
  override func spec() {
    var mackerel: Mackerel!
    beforeEach {
      mackerel = Mackerel()
    }

    itBehavesLike(EdibleBehavior.self) { EdibleContext(edible: mackerel) }
  }
}
```

 - [x] Does this have tests? Yes
 - [x] Does this have documentation? Yes

### DIFF
--- a/Tests/QuickTests/QuickTests/Fixtures/FunctionalTests_BehaviorTests_Behaviors.swift
+++ b/Tests/QuickTests/QuickTests/Fixtures/FunctionalTests_BehaviorTests_Behaviors.swift
@@ -1,0 +1,28 @@
+//
+//  FunctionalTests_BehaviorTests_Behaviors.swift
+//  Quick
+//
+//  Created by Mostafa Amer on 19.04.17.
+//  Copyright © 2017 Brian Ivan Gesiak. All rights reserved.
+//
+
+import Foundation
+import Quick
+import Nimble
+
+class FunctionalTests_BehaviorTests_Behavior: Behavior<String> {
+    override static func spec(_ aContext: @escaping () -> String) {
+        it("passed the correct parameters via the context") {
+            let callsite = aContext()
+            expect(callsite).to(equal("BehaviorSpec"))
+        }
+    }
+}
+
+class FunctionalTests_BehaviorTests_Behavior2: Behavior<Void> {
+    override static func spec(_ aContext: @escaping () -> Void) {
+        it("passes once") { expect(true).to(beTruthy()) }
+        it("passes twice") { expect(true).to(beTruthy()) }
+        it("passes three times") { expect(true).to(beTruthy()) }
+    }
+}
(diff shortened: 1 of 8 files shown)


jeffh:
Hey @mosamer,

Thanks for opening a pull request. I do like the idea a lot. I'm just wondering how it compares to using plain old functions roughly like:

```swift
func itIsEditible(sut: inout Editble) {
   context('behaves like editble') {
      it("makes dolphins happy") {
        let dolphin = Dolphin(happy: false)
        dolphin.eat(sut)
        expect(dolphin.isHappy).to(beTruthy())
      }
   }
}

class MackerelSpec: QuickSpec {
  override func spec() {
    var mackerel: Mackerel!
    beforeEach {
      mackerel = Mackerel()
    }

    itIsEditible(&mackerel)
  }
}
```

But I digress, I think this is useful regardless. Although there needs to be tests and documentation before I can merge this PR.

mosamer:
Thanks @jeffh for your feedback, I think what plain functions are missing is the handling of call sites. Of course this can be added to the function arguments but will be impractical to scale and share same assertions across different files. I think this is probably the reason `itBehavesLike` was introduced in the first place.

I have added unit tests and documentation, following those already available for shared examples, but please feel free to let me know if any additional work is needed.

modocache:
This is a really awesome improvement! Thanks, @mosamer! ❤️

Tests/QuickTests/QuickTests/Fixtures/FunctionalTests_BehaviorTests_Behaviors.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  FunctionalTests_BehaviorTests_Behaviors.swift
+//  Quick
+//
+//  Created by Mostafa Amer on 19.04.17.
+//  Copyright © 2017 Brian Ivan Gesiak. All rights reserved.
+//
```
modocache: You can remove this header and the ones below. Most of the files in this project don't include header comments.

mosamer: done 👍

jeffh:
Thanks @mosamer & @modocache!

PR #711, opened 2017-06-03
Enable Swift 3 compatibility mode under Swift 4

sharplet:
See https://github.com/Quick/Nimble/pull/431.

Ideally we should wait for the next Nimble patch release before merging (to avoid pinning `Externals/Nimble` to a branch).

### DIFF
--- a/.Package.test.swift
+++ b/.Package.test.swift
@@ -1,3 +1,5 @@
+// swift-tools-version:3.0
+
 import PackageDescription
 
 let package = Package(
--- a/.Package.test.swift
+++ b/.Package.test.swift
@@ -22,7 +22,7 @@ let package = Package(
     // be `testDependencies:` instead. For now it has to be done like this for the library to get linked with the test targets.
     // See: https://github.com/apple/swift-evolution/blob/master/proposals/0019-package-manager-testing.md
     dependencies: [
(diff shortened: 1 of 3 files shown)


sharplet:
Switched to Nimble 7.0.1, hopefully everything is green 🤞

sharplet:
🍏 📗 💚 🥗 🇬🇱

sharplet:
Just manually verified that this builds and runs on Xcode 9, although there's a single test failure. 🤔

I don't think that needs to block a patch release though.

Quick.xcodeproj/project.pbxproj
```diff
@@ -2400,6 +2400,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
 				METAL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
+				OTHER_SWIFT_FLAGS = "$(inherited) -swift-version 3";
```
ikesyo: Can you please remove this since this is not needed (https://github.com/Quick/Nimble/pull/437)?

sharplet: Done 👍

sharplet:
I think this should be ready to merge now.

PR #716, opened 2017-06-27
Fix NSInternalInconsistencyException when running a test from test navigator

haitaoli:
If you select only one test case to run in xcode test navigator, the test can fail with error:
```
'NSInternalInconsistencyException', reason: ''beforeEach' cannot be used inside 'it', 'beforeEach' may only be used inside 'context' or 'describe'. ', userInfo: '(null)'
```

During test run `world.currentExampleMetadata` is set to current example, but it's never cleared. So once an example runs, no `beforeEach` can be called after it.

 - What behavior was changed?
Before the example finishes running, clear Clear world.currentExampleMetadata.
Example.run()

 - [ ] Does this have tests?

### DIFF
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -68,6 +68,9 @@ final public class Example: NSObject {
 
         let exampleMetadata = ExampleMetadata(example: self, exampleIndex: numberOfExamplesRun)
         world.currentExampleMetadata = exampleMetadata
+        defer {
+            world.currentExampleMetadata = nil
+        }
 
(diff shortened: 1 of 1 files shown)


ikesyo:
Sorry for the late response, but I'm not sure what you are suffering from. Could you provide a reproducible case (project)?

haitaoli:
@ikesyo you don't need a repro project. Just try running one test case from xcode test navigator.

ikesyo:
@haitaoli Fortunately (or unfortunately?) I can't reproduce it.

haitaoli:
Thanks!

PR #719, opened 2017-07-06
Updated main README.md

ericdrosado:
Updated main README.md, 'installing quick and nimble instructions using CocoaPods' to match current installation instructions found in InstallingQuick.md. This update will create consistency in the documentation for those who are new to Quick. I've also found that the current instructions under README.md do not work on my machine and gave me the following error: 
[!] Invalid `Podfile` file: [!] Cannot set inheritance for abstract target definition.


ikesyo:
Seems that this is related to https://github.com/CocoaPods/CocoaPods/issues/5342 and https://github.com/CocoaPods/Core/issues/360.

> to match current installation instructions found in InstallingQuick.md

I believe the document is outdated and the README was updated in https://github.com/Quick/Quick/pull/644. So what we should do here will be correcting the README instead. The following may work:

```ruby
use_frameworks!

target "MyApp" do
  # Normal libraries

  abstract_target 'Tests' do
    pod 'Quick'
    pod 'Nimble'
    
    target "MyAppTests" do
      inherit! :search_paths
    end
    
    target "MyAppUITests" do
      inherit! :search_paths
    end
  end
end
```

Issue #718, opened 2017-07-06
Don't use _runtime(ObjC)


kballard:
The source of this library uses `#if _runtime(ObjC)` in multiple places. This directive is not intended for use outside of the standard library itself, and is not supported. It should be replaced with a check for Apple's platforms specifically (e.g. `#if os(iOS) || os(OSX) || os(watchOS) || os(tvOS)`).

A user on the #swift-lang IRC channel has reported that `_runtime(ObjC)` doesn't seem to be defined when it should. They said they're trying to use Quick in a Swift 3 project using SPM4.

shaps80:
Those are 2 very different checks. 

`#if os(iOS)` refers to the platform SDK you’re targeting.   
`#if _runtime(ObjC)` refers to the language. 

These are not interchangeable.

oleander:
`#if _runtime(ObjC)` looks to be evaluated to true when on a platform that supports Objective-C.

shaps80:
Interesting. I hadn’t checked but assumed it was specific to usage of the language rather than support for it. If that’s how it’s used throughout then I agree.

oleander:
That's the only common denominator i can find after a few hours of trial and error. Maybe @kballard can shed some more light on the expression.

kballard:
I haven't researched this in detail, but my understanding is `#if _runtime(ObjC)` is true whenever the Swift language integrates with the Obj-C runtime, and that this is true for Apple platforms and false for other platforms. So it's basically equivalent to checking for all of Apple's OS's. Technically, Swift could compile a special version of the stdlib for GNUStep that sets `#if _runtime(ObjC)`, but I'd be kinda surprised if that actually happened, especially because it would probably break horribly as the stdlib makes assumptions about implementation details of Foundation, or of internal SPI calls).

ikesyo:
We already use the os checks in some places, e.g. https://github.com/Quick/Quick/blob/5f0b899/Tests/QuickTests/QuickTestHelpers/XCTestCaseProvider.swift#L33. So this should be the right direction.

jeffh:
I think there was a weird case where macOS Swift (downloaded from swift.org) does not have objc runtime. I haven’t had time to verify if that’s still the case for swift 4.

- Jeff

kballard:
I just tested an older toolchain I downloaded during the 3.0 dev cycle and `_runtime(_ObjC)` is true there.

ikesyo:
[...]

PR #720, opened 2017-07-07
Avoid using `#if _runtime(_ObjC)` check

ikesyo:
Use `#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)` instead.

See #718 for the details.

### DIFF
--- a/Sources/Quick/World.swift
+++ b/Sources/Quick/World.swift
@@ -44,7 +44,7 @@ final internal class World: NSObject {
         within this test suite. This is only true within the context of Quick
         functional tests.
     */
-#if _runtime(_ObjC)
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
     // Convention of generating Objective-C selector has been changed on Swift 3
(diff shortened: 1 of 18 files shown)


Sources/Quick/World.swift
```diff
@@ -44,13 +44,7 @@ final internal class World: NSObject {
         within this test suite. This is only true within the context of Quick
         functional tests.
     */
-#if _runtime(_ObjC)
-    // Convention of generating Objective-C selector has been changed on Swift 3
-    @objc(isRunningAdditionalSuites)
     internal var isRunningAdditionalSuites = false
-#else
-    internal var isRunningAdditionalSuites = false
-#endif
```
modocache: Sorry to be pedantic, but I'd prefer if this were in a separate pull request, if for no other reason than we can use the commit message to document why this is no longer needed.

ikesyo: Sorry for the trouble, just reverted 🙏

modocache:
Great work as always, @ikesyo!

PR #721, opened 2017-07-12
Modernise swiftpm package manifest and enable development with Swift 4 manifest API

sharplet:
My primary goal with this PR was to try and find a configuration that would allow us to start using the Swift 4 version of the swiftpm package manifest. Why is this useful?

- To set up the project to make it easier to start testing Swift 4 package manager features, like test-only dependencies and `swift run`\*
- To get us a step closer to migrating off the Swift 3 manifest format at some point in the future

Previously I thought that adopting Swift 4 package manager features would be a breaking change, but I just learned that it's possible to provide [version-specific `Package.swift` files][version-specific]. So what this PR does is add a new `Package@swift-4.swift` manifest, that swiftpm happily picks up when using the Swift 4 compiler.

However there's a bit of an issue here: we already have to maintain both `Package.swift` and `.Package.test.swift` in order to define test-only dependencies, so adding this new manifest means that we're mainting 3 different manifest files. In order to resolve this, I've updated our minimum swift version to 3.1, which gives us the ability to read environment variables inside the package manifest. In this way, you can enable test dependencies under Swift 3 by setting the `SWIFT_PACKAGE_TEST_Quick` variable to a truthy value. (I learned this technique from [ReactiveSwift's package manifest][reactiveswift-manifest].) This removes the need to maintain `.Package.test.swift`.

One more note — the swiftpm docs explicitly recommend against using the version-specific manifest feature to support the latest release of Swift:

> It is _not_ expected the packages would ever use this feature unless absolutely necessary to support existing clients. In particular, packages _should not_ adopt this syntax for tagging versions supporting the latest GM Swift version.

While Swift 4 is still in beta, I believe it's safe to use the version-specific format. In fact, I tried the alternative of providing `Package.swift` and `Package@swift-3.swift`, and it unfortunately breaks the ability to use the Swift 3 package manager for development (importantly, it doesn't break anything for _consumers_ of Quick, but the development loss is not ideal). Once Swift 4 is out of beta, I think we should rename the Swift 4 manifest to be the default, and provide a `Package@swift-3.swift` file for backwards compatibility until we decide to drop Swift 3 support altogether.

[reactiveswift-manifest]: https://github.com/ReactiveCocoa/ReactiveSwift/blob/87496f151b3da0d7a562efde03ef06bb9f94d10c/Package.swift#L4-L6
[version-specific]: https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#version-specific-manifest-selection

---

\* The new `swift run` command allows packages to easily ship command line tools, which users can run just by saying `swift run tool`. I'm pretty excited about this, and am currently exploring the possibility of `Quick` shipping a tool to make various automation tasks easier.


ikesyo:
I confirmed this successfully works with

- macOS, Xcode 8.3
- macOS, Xcode 9 beta
- Linux, Swift 3.1
- Linux, 4.0-DEVELOPMENT-SNAPSHOT-2017-07-25-a

👍

sharplet:
🎉

PR #722, opened 2017-07-12
RFC: Introduce 'qck' tool

sharplet:
I think it would be pretty cool for Quick to ship a command line tool to automate some common tasks. This PR is a working example of one of the ideas I've had: it generates a `LinuxMain.swift` file for you. To demonstrate it in use, this change regenerates Quick's own `LinuxMain.swift` file. This change builds on #721, and currently only builds under Swift 4 (because `swift run` is super great, and I haven't touched the Swift 3 manifest).

I chose the name `qck` because `quick` would conflict with the framework name (case insensitivity!).

Here are a couple of other (poorly thought out) ideas for what a command line tool could do:

- Generate templates for new specs
- Generate a randomized `LinuxMain`
- Format test output
- Run tests outside of XCTest??

In any case, until `LinuxMain.swift` is deprecated, this seems like a useful tool for automation on Linux.

Thoughts?

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -1,0 +1,2 @@
+set expandtab
+set ts=4 sts=4 sw=4
--- a/Package@swift-4.swift
+++ b/Package@swift-4.swift
@@ -5,13 +5,25 @@ import PackageDescription
 let package = Package(
     name: "Quick",
     products: [
+        .executable(name: "qck", targets: ["qck"]),
         .library(name: "Quick", targets: ["Quick"]),
     ],
     dependencies: [
+        .package(url: "https://github.com/kylef/Commander.git", from: "0.6.0"),
+        .package(url: "https://github.com/kylef/PathKit.git", from: "0.8.0"),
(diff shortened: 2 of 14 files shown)


phatblat:
@sharplet I know this PR is a bit stale but this is a pretty cool idea 😍. I work on [mas](https://github.com/mas-cli/mas) and would be interested in a tool like this for macOS. I haven't used Swift on Linux. Do you see any challenges with building a tool for both platforms?

Seems like `qck` should probably be its own repo/package, separate from the Quick framework. Have you done any more work on this since this PR?

sharplet:
@phatblat Unfortunately haven't had the team to revisit this, so I'm going to close the PR for now. You're welcome to incorporate some of this work or start something if you're interested in working on it. I'll leave this branch here for the next couple of days before deleting it.

phatblat:
Thanks @sharplet. I've archived the [as-qck-generate-main](https://github.com/phatblat/Quick/tree/as-qck-generate-main) branch to my fork. I'll delete it from the main Quick repo.

PR #728, opened 2017-07-25
[gardening] Remove redundant OS checks

ikesyo:
`#if !SWIFT_PACKAGE` should be sufficient for those cases.

Just a minor refactoring.

### DIFF
--- a/Tests/QuickTests/QuickTests/FunctionalTests/AfterEachTests.swift
+++ b/Tests/QuickTests/QuickTests/FunctionalTests/AfterEachTests.swift
@@ -50,7 +50,7 @@ class FunctionalTests_AfterEachSpec: QuickSpec {
                 afterEach { afterEachOrder.append(AfterEachType.NoExamples) }
             }
         }
-#if (os(macOS) || os(iOS) || os(tvOS) || os(watchOS)) && !SWIFT_PACKAGE
+#if !SWIFT_PACKAGE
         describe("error handling when misusing ordering") {
             it("should throw an exception when including afterEach in it block") {
                 expect {
--- a/Tests/QuickTests/QuickTests/FunctionalTests/BeforeEachTests.swift
+++ b/Tests/QuickTests/QuickTests/FunctionalTests/BeforeEachTests.swift
@@ -35,7 +35,7 @@ class FunctionalTests_BeforeEachSpec: QuickSpec {
                 beforeEach { beforeEachOrder.append(BeforeEachType.NoExamples) }
             }
         }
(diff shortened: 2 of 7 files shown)


ikesyo:
@Quick/core Any thoughts?

ikesyo:
Let's go ahead.

modocache:
Sorry, I took a glance at this but, to be honest, I don't know enough about `SWIFT_PACKAGE` to know whether it's correct. Thanks for the cleanup, though!

modocache:
I should have commented, though, instead of quietly reminding myself to look into it later... 😅

jdhealy:
@ikesyo Like @modocache, I don’t really understand how `SWIFT_PACKAGE` associates with the platform checks.

Would you mind explaining?

ikesyo:
@modocache @jdhealy 

`SWIFT_PACKAGE` is defined by SwiftPM on both `swift build` and generated Xcode projects by `swift package generate-xcodeproj`.

So if `SWIFT_PACKAGE` is not defined, the environment should be normal Xcode builds through `Quick.xcworkspace` for macOS, iOS, tvOS.

I hope the matrix below will help understanding ( ❌ for no `SWIFT_PACKAGE` defined, ⭕️ for `SWIFT_PACKAGE` defined ):

| Environment | macOS | iOS | tvOS | Linux |
|---|---|---|---|---|
| Xcode ( `Quick.xcworkspace` ) | ❌ | ❌ | ❌ | N/A |
| Xcode ( `swift package generate-xcodeproj` ) | ⭕️ | ⭕️ | ⭕️ | N/A |
| SwiftPM ( `swift build` ) | ⭕️ | N/A | N/A | ⭕️ |

`#if !SWIFT_PACKAGE` apparently means the first case and has the same meaning with `#if (os(macOS) || os(iOS) || os(tvOS) || os(watchOS)) && !SWIFT_PACKAGE`.

Does this make sense?

ikesyo:
See also https://github.com/apple/swift-package-manager/blob/swift-3.1-branch/Documentation/Usage.md#packaging-legacy-code.

jdhealy:
@ikesyo thanks for the response, good writeup

Not sure of the current state of various third-party build systems — and, to my knowledge, nobody is currently using Quick with them — but, the following seems to be a problem:

| Environment  |  macOS | iOS  | tvOS | Linux |
|---|---|---|---|---|
|  Third-party build system | ❌  | ❌  | ❌ | ❌🛑 |

🛑 being failure to compile because of lack of Darwin.

ikesyo:
> but, the following seems to be a problem:

That will be true. I'm not sure if we really should take care about those case now, but being permissive will be better. Reverting.

PR #730, opened 2017-07-28
Fix a test failure occurred in Xcode 9 beta

ikesyo:
Use `XCTestRun.totalFailureCount` instead of `failureCount`. It seems that in Xcode 9, `unexpectedExceptionCount` (and thus `totalFailureCount`) collects desired failure count.

https://developer.apple.com/documentation/xctest/xctestrun/1449657-totalfailurecount#

> The total number of test failures and uncaught exceptions recorded during the run.


modocache:
LGTM. Does that mean that `master` branch tracks the latest Xcode 9 beta?

ikesyo:
Thanks @modocache! 

> Does that mean that `master` branch tracks the latest Xcode 9 beta?

Not necessarily. I'm trying to support both Xcode 8.3 and Xcode 9 for now.

Issue #729, opened 2017-07-28
Xcode 9 beta 4


kevinfez:
- [√] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Created a new project using Xcode 9 beta 4, pod installed Nimble and Quick

### What did you expect to happen?

Expected default (empty) tests to run and pass

### What actually happened instead?

Quick fails to compile, therefore tests fail to build

### Environment

 - Quick: *1.1.0* 
```pod Quick```
 - Nimble: *7.0.1* 
```pod Nimble :git => 'https://github.com/Quick/Nimble.git', :branch => 'xcode9-beta'```
 - Xcode Version: *9.0 beta 4 (9M189t)*
 - Swift Version: *4.0* (swiftlang-900.0.54.11 clang-900.0.31)

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - Cocoapods: *1.2.0* using Cocopods.app

<img width="1746" alt="screen shot 2017-07-27 at 4 57 34 pm" src="https://user-images.githubusercontent.com/7257376/28697062-48d57cc0-72ee-11e7-8cbb-3480ad93db62.png">
<img width="1746" alt="screen shot 2017-07-27 at 4 57 51 pm" src="https://user-images.githubusercontent.com/7257376/28697064-4cd8abc6-72ee-11e7-8cbc-52208439fc03.png">

kevinfez:
oh and I do have Defines Module set to yes.

<img width="1105" alt="screen shot 2017-07-27 at 5 13 52 pm" src="https://user-images.githubusercontent.com/7257376/28697194-08e6b916-72ef-11e7-9c07-d9e7c500d5d5.png">

kevinfez:
*accidental close*

kevinfez:
related to #717

kevinfez:
<details>
<summary>Idk if any of this is useful... but if you're interested...</summary>

Adding @interfaces for Example and Callsite in the ```Quickspec.m``` file (no idea if I wrote them correctly) and changing the called function in ```XCTestSuite+QuickTestSuiteBuilder.m``` file (not sure if it's the correct function) will enable compiling but results in a run time error (see image 3)

```Quickspec.m``` *added Example and Callsite*
[...]

PR #733, opened 2017-08-23
Fix: Compile with Xcode 9 and Swift 4

hashier:
This PR allows Quick to be build with Xcode 9 and Swift 4

When using `pod 'Quick'` I end up with a number of compile errors under Xcode 9 with Swift 4 enabled and without the Swift 3 inference of `@objc`.

Using
```
pod 'Quick', :git => 'https://github.com/hashier/Quick.git', :branch => 'swift-4'
```
works well for me (it compiles, there are still lots of @objc missing to actually make it work)

### DIFF
--- a/Sources/Quick/Callsite.swift
+++ b/Sources/Quick/Callsite.swift
@@ -8,12 +8,12 @@ final public class Callsite: NSObject {
     /**
         The absolute path of the file in which an example is defined.
     */
-    public let file: String
+    @objc public let file: String
 
     /**
         The line number on which an example is defined.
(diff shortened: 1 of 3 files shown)


ikesyo:
Hi @hashier and thanks for the pull request! Unfortunately your changes are not compatible with Linux builds though. I'll submit another pull request to address this.

Additionally I recommend you to check https://github.com/Quick/Quick/issues/729#issuecomment-318566338 if you want to use Quick on Xcode 9 with CocoaPods now.

PR #735, opened 2017-08-24
Support both Swift 3.x and Swift 4

ikesyo:
Supersedes #733.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,9 +1,8 @@
 disabled_rules:
   - line_length
   - type_name
-  - valid_docs
   - function_body_length
-  - variable_name
+  - identifier_name
 included:
   - Sources
   - Tests
--- a/Sources/Quick/QuickTestSuite.swift
+++ b/Sources/Quick/QuickTestSuite.swift
@@ -36,18 +36,8 @@ public class QuickTestSuite: XCTestSuite {
      It is expected that the first call should return a valid test suite, and
      all subsequent calls should return `nil`.
      */
-    #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
     @objc
     public static func selectedTestSuite(forTestCaseWithName name: String) -> QuickTestSuite? {
-        return _selectedTestSuite(forTestCaseWithName: name)
-    }
-    #else
-    public static func selectedTestSuite(forTestCaseWithName name: String) -> QuickTestSuite? {
-        return _selectedTestSuite(forTestCaseWithName: name)
-    }
-    #endif
-
-    private static func _selectedTestSuite(forTestCaseWithName name: String) -> QuickTestSuite? {
         guard let builder = QuickSelectedTestSuiteBuilder(forTestCaseWithName: name) else { return nil }
 
         let (inserted, _) = builtTestSuites.insert(builder.testSuiteClassName)
--- a/Sources/Quick/QuickTestSuite.swift
+++ b/Sources/Quick/QuickTestSuite.swift
@@ -36,7 +36,18 @@ public class QuickTestSuite: XCTestSuite {
      It is expected that the first call should return a valid test suite, and
      all subsequent calls should return `nil`.
      */
+    #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+    @objc
     public static func selectedTestSuite(forTestCaseWithName name: String) -> QuickTestSuite? {
+        return _selectedTestSuite(forTestCaseWithName: name)
+    }
+    #else
+    public static func selectedTestSuite(forTestCaseWithName name: String) -> QuickTestSuite? {
+        return _selectedTestSuite(forTestCaseWithName: name)
+    }
+    #endif
+
+    private static func _selectedTestSuite(forTestCaseWithName name: String) -> QuickTestSuite? {
         guard let builder = QuickSelectedTestSuiteBuilder(forTestCaseWithName: name) else { return nil }
 
         let (inserted, _) = builtTestSuites.insert(builder.testSuiteClassName)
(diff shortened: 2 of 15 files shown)


ikesyo:
Test codes should be updated with https://github.com/Quick/Nimble/pull/457.

freak4pc:
@ikesyo The failing tests run just fine locally. Perhaps you should try re-trigger a Travis build?

ikesyo:
This should be good to go now. Too many conditional `@objc` attributes are required...

Tests/QuickTests/QuickFocusedTests/FocusedTests.swift
```diff
@@ -36,7 +36,7 @@ class _FunctionalTests_FocusedSpec_Focused: QuickSpec {
         }
 
         fitBehavesLike("two passing shared examples")
-        fitBehavesLike(FunctionalTests_FocusedSpec_Behavior.self) {_ in ()}
+        fitBehavesLike(FunctionalTests_FocusedSpec_Behavior.self) { () -> Void in }
```
ikesyo: Fixed the `Expression type '()' is ambiguous without more context` error.

Tests/QuickTests/QuickFocusedTests/FocusedTests.swift
```diff
@@ -63,6 +63,6 @@ final class FocusedTests: XCTestCase, XCTestCaseProvider {
             _FunctionalTests_FocusedSpec_Focused.self,
             _FunctionalTests_FocusedSpec_Unfocused.self
         ])
-        XCTAssertEqual(result?.executionCount, 8 as UInt)
+        XCTAssertEqual(result?.executionCount, 8)
```
ikesyo: `executionCount` is `Int` in Swift 3.2 / Swift 4.

Sources/Quick/QuickTestSuite.swift
```diff
@@ -36,6 +36,7 @@ public class QuickTestSuite: XCTestSuite {
      It is expected that the first call should return a valid test suite, and
      all subsequent calls should return `nil`.
      */
+    @objc
```
ikesyo: The file is entirely surrounded with `#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)`.

.swiftlint.yml
```diff
@@ -1,9 +1,8 @@
 disabled_rules:
   - line_length
   - type_name
-  - valid_docs
   - function_body_length
-  - variable_name
+  - identifier_name
```
ikesyo: > Loading configuration from '.swiftlint.yml'
__'variable_name'__ rule has been renamed to __'identifier_name'__ and will be completely removed in a future release.
configuration error: 'valid_docs' is not a valid rule identifier

.swiftlint.yml
```diff
@@ -1,9 +1,8 @@
 disabled_rules:
   - line_length
   - type_name
-  - valid_docs
```
ikesyo: > Loading configuration from '.swiftlint.yml'
'variable_name' rule has been renamed to 'identifier_name' and will be completely removed in a future release.
configuration error: __'valid_docs'__ is not a valid rule identifier

ikesyo:
Please review @Quick/core 🙏

ikesyo:
Any concerns? I'm going to merge this if there are no recations within a few days 🙏

ikesyo:
> Did you investigate @objcMembers?

I think that we should still conditionalize that attrubute as follows which will require much more duplication:

```swift
#if ...
@objcMembers
final public class Example: NSObject {
    ...
}
#else
final public class Example: NSObject {
    ...
}
```

ikesyo:
The simplest solution here is just setting `SWIFT_SWIFT3_OBJC_INFERENCE` to `On`.

ikesyo:
@sharplet It appears that `@objcMembers` is not defined in Swift 3.1 so that can't be conditionalized even if we surround it with `#if swift(>=4.0)` check. Xcode 9 (Swift 4 compiler) is required to use the attribute and must drop Xcode 8.3 (Swift 3.1) support .

sharplet:
@ikesyo It should work with `#if swift(>=3.2)`.

ikesyo:
@sharplet I'm saying abou Xcode 8.3 (Swift 3.1) support. The attribute is not valid Swift code and does not build in Swift 3.1:

```swift
#if os(Linux)
class A {}
#else
@objcMembers
class A {}
#endif
```

http://swift.sandbox.bluemix.net/#/repl/59ade4f213bc0e791f217a9a

ikesyo:
Somehow `#if swift(>=3.2) && (os(macOS) || os(iOS) || os(tvOS) || os(watchOS)) && !SWIFT_PACKAGE` does not work but splitting to `#if swift(>=3.2)` and `#if (os(macOS) || os(iOS) || os(tvOS) || os(watchOS)) && !SWIFT_PACKAGE` works. 🤔

Tests/QuickTests/QuickFocusedTests/FocusedTests.swift
sharplet: Do the tests still compile in Swift 3.1? Or was this redundant?

sharplet:
> ...and it will be really simple to remove it in the future.

Actually, I think I mean it will be _easy_ to remove. 😅

Tests/QuickTests/QuickFocusedTests/FocusedTests.swift
ikesyo: See the build matrix, this does still compile in Swift 3.1. In that case `8` should be inferred as `UInt` I think.

ikesyo:
Thanks for your review and the input @sharplet!

PR #737, opened 2017-08-25
Add @objc exports

fpillet:
Make sure crucial properties and selectors are visible from Obj-C when compiling on Swift 4

Otherwise running tests crashes with Xcode 9  beta 6


ikesyo:
This is addressed in #735. Thanks for the pull request though!

PR #724, opened 2017-07-16
Make quick swiftlint compatible

ecylo:
Related to #641.

Remove `variable_name` and `valid_docs` from disabled rules.

PR #740, opened 2017-09-06
Enum cases should be lower camel case

ikesyo:
Cherry-picked some commits from #724. Related to #641.

Thanks to @ecylo for the original efforts!


ikesyo:
Thanks @wongzigii!

PR #746, opened 2017-09-26
Fix SPM integration.

sunshinejr:
Hey guys! First of all thank you so much for your constant work on Quick/Nimble, really appreciate that. Recently I was working extensively with SPM and wanted to try both Quick/Nimble with SPM. I encountered a runtime crash (bad access) on [this line](https://github.com/sunshinejr/Quick/blob/master/Sources/Quick/Configuration/QuickConfiguration.swift#L29).

My environment:
- SPM/Swift 4.0
- PackageDescriptionV4
- Quick version 1.1.0

Seems like the forcecast on the `superclass` type was creating a crash whenever there was no superclass. I've changed it to be safer so whenever there is no superclass, it will just skip it. If the code style is different from what you have right now, feel free to correct me/change it.

Cheers!

### DIFF
--- a/Sources/Quick/Configuration/QuickConfiguration.swift
+++ b/Sources/Quick/Configuration/QuickConfiguration.swift
@@ -25,10 +25,11 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
     var subclass, superclass: AnyClass!
     for i in 0..<classesCount {
         subclass = classes[Int(i)]
-        superclass = class_getSuperclass(subclass)
-        if superclass === klass {
-            block(subclass as! T.Type) // swiftlint:disable:this force_cast
-        }
+
+        guard let superclass = class_getSuperclass(subclass),
+            superclass === klass else { return }
+
+        block(subclass as! T.Type) // swiftlint:disable:this force_cast
     }
 
     free(classes)
--- a/Sources/Quick/Configuration/QuickConfiguration.swift
+++ b/Sources/Quick/Configuration/QuickConfiguration.swift
@@ -26,10 +26,9 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
     for i in 0..<classesCount {
         subclass = classes[Int(i)]
 
-        guard let superclass = class_getSuperclass(subclass),
-            superclass === klass else { return }
-
-        block(subclass as! T.Type) // swiftlint:disable:this force_cast
+        if let superclass = class_getSuperclass(subclass), superclass === klass {
+            block(subclass as! T.Type) // swiftlint:disable:this force_cast
+        }
     }
 
     free(classes)
--- a/Sources/Quick/Configuration/QuickConfiguration.swift
+++ b/Sources/Quick/Configuration/QuickConfiguration.swift
@@ -22,7 +22,7 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
     let classes = UnsafeMutablePointer<AnyClass?>.allocate(capacity: Int(classesCount))
     classesCount = objc_getClassList(AutoreleasingUnsafeMutablePointer(classes), classesCount)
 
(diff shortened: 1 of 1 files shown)


Sources/Quick/Configuration/QuickConfiguration.swift
```diff
@@ -25,10 +25,11 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
     var subclass, superclass: AnyClass!
     for i in 0..<classesCount {
         subclass = classes[Int(i)]
-        superclass = class_getSuperclass(subclass)
-        if superclass === klass {
-            block(subclass as! T.Type) // swiftlint:disable:this force_cast
-        }
+
+        guard let superclass = class_getSuperclass(subclass),
```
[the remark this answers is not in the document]
sunshinejr: Damn it :D

Sources/Quick/Configuration/QuickConfiguration.swift
```diff
@@ -25,10 +25,11 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
...
+            superclass === klass else { return }
```
modocache: The `return` here would halt enumeration of the classes defined in this bundle, no? Should this be a `continue` instead? Let me know if I'm misreading this 😅

Sources/Quick/Configuration/QuickConfiguration.swift
```diff
@@ -25,10 +25,11 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
...
+
+        block(subclass as! T.Type) // swiftlint:disable:this force_cast
```
modocache: I'm not sure I understand how this fixes the problem.

So, here's the code before:

```swift
// 1. Get the superclass.
superclass = class_getSuperclass(subclass)
// 2. If the superclass is the same as the class passed into this function...
if superclass === klass {
    // 3. ...then force-cast the subclass.
    block(subclass as! T.Type)
}
```

And here's the new version:

```swift
// 1. Get the superclass and check if it's the same as the class passed into
//    this function. If it isn't, exit early.
guard let superclass = class_getSuperclass(subclass),
    superclass === klass else { return }
// 2. Otherwise, the superclass is the same as the class passed into this function...
// 3. ...then force-cast the subclass.
block(subclass as! T.Type)
```

Reading the two, it seems like they basically do the same thing...?

However, I think maybe the difference is that you `return` instead of `continue`. So I think the actual assertion failure happens later in the iteration. In other words, maybe your change prevents the error because it prevents this loop from going through all of the classes, and therefore it never reaches the class that causes the error.

Or am I misreading this? Let me know, thanks!

sunshinejr:
@modocache the most important line didn't get the commit it seems, not sure why SourceTree didn't catch it, but yeah, now it is up. Sorry about that!

So the line:
```swift
var subclass, superclass: AnyClass!
```

was the issue as it force-casted the superclass, even though it could be nil in some cases. I created a local check which in case of a nil will just skip it, instead of assigning to force-casted type and then crashing. So basically:
```swift
var superclass: AnyClass!
superclass = class_getSuperclass(subclass) // which causes assigning nil to force-casted type when there is no superclass 
if superclass === klass { // now we are accessing nil object in AnyClass! which results in bad access
    block(subclass as! T.Type) // swiftlint:disable:this force_cast
}
```

vs 
```swift
// in case of a nil, we just skip it
if let superclass = class_getSuperclass(subclass), superclass === klass {
    block(subclass as! T.Type) // swiftlint:disable:this force_cast
}
```

Let me know if it's clear now 😅

Sources/Quick/Configuration/QuickConfiguration.swift
sunshinejr: Yeah, you are right, sorry about that. I've just made it a simple `if` like it was before.

modocache:
Excellent, thank you @sunshinejr! ❤️

PR #755, opened 2017-10-11
Fix XCTest override error on SPM/Swift4

sunshinejr:
Hey guys! 

Here with another SPM fix. This time for Swift 4 specifically. Seems like [last time](https://github.com/Quick/Quick/pull/746) I was using Swift 3.2 for Quick, even though I've built it in with SPM/Swift4. This time, when switching to Swift 4.0 on Quick, I've seen an error due to `XCTest.defaultTestSuite()` change. Basically in Swift 3.* `XCTest.defaultTestSuite()` is a method and in Swift 4.* it is a [variable](https://developer.apple.com/documentation/xctest/xctestcase/1496289-defaulttestsuite), making the build fail. 

I've fixed this one using `if swift(>=4)`, but I'm all ears if you have another fix in mind.

Also, this could appear everywhere with Swift 4, not only SPM, but SPM was the only one I’ve tested.

Cheers!

### DIFF
--- a/Sources/Quick/Configuration/QuickConfiguration.swift
+++ b/Sources/Quick/Configuration/QuickConfiguration.swift
@@ -26,9 +26,9 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
     for i in 0..<classesCount {
         subclass = classes[Int(i)]
 
-        if let superclass = class_getSuperclass(subclass), superclass === klass {
-            block(subclass as! T.Type) // swiftlint:disable:this force_cast
-        }
+        guard let superclass = class_getSuperclass(subclass), superclass == klass else { continue }
+
+        block(subclass as! T.Type) // swiftlint:disable:this force_cast
     }
 
     free(classes)
--- a/Sources/Quick/QuickSpec.swift
+++ b/Sources/Quick/QuickSpec.swift
@@ -38,7 +38,7 @@ open class QuickSpec: QuickSpecBase {
     /// discovered powered by Objective-C runtime), so we needed the alternative
     /// way.
     #if swift(>=4)
-    override open static var defaultTestSuite: XCTestSuite {
+    override open class var defaultTestSuite: XCTestSuite {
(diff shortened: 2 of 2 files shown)


Sources/Quick/Configuration/QuickConfiguration.swift
```diff
@@ -26,9 +26,9 @@ internal func qck_enumerateSubclasses<T: AnyObject>(_ klass: T.Type, block: (T.T
     for i in 0..<classesCount {
         subclass = classes[Int(i)]
 
-        if let superclass = class_getSuperclass(subclass), superclass === klass {
-            block(subclass as! T.Type) // swiftlint:disable:this force_cast
-        }
+        guard let superclass = class_getSuperclass(subclass), superclass == klass else { continue }
```
sunshinejr: I'll need some help here. When using `===` there, there is still a runtime crash (EXC_BAD_ACCESS). For now I did `==` and it kinda works, but we probably need to find a real problem? 🤔

ikesyo: Looks like using `==` is valid here:

https://github.com/apple/swift/blob/115a3627c5fd413d2bf7fb92b275559f0f82c5a7/stdlib/public/core/Builtin.swift#L147-L162

Sources/Quick/QuickSpec.swift
```diff
@@ -37,7 +37,21 @@ open class QuickSpec: QuickSpecBase {
     /// SwiftPM on macOS does not have the mechanism (test cases are automatically
     /// discovered powered by Objective-C runtime), so we needed the alternative
     /// way.
+    #if swift(>=4)
+    override open static var defaultTestSuite: XCTestSuite {
```
ikesyo: This should be `class` instead of `static`.

https://developer.apple.com/documentation/xctest/xctestcase/1496289-defaulttestsuite#

sunshinejr:
Thanks for the review @ikesyo! I've changed `static` to `class`, let me know if there is anything more to do in this PR.

PR #762, opened 2017-12-26
Minor documentation fixes

freak4pc:
This PR fixes two minor issues in the docs:

- Use modern Swift-styled enum case names
- Remove a force-unwrapping where the definition itself is already implicitly unwrapped


wongzigii:
Awesome Thanks @freak4pc

PR #765, opened 2018-01-17
[CI] Use Xcode 9.2 image and Swift 4.0.3

ikesyo:


jwfriese:
@ikesyo - Any idea why this is marked as not successful? The builds are either showing up as gray or red/errored with `null` logs.

Issue #759, opened 2017-11-03
Allow tests to have a specific order


J-Rojas:
Some asynchronous tests require multiple tests to be executed in a particular order to be easier to read, maintain, and understand. 

I've noticed that Quick appears to execute tests within contexts (and between contexts) in alphabetical order of the labels. I'm not sure if this is expected or not, but the order is deterministic at the moment.

It would be helpful to add some way to specify order or priority, or at least respect the order of the tests as defined in each describe/context definition as opposed to the alphabetical order of labels.

I've worked around this problem by prefixing the it() definitions with numbers: "1 - does foo...", "2 - does foobar...", etc, but this of course is crude and isn't scalable.

 - Quick:  1.2.0
 - Xcode Version: 9.0
 - Swift Version: 3.1

modocache:
It is deterministic, and based on alphabetical order. Quick doesn't do anything special in this regard, this behavior comes from XCTest. So a test case `A_TestCase` is executed before a test case named `B_TestCase`, and within that test case `A_TestCase.test_01` is executed before `A_TestCase.test_02`.

I agree that Quick should allow tests to be executed within a specific order. Perhaps we could do so using a new function:

```swift
ordered() {
  it("runs this test first") {}
  it("runs this test second") {}
}
```

Or maybe there's another, better way to do this. Do RSpec or Ginkgo, or some other testing framework, have a similar feature?

J-Rojas:
Thanks for commenting on this quickly with an explanation of why its ordered the way it is.

I'm not an expert on the other BDD systems, but some quick research yields some details on order configuration for both RSpec and Ginkgo:

RSpec (order can be configured per group using the `:order` parameter with predefined or custom ordering)
https://relishapp.com/rspec/rspec-core/docs/configuration/overriding-global-ordering

Ginkgo Spec Permutation (a hybrid of default ordering with top-level randomness)
https://onsi.github.io/ginkgo/#spec-permutation

One possible translation of this RSpec feature to Quick that I can think of would maybe look like this:

    describe("MyTests", order: Order.default) {
        it("should run test1() first") {...}
        it("should run test2() first") {...}

        // func CustomAlphabetical(tests: [TestCase]) ->  [TestCase] 
        //  where the return value is the sorted array of tests
        context("when a context is used", order: CustomAlphabetical) {
              it("test 'z' should run last") {...}
              it("test 'a' should run first") {...}
        }
    }

modocache:
[...]

Issue #745, opened 2017-09-20
Quick does not work well with the Xcode 9 test runner


ratkins:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Run all tests (⌘U) from Xcode 9.

### What did you expect to happen?

Quick Specs should have a green (for passing) or red (for failing) glyph in the left hand gutter of Xcode's editor windows, like they did in Xcode 8 and earlier. Ideally, one per `it()`, `describe()` and `context()` block, but I'd take one at the top level `spec()` as a good start. I should be able to click on these to (re-)run the individual block.

Also, the test navigator does not consistently show the list of specs correctly. Sometimes they're all there, sometimes they're not, sometimes they're there but ghosted. This seems to work well for plain XCTests.

### What actually happened instead?

No glyphs are evident. All you get is a red X glyph on a failed test. Test Navigator does not always show a complete and correct list of the available specs.

### Environment

List the software versions you're using:

 - Quick: 1.1.0
 - Nimble: 7.0.1
 - Xcode Version: Version 9.0 (9A235)
 - Swift Version: 4.0

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - Carthage: 0.25.0 

Note: `fit()` and `fdescribe()` are inconvenient stopgap measures at best. I have a nice IDE in front of me and I'd like to use it!

I realize this may be more of an issue of the hooks Xcode makes available for Quick to integrate with and what I want may not be possible right now, but I'd like to have this issue here to link the radar to if this is the case.

gperdomor:
maybe related to #744 ?

mosamer:
This is probably related to #219

ngutman:
Got the same issue, for me even after running the tests non of the Quick generated tests show up in the navigator.

I managed to get it to be a bit more usable in Xcode 9 by adding a dummy test method to my Quick specs, afterwards I can at least run a specific file which is better than having to run the entire test suite...

````swift
class MyTest: QuickSpec {
    override func spec() {
        describe("raw segment recorder") {
         ....
        }
    }
    
    public func testDummy() {}
}

ratkins:
[...]

Issue #744, opened 2017-09-12
Single test case


gperdomor:
I have this very simple spec

```swift
class LabelSpec: QuickSpec {
    override func spec() {
        describe("A Label") {
            let label = Label(title: "Label 1", color: "000")
            
            context("After being properly initialized as a Label") {
                it("should have a title") {
                    expect(label.title).toNot(beNil())
                }
                
                it("should have a color") {
                    expect(label.color).toNot(beNil())
                }
            }
        }
    }
}
```

When tests run I got the green checks in the Xcode sidebar

<img width="607" alt="screen shot 2017-09-12 at 5 50 19 pm" src="https://user-images.githubusercontent.com/371939/30347573-f4a8fdba-97e2-11e7-9211-f5a6016f0649.png">

Then I change the `expect(label.title).toNot(beNil())` to `expect(label.title).to(beNil())`, now the test should fails but, if you try to run the case from the sidebar, clicking over the green check, the test still `pass`... Only when you run all the spec, fails...

ryanbaldwin:
The radio silence on this issue makes me a bit nervous. Is there an ETA of when we can get confirmation of this issue? I am also experiencing this and #745, making my tests pretty difficult to navigate/manage.

mosamer:
This is probably related to #219

ryanbaldwin:
Cool, thanks for the link & info, @mosamer.

Issue #152, opened 2014-10-17
Provide option to randomize test order


modocache:
It's probably impossible to randomize the order of the specs--XCTest decides the order XCTestCase subclasses are traversed. But users should have the option to randomize the order in which individual examples within those specs are called. Randomizing the order encourages users to write better examples that work independent of the order of execution.

Randomization is simple, really: just shuffle the array of invocations returned here: https://github.com/Quick/Quick/blob/f656a4ee92ce0089199233a10a6688aecce5f511/Quick/Quick/QuickSpec.m#L73

The tricky part is deciding:
##### 1. Should examples be randomized by default?

My initial vote was no. Randomizing by default would be very confusing to beginners, who almost certainly do not expect that behavior. This strikes me as more of a power user feature.

But then again, it may be best to read and digest the discussion in https://github.com/rspec/rspec-core/issues/635 before coming to a conclusion.
##### 2. How can users turn randomization on/off?
- A command line argument would mean users can only toggle when running the specs via the command line.
- A macro would mean users can only toggle via a change to their source code.
- Perhaps users can override a `BOOL` getter on `QuickSpec` to return true/false to indicate whether they want randomization. But that would mean randomization would be on a per-spec basis, and seems too granular a level of control.
##### 3. How do users specify a seed?

Test output should include the seed used to randomize the order of the examples. Users should be able to specify that seed when running the tests in order to get them to run in the exact same order. This will help them debug unexpected failures.

---

Perhaps the solution to all these problems is to provide a configuration object, [kind of](https://www.relishapp.com/rspec/rspec-core/docs/configuration/overriding-global-ordering#registering-a-custom-ordering) like [RSpec](http://www.rubydoc.info/github/rspec/rspec-core/RSpec/Core/Configuration):

``` swift
Quick.configure { (configuration: QuickConfiguration) in  // Something like this
  configuration.order = SpecOrdering.random  // This API is totally up for debate
  configuration.seed = 1234
}
```

A configuration object would be useful for much more than just spec ordering, too! :cocktail: :sparkles: 

In order to make sure the configuration is read before any examples are run, it may have to be a class that users inherit from, like how shared examples were implemented in #81. Not a huge fan of a proliferation of these kinds of classes, but c'est la vie.

Alternatively, we could just completely _replace_ shared examples with a configuration object:

``` swift
Quick.configure { (configuration: QuickConfiguration) in
  configuration.registerSharedExamples("something edible") { /* ... */ }
}
```

[...]

PR #766, opened 2018-01-25
Allow example groups to order tests either randomly OR as they are defined

jwfriese:
This PR attempts to add the ordering functionality requested in #759 . 

As written, this PR adds a new argument to `describe`-type and `context`-type DSL methods -- an `order` enum value, which can be one of `.defined` or `.random`.

`.defined` will ensure that the examples are run in the order they are defined within the `describe` or `context` blocks. This includes respecting nested `describe` and `context` blocks in turn. The new test file (`ExplicitOrderingTests.swift`) shows you exactly how you can expect it to handle nested describes. 

`.random` will randomize the examples within the describe, although it takes care to respect the `order` values defined by nested `describe` and `context` blocks. 

Now, there exist some caveats to note here. The most significant of these is the fact that XCode 9 appears to revert to using alphabetical-by-name ordering of tests _only when you use the side tray candies to run the tests_. In other words, XCTest does not appear to respect the ordering we set as the return value of `testInvocations` from `QuickSpec`. I suspect this is related to #744 and #745, but I'm not sure. 

I argue that this PR is still worth adding, because it's still giving many users (anyone using Cmd+U or the command line to run their tests) new goodness. Please let me know if you disagree!

Also, I chose to default the Order to `.defined` because of the [principle of least surprise](https://softwareengineering.stackexchange.com/questions/187457/what-is-the-principle-of-least-astonishment) -- that's closest to how things work now. 

Technically, we should probably default to `.random` since having an explicit ordering to tests is not the best practice. 

Here's some stuff I still want to do with this PR once I get some feedback of the high-level design: 
- ~~complete documentation of how to use the feature~~ ✅ 
- ~~it's probably unnecessary to maintain `childGroups` and `childExamples` as separate collections now within the `ExampleGroup` class~~ True, but not an optimization that I feel is worth it for now.
- ~~add some `context` blocks to the test~~✅ 
- super bonus: allow user to define a sorting function (I actually think this should be done separately though)

### DIFF
--- a/Sources/Quick/DSL/DSL.swift
+++ b/Sources/Quick/DSL/DSL.swift
@@ -65,15 +65,15 @@ public func sharedExamples(_ name: String, closure: @escaping SharedExampleClosu
     - parameter closure: A closure that can contain other examples.
     - parameter flags: A mapping of string keys to booleans that can be used to filter examples or example groups.
 */
-public func describe(_ description: String, flags: FilterFlags = [:], closure: () -> Void) {
-    World.sharedWorld.describe(description, flags: flags, closure: closure)
+public func describe(_ description: String, order: Order = Order.defined, flags: FilterFlags = [:], closure: () -> Void) {
+    World.sharedWorld.describe(description, order: order, flags: flags, closure: closure)
 }
 
 /**
     Defines an example group. Equivalent to `describe`.
 */
-public func context(_ description: String, flags: FilterFlags = [:], closure: () -> Void) {
-    World.sharedWorld.context(description, flags: flags, closure: closure)
+public func context(_ description: String, order: Order = Order.defined, flags: FilterFlags = [:], closure: () -> Void) {
+    World.sharedWorld.context(description, order: order, flags: flags, closure: closure)
 }
 
 /**
@@ -198,8 +198,8 @@ public func pending(_ description: String, closure: () -> Void) {
     Use this to quickly mark a `describe` closure as pending.
     This disables all examples within the closure.
 */
-public func xdescribe(_ description: String, flags: FilterFlags, closure: () -> Void) {
-    World.sharedWorld.xdescribe(description, flags: flags, closure: closure)
+public func xdescribe(_ description: String, order: Order = Order.defined, flags: FilterFlags, closure: () -> Void) {
+    World.sharedWorld.xdescribe(description, order: order, flags: flags, closure: closure)
 }
 
 /**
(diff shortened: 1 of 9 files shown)


jwfriese:
By the way, this pull request also solves #152 🎉

jwfriese:
@modocache or @jeffh - This has been out here for about a week now, and has a couple of positive reactions. I'd love to hear feedback if you have any. Otherwise, I think this is ready to go! 😃

Sources/Quick/DSL/DSL.swift
```diff
@@ -65,15 +65,15 @@ public func sharedExamples(_ name: String, closure: @escaping SharedExampleClosu
...
+public func describe(_ description: String, order: Order = Order.defined, flags: FilterFlags = [:], closure: () -> Void) {
+    World.sharedWorld.describe(description, order: order, flags: flags, closure: closure)
 }
 
 /**
     Defines an example group. Equivalent to `describe`.
 */
-public func context(_ description: String, flags: FilterFlags = [:], closure: () -> Void) {
-    World.sharedWorld.context(description, flags: flags, closure: closure)
+public func context(_ description: String, order: Order = Order.defined, flags: FilterFlags = [:], closure: () -> Void) {
```
NachoSoto: Nit: I would avoid duplicating the type in all these: `order: Order = .defined`.

Sources/Quick/DSL/World+DSL.swift
```diff
@@ -143,7 +143,7 @@ extension World {
         }
         let callsite = Callsite(file: file, line: line)
         let closure = behavior.spec
-        let group = ExampleGroup(description: behavior.name, flags: flags)
+        let group = ExampleGroup(description: behavior.name, order: Order.defined, flags: flags)
```
NachoSoto: Nit: I’d use type inference here: `.defined`.

jwfriese:
Thanks for the review and suggestions @NachoSoto 👍

Thiryn:
Just a small up to know if this feature is still considered or not.

jwfriese:
I'd like to know too. I have a few pull requests waiting to be merged.

VojtaStavik:
+1 for merging this 👍

ArtemLyksa:
+1 to merge this. Should be really helpful.

jwfriese:
Since it looks like this feature is still in demand, I am in the process of updating this branch since it's a year old and had a couple merge conflicts. I currently have the tests all passing locally on my Mac. It's running through CI now. I'll fix any breakages that might find.

cmillani:
+1 to merge this, would love to be able to use this 😃

jradivojsa:
I also believe this PR would benefit all of us greatly. To not include such a magnificent feature is nothing short of a sin.
pls murg 👍

cbauer10:
Any movement on this?

ashikahmad:
Badly missing ordering. Looking forward to see this merged 🤩

ValeFCF:
What is the status for this?

jwfriese:
I have updated this PR now, resolving merge conflicts, and ensuring we still have a clean local `rake` run. I'll address any check failures as necessary. @ikesyo please let me know if there's anything else you need me to address in order to get this code merged.

ZevEisenberg:
+1 from the future. If you're still in the past, maybe stay there for a bit 😅

Sources/Quick/ExampleGroup.swift
```diff
@@ -28,20 +31,76 @@ final public class ExampleGroup: NSObject {
...
+    }
+
+    private var randomizedExamples: [Example] {
+        var randomized = Array(groupUnits)
+        for (firstUnshuffled, unshuffledCount) in zip(randomized.indices, stride(from: randomized.count, to: 1, by: -1)) {
+            #if os(Linux)
+                srandom(UInt32(time(nil)))
+                let offset: Int = Int(random() % (unshuffledCount + 1))
+            #else
+                let offset: Int = numericCast(arc4random_uniform(numericCast(unshuffledCount)))
```
ZevEisenberg, line 81: Since Swift 4.2, perhaps this could be replaced with the newer randomness API?

jessesquires:
Hey there! 👋🏼 

younata:
@jwfriese Hey, I'm coming back to this. I'm curious what you think about changing Quick's default behavior so that it always attempts to execute tests in the order they're defined in (as you've noted, XCTest will ignore that attempt whenever it desires). Additionally, we'll give the guidance that developers should enable "Randomize Execution Order" in the xctestplan when they want to randomize execution order.

This is less configurable than what this PR specifies, but I think overall what I just outlined will be better/easier for people to understand.

PR #767, opened 2018-01-25
Add script to run Linux tests in a container using Docker

jwfriese:
I develop on a Mac. Yesterday, I broke the Linux build working on a [PR to address the concerns raised by another issue](https://github.com/Quick/Quick/pull/766), and found myself wishing I had an easy way to run the Linux test suite locally before pushing, lest I drive the community and myself crazy with the churn. 

I ended up doing some stuff with Docker to see how easy that process could be, and that's where this PR comes from. It creates a new script that builds an image and uses it to run Quick's Linux tests. I smoke tested it with both a compilation error and an actual test failure and it works great!

This commit includes an update to the contributors' documentation explaining the test script and how to use it. 

I understand that this is purely a quality-of-life change for the codebase's developers, and that we should be wary of those types of changes. I'd love feedback on this though, especially feedback on whether the community even believes this is a valuable addition to the repository. 

Thanks, y'all 😃

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,14 +2,14 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**  *generated with [DocToc](http://doctoc.herokuapp.com/)*
 
-- [Welcome to Quick!](#welcome-to-quick!)
+- [Welcome to Quick!](#welcome-to-quick)
   - [Reporting Bugs](#reporting-bugs)
   - [Building the Project](#building-the-project)
   - [Pull Requests](#pull-requests)
     - [Style Conventions](#style-conventions)
   - [Core Members](#core-members)
-    - [Code of Conduct](#code-of-conduct)
   - [Creating a Release](#creating-a-release)
+  - [Testing for Linux from MacOS](#testing-for-linux-from-macos)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -109,3 +109,18 @@ The process is relatively straight forward, but here's is a useful checklist for
 - The script will prompt you to create a new [GitHub release](https://github.com/Quick/Quick/releases).
   - Use the same release notes you created for the tag, but tweak up formatting for GitHub.
 - Announce!
+
+## Testing for Linux from MacOS
+
+If you develop on a Mac, it can prove difficult to run Quick tests against Linux. This can result in 
+the temptation to engage in an unpleasant and expensive cycle where you push to CI, watch your stuff fail
+on Linux, try to fix it, then push again and cross your fingers. 
+
+We'd like to make it easier for you to test your PR contents on Linux before pushing, so Quick provides a
+script that uses Docker to create a Linux container and run Quick's tests on it. To run it, simply execute 
+the following command from the root directory:
+
+```
+./script/test-linux-on-macos
+```
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,4 +123,8 @@ the following command from the root directory:
 ```
 ./script/test-linux-on-macos
 ```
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -2,14 +2,14 @@
...
 **Table of Contents**  *generated with [DocToc](http://doctoc.herokuapp.com/)*
 
-- [Welcome to Quick!](#welcome-to-quick!)
+- [Welcome to Quick!](#welcome-to-quick)
   - [Reporting Bugs](#reporting-bugs)
   - [Building the Project](#building-the-project)
   - [Pull Requests](#pull-requests)
     - [Style Conventions](#style-conventions)
   - [Core Members](#core-members)
-    - [Code of Conduct](#code-of-conduct)
```
jwfriese: I guess this line is breaking some kind of rule from the QuickBot? That's weird because this change just came from running `doctoc` like it says to do.

jwfriese: Nevermind, this is from changing `CONTRIBUTING.md`. I feel as though the documentation for this is most relevant in there, since it only impacts folks who want to contribute to Quick.

jwfriese:
@modocache or @jeffh - What's the standard way to get around the block on making modifications to `CONTRIBUTING.md`? Or is there a better place to put the documentation on how to use the new Linux test running?

jeffh:
I don’t know of a standard way of getting around it. I have been busy with personal stuff for a while now and I haven’t had the time to be able to work on Quick stuff in a while.

jwfriese:
Refreshing this PR, as I continue to find it personally useful, and think others might as well.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY . ./
+
+RUN apt-get -q update && \
+    apt-get -q install -y \
+    lsb-release \
+    rake
+
+RUN ./script/travis-install-linux
+
+ENTRYPOINT ./script/travis-script-linux
```
ikesyo: Instead of using swiftenv as said above, just running `rake test:swiftpm` will be enough?

CONTRIBUTING.md
```diff
@@ -2,14 +2,14 @@
...
   - [Creating a Release](#creating-a-release)
+  - [Testing for Linux from MacOS](#testing-for-linux-from-macos)
```
ikesyo: Could you place this next to the "Building the Project" section?

jwfriese:
@ikesyo - I have applied your suggestions above -- thank you for the review!

As long as the tests here pass, I think this is good to merge in.

script/travis-install-linux
```diff
@@ -1,5 +0,0 @@
-#!/usr/bin/env bash
-set -e
-
-# See: https://github.com/kylef/swiftenv/wiki/Travis-CI
-curl -sL https://gist.github.com/kylef/5c0475ff02b7c7671d2a/raw/621ef9b29bbb852fdfd2e10ed147b321d792c1e4/swiftenv-install.sh | bash
```
ikesyo: This file is still used here: https://github.com/Quick/Quick/blob/9767dfd0016f4654af495bb938f198d14eb8fa9d/.travis.yml#L22

jwfriese: @ikesyo - Oh gosh, I am so sorry! I've restored the file in my latest commit.

jwfriese:
@ikesyo - Seems like this check failed because it couldn't pull Nimble? It doesn't smell like it's related to my change, but I could be mistaken.

script/travis-script-linux
```diff
@@ -1,4 +1,3 @@
 #!/usr/bin/env sh
 
-. ~/.swiftenv/init
```
ikesyo: Travis jobs keep failing because of this. Let's revert this as well and run `rake test:swiftpm` directly in Dockerfile (as stated in https://github.com/Quick/Quick/pull/767/files#r249814869).

ikesyo:
Closing as this is stale.

Issue #770, opened 2018-02-06
Better `pending` test case output


shx7:
Hello guys, I’m working in AppCode, and we are very excited to add support for Quick framework to our IDE soon. But we encountered several problems during implementation of pending test cases support. Essentially it’s not very convenient for us to parse ‘pending’ test cases because pending() produces pretty ambiguous output. Here is an example:
```
class PendingTestsEx : QuickSpec {
  func spec() {
    it("it") {}
    context(“context”) {
      it(“it”) { /* .. */ }
      pending(“it”) { /* .. */ }
      pending(“it”) { /* .. */ }
      it(“it”) { /* .. */ }
     }   
  }
}
```

Xcode console output:

Pending: it
Pending: it
Pending: it
Test Suite 'Selected tests' started at 2018-02-06 17:18:08.129
Test Suite 'Quick - macOSTests.xctest' started at 2018-02-06 17:18:08.135
Test Suite 'PendingTestsEx' started at 2018-02-06 17:18:08.136
Test Case '-[Quick___macOSTests.PendingTestsEx context__it]' started.
Test Case '-[Quick___macOSTests.PendingTestsEx context__it]' passed (0.985 seconds).
Test Case '-[Quick___macOSTests.PendingTestsEx context__it_2]' started.
Test Case '-[Quick___macOSTests.PendingTestsEx context__it_2]' passed (0.005 seconds).
Test Suite 'PendingTestsEx' passed at 2018-02-06 17:18:09.129.
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.990 (0.993) seconds
Test Suite 'Quick - macOSTests.xctest' passed at 2018-02-06 17:18:09.131.
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.990 (0.996) seconds
Test Suite 'Selected tests' passed at 2018-02-06 17:18:09.132.
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.990 (1.003) seconds
Program ended with exit code: 0

As you can see, It's unclear, which one of 'it' is pending. For 'pending' cases I'd love to see test id's as for usual tests. Probably output may be something like (now we can identify where pending tests are, and put them into test tree in right place):

Pending: Quick___macOSTests.PendingTestsEx it
Pending: Quick___macOSTests.PendingTestsEx context__it
Pending: Quick___macOSTests.PendingTestsEx context__it_3
Test Suite 'Selected tests' started at 2018-02-06 17:18:08.129
[...]

PR #491, opened 2016-02-23
Pending Example Impovement

takecian:
This PR is related to #490 .
- print detailed description for pending example
- execute before/after hooks for pending example
- rename Filter.pending to Filter.excluded

please review, thanks.
Takeshi.

Issue #490, opened 2016-02-20
Using pending causes custom Configuration hooks not to run


JeffBNimble:
If you create a custom Configuration class and add your own custom closures for:

beforeSuite,
afterSuite,
beforeEach
afterEach
and then use pending (instead of it/xit/fit) on your Quick tests, none of the configuration closures for beforeEach and afterEach are called, beforeSuite or afterSuite are called.

I typically begin my testing by writing out my describe/context(s) and use pending on everything. When I run a test with ALL pending and no it/xit/fit, none of the hooks are invoked.

Here is a very simple example of what I'm talking about:

``` swift
class SQLQueryOperationPendingSpec : QuickSpec {
    override func spec() {
        describe("With a SQLQueryOperation") {

            context("when executing a query without a table") {

                pending("it will throw a MissingTableName exception") {}

            }

            context("when executing a query with named parameters") {

                pending("it will execute the query against the database") {}

            }

            context("when executing a query with an array of parameters") {

                pending("it will execute the query against the database") {}
            }

        }
    }
}
```

FWIW, I have not tested what happens when I have tests with a mix of pending/it, so I don't know what happens there.

JeffBNimble:
It screwed my code formatting up for some reason.

briancroom:
To Quick, a pending test isn't actually a test at all, just a placeholder for one. beforeEach and afterEach are called for each test, but there aren't any tests in your example, resulting in an empty test suite.

JeffBNimble:
OK, understood. The only possible confusing thing then is the way Quick prints pending tests. Here is what is printed to the console in my scenario.

Pending: it will throw a MissingTableName exception
Pending: it will execute the query against the database
Pending: it will execute the query against the database

Since it only prints the text of the pending item, it's somewhat confusing. If I define lots of pending tests at first, which I commonly do, and some or many have similar pending descriptions, then it looks like things get duplicated since nothing else is printed.

I realize this is relatively minor. Would it be your suggestion then to use unique pending descriptions so as to provide additional context on exactly what is pending?

briancroom:
I see, interesting. 

I imagine Quick should be constructing the full test name (including description/context strings) when printing the statements about the pending tests. That would clarify things, right?

JeffBNimble:
I think it would. I was trying to do something similar with the hooks, but since they don't run, I'm unable to construct a meaningful description of pending tests.

I was attempting to do something like this:

With a SQLQueryOperation
    when executing a query without a table
[...]

PR #771, opened 2018-02-07
Unique identifiers for all test examples

jwfriese:
This PR addresses #770 by doing two things: 
1) Assigning unique identifiers to all examples based on the description chain of the example, adding numbers at the end to deduplicate if necessary. 
2) Elevating `pending` elements to full-fledged example objects that print these unique identifiers but still do not run their content. 

There are a couple specific things that I would appreciate feedback on:
- For the sake of trying to change one thing at a time (at least in spirit), I changed the `Filter.pending` flag to `Filter.excluded`. This is because I did not know whether there was some strong reason that `xit` and the like were removed entirely from the Quick test list as opposed to just neutering their content. I'll lean on @modocache and @jeffh to hopefully provide any history they can. It makes sense to me and would simplify the code to just treat `x`-type examples the same as pending, but I don't want to make that decision alone. 

  Also, @shx7, will you need those `xit`, `xdescribe`, etc examples to print `Pending: blah_blah_blah` as well? If so, then we'll definitely have to make some more changes. Fortunately, I already experimented with that and it will be easy for me to do. I just need to know whether that's what we want. 

- Tests only start appending digits once its apparent they are duplicated. In other words, a set of tests which happen to be duplicated will have the following identifiers:
```
test_with_duplicated_identifier
test_with_duplicated_identifier_2
test_with_duplicated_identifier_3
```
Notice that the first one does not have a `1` appended to it. @shx7 will that satisfy the ask? 

I still have a couple things to do on this PR, but wanted to start a review of the code and a conversation about its progress ASAP. I'll maintain the list here:

- Prepend unique identifiers with test class name
- (maybe) Refactor `xit` and its like elements to use the new pending code

Cheers, y'all.

### DIFF
--- a/Quick.xcodeproj/project.pbxproj
+++ b/Quick.xcodeproj/project.pbxproj
@@ -125,6 +125,9 @@
 		7B5358CE1C3D4FBC00A23FAA /* ContextTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7B5358CA1C3D4E2A00A23FAA /* ContextTests.swift */; };
 		7B5358CF1C3D4FBE00A23FAA /* ContextTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7B5358CA1C3D4E2A00A23FAA /* ContextTests.swift */; };
 		7B5358D01C3D4FC000A23FAA /* ContextTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7B5358CA1C3D4E2A00A23FAA /* ContextTests.swift */; };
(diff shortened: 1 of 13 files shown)


jwfriese:
I've started updating this PR as well, similar to the other two (https://github.com/Quick/Quick/pull/767 and https://github.com/Quick/Quick/pull/766).

jwfriese:
@ikesyo - The file with the "c99 identifier" behavior is the only one that had to significantly change as part of my work to rebase this back into a happy state with `master`. I _think_ that my accommodations as part of the merge keep in good faith with the changes that originally were made in there, but please feel free to suggest or discuss any concerns you have or things you'd like to see done differently in there.

ikesyo:
This is related to #490 and #491.

jessesquires:
Hey there! 👋🏼 

PR #781, opened 2018-04-17
Support Cocoapods 1.5

### LINKED ISSUE #780: Can't compile because "Nimble/Nimble-Swift.h" can't be found. CocoaPods 1.5 & static
I've been waiting for cocoaPods 1.5 to come out so I can finally use Quick and Nimble! The reason I couldn't before was that some of the pods I use (hockeyapp and new relic) use static libraries. This meant I could not have use_frameworks! in my Podfile because that led to the transitive_dependencies issue when mixing dynamic and static libraries with Cocoapods. I should perhaps also mention that I have a private pod that integrates with hockeyapp and new relic. I think it's the combination of all those things that leads to the transitive_dependencies issue.

CocoaPods 1.5 http://blog.cocoapods.org/CocoaPods-1.5.0/ now supports Swift Static Libraries. Therefore I thought I'd now be able to use Quick and Nimble.

However, building Quick and Nimble in Cocoapods 1.5 as static libraries seems to cause an error to to occur in Quick that has happened before in Nimble. Essentially there seem to be a private objective-c test file (QCKDSL.m) that references a bridging header (Nimble/Nimble-Swift.h) that does not exist. This is the similar issue that impacted Nimble: https://github.com/Quick/Nimble/pull/283 

To use static libraries, you just remove "use_frameworks!"  from the Podfile, so that's what I did as well as include Quick and Nimble. 

I hoped that my tests would build and I'd be able to start using Quick and Nimble.

What happened instead was a build error: 

Can't compile because "Nimble/Nimble-Swift.h" can't be found. 

### Environment

 - Quick: 1.2.0
 - Nimble: 7.1.0
 - Xcode Version: 9.2
 - Swift Version: 4.0 

 - Cocoapods: 1.5

I've not included a project, because my project is large and involved. Also, I'm assuming that any project will have this issue if you try to include it as a static library by removing "use_frameworks!"  from the Podfile. Plus I'm assuming that the solution will be similar to what ever the fix was that resolved what seems a similar issue that occurred with Nimble before.

davidahouse:
This PR fixes #780 in relation to Cocoapods 1.5. In this new version of Cocoapods, some things have changed in how header files are found. The recommendation from the Cocoapods team seems to be to change the imports to this:

```
#if __has_include("Quick-Swift.h")
#import "Quick-Swift.h"
#else
#import <Quick/Quick-Swift.h>
#endif
```

I tested this against my app and it seems to work fine with no side-effects.

### DIFF
--- a/Sources/QuickObjectiveC/DSL/World+DSL.h
+++ b/Sources/QuickObjectiveC/DSL/World+DSL.h
@@ -1,4 +1,9 @@
+#if __has_include("Quick-Swift.h")
+#import "Quick-Swift.h"
+#else
 #import <Quick/Quick-Swift.h>
+#endif
+
 
 @interface World (SWIFT_EXTENSION(Quick))
 - (void)beforeSuite:(void (^ __nonnull)(void))closure;
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -1,7 +1,12 @@
 #import "QuickSpec.h"
 #import "QuickConfiguration.h"
 #import "World.h"
+#if __has_include("Quick-Swift.h")
+#import "Quick-Swift.h"
+#else
 #import <Quick/Quick-Swift.h>
+#endif
+
 
 static QuickSpec *currentSpec = nil;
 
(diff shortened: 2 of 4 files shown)


davidahouse:
@ikesyo fixed!

davidahouse:
It looks correct on Github. I did force-push to my branch so if you have it locally then you should delete and grab it again. Sorry about that.

ikesyo:
The three extra empty lines I noted above still exist on GitHub GUI. But those are just stylistic ones so let’s merge this as is.

Thanks for the pull request!

PR #784, opened 2018-04-21
Fixed testing error on Linux

felix91gr:
* While building on Linux, `isTesting` would evaluate to `false` and therefore `swift test` wouldn't finish compiling the test suites.
* Fixed by introducing a conditional compilation flag that while on Linux, defaults SPM to always downloading `Nimble`.
* Behavior in MacOS is unchanged.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -40,8 +40,17 @@ var package = Package(
     }()
 )
 
-if isTesting {
-  package.dependencies.append(contentsOf: [
-    .Package(url: "https://github.com/Quick/Nimble.git", majorVersion: 7),
-  ])
+var downloadNimble = false
+
+#if os(macOS)
+    downloadNimble = isTesting
+
+#else 
+    downloadNimble = true
+#endif
+
+if(downloadNimble) {
+    package.dependencies.append(contentsOf: [
+        .Package(url: "https://github.com/Quick/Nimble.git", majorVersion: 7),
+    ])
 }


ikesyo:
Thanks for the PR! However, we are successfully running tests on Linux as seen in https://travis-ci.org/Quick/Quick/builds/368966814. So I’m not sure why the proposed changes are required/requested.

And also once we move to PackageDescription v4 syntax (which will be a part of forthcoming Quick 2.0), the conditional flag will be unnecessary and removed.

ikesyo:
You should also check this https://github.com/Quick/Quick/blob/a0e1457029c2a451a321fdd3a6a6f36ac367010f/Package%40swift-4.swift

ikesyo:
See https://github.com/Quick/Quick/pull/637 and https://github.com/Quick/Quick/pull/721 for the historical reasons.

ikesyo:
So I recommend that you use `rake test:swiftpm` for now.

- https://github.com/Quick/Quick/blob/3e3023569c8d4c4a0d000f58db765df53041117f/script/travis-script-linux#L4
- https://github.com/Quick/Quick/blob/3e3023569c8d4c4a0d000f58db765df53041117f/Rakefile#L36-L40

felix91gr:
Ahhhh I see, I should've ran the tests with `rake`. It works, thanks! :)

That makes me ask: why isn't `swift test` used directly? I don't know much about `rake`, so I can't really guess why you're using it, but you probably have a good reason to do so.

ikesyo:
> That makes me ask: why isn't swift test used directly? I don't know much about rake, so I can't really guess why you're using it, but you probably have a good reason to do so.

The original concern raised in https://github.com/Quick/Quick/issues/545 would be a help.

> It's unfortunate, but most Swift package indexes list Quick as having a dependency: Nimble. This is not true, of course -- Quick isn't dependent upon Nimble, it simply uses Nimble as part of the test suite that tests Quick.

PR #787, opened 2018-04-27
Add @escaping to Swift example

bellebethcooper:
`SharedExampleContext` should be marked with `@escaping` in Swift example code.


modocache:
Thanks, @bellebethcooper! 💯

Issue #785, opened 2018-04-21
Unrecognized Selector crash in QuickSpec.m on Xcode 9.3


zcsipler:
When I try to run tests there is a crash in QuickSpec.m line 113:

_NSString *originalName = example.name.qck_c99ExtendedIdentifier;_

And it says:

2018-04-21 14:13:36.409506+0200 OneApp[90978:6333181] -[Swift._NSContiguousString qck_c99ExtendedIdentifier]: unrecognized selector sent to instance 0x604000455990

I got this crash since I use Xcode 9.3 and Swift 4.1. 
Quick version is 1.3.0

amadeu01:
I'm commenting just to know if that got fixed.

kevinwo:
I too am receiving this error, under the following circumstances:

1. This occurs any time I implement an `it()` block anywhere in any spec of my test target.
2. My test target is associated with a Cocoa Touch Framework, internal to my project. All the tests in my other test target, associated with my main app, run normally as expected.
3. Xcode 9.3 / Swift 4.1 / Quick 1.3.0 / CocoaPods 1.5.2

I've tried to reproduce the issue in a new test project, but am unable to – all tests in the internal Cocoa Touch Framework run as expected with no issue.

zcsipler:
For us the solution was to change building QuickSpec from static to dynamic library. We use Carthage and as long as we built it as a static lib this error appeared but when we built it as a dynamic lib the issue did not come up anymore. Though I have no idea what is the relation but we had to do the same also with Nimble to make it work properly.

ikesyo:
> We use Carthage and as long as we built it as a static lib this error appeared but when we built it as a dynamic lib the issue did not come up anymore.

Thanks for the clue, that is helpful. 👍

ikesyo:
Fixed by #803.

PR #792, opened 2018-05-23
[1.x] `QuickSpec+QuickSpec_MethodList` should be in the test targets, not in the library targets

ikesyo:
I'll say that this is a preliminary task for fixing #785.

### DIFF
--- a/Tests/QuickTests/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
+++ b/Tests/QuickTests/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
@@ -4,27 +4,19 @@
 
 @implementation QuickSpec (QuickSpec_MethodList)
 
-/**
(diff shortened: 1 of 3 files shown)


Tests/QuickTests/QuickTests/Helpers/QuickSpec+QuickSpec_MethodList.m
```diff
@@ -4,27 +4,19 @@
 
 @implementation QuickSpec (QuickSpec_MethodList)
 
-/**
- *  This method will instantiate an instance of the class on which it is called,
- *  returning a list of selector names for it.
- *
- *  @warning Only intended to be used in test assertions!
```
ikesyo: As the message said, the function should not be defined in `Quick-macOS` and so on.

ikesyo:
Ping @Quick/core

ikesyo:
Thanks @sharplet!

PR #802, opened 2018-06-20
Add workaround for running parallel tests on Xcode 10

### LINKED ISSUE #801: Issues running tests in parallel with Xcode 10

This has been a a requested feature for a while (#383), and it's finally here with `Xcode 10.0`!
However looks like `Quick` will need some changes to support it correctly.

### What did you do?

My tests pass in Xcode `9.4` and `10.0 beta 1`, but not when enabling the new parallel option:
screen shot 2018-06-17 at 10 12 06 am

The failure occurs whether it's randomized or not.

Tests pass.

An exception is thrown:
> 'context' cannot be used inside 'it', 'context' may only be used inside 'context' or 'describe'.

### Environment

 - Quick: [37ed3f5e28acc62ab8e2d612ea49a78f44c7b79a](https://github.com/Quick/Quick/commit/37ed3f5e28acc62ab8e2d612ea49a78f44c7b79a)
 - Nimble: [69bb1612cf74ae072fed2675eebe7dab51da4d26](https://github.com/Quick/Nimble/commit/69bb1612cf74ae072fed2675eebe7dab51da4d26)
 - Xcode Version: `Version 10.0 beta 1 (10L176w)`
 - Swift Version: `Apple Swift version 4.2 (swiftlang-1000.0.16.7 clang-1000.10.25.3)`

 - Carthage: `0.29.0`

marcelofabri:
Fixes #801

XCTest can call a test case class’s `+initialize` on a non-main thread when running tests in parallel.

This forces the initialization code to run on the main queue.

### DIFF
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -26,28 +26,35 @@ static QuickSpec *currentSpec = nil;
  included an expectation outside of a "it", "describe", or "context" block.
  */
 + (void)initialize {
-    [QuickConfiguration initialize];
+    // Workaround for running parallel tests on Xcode 10 (https://github.com/Quick/Quick/issues/801)
+    // From Xcode 10 Beta 2 release notes:
+    // "Tests using a third-party testing framework that integrates with XCTest may experience test
+    // failures when running tests in parallel due to a test case class’s +initialize method being
+    // called on a non-main thread. (40939090)"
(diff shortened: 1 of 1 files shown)


marcelofabri:
Actually, this would probably cause other issues 🤔

ikesyo:
- If in main thread (main queue), execute the body directly
- If in non-main thread, execute the body with `dispatch_sync(dispatch_get_main_queue(), ^{ ... })`

may work. Async execution should be avoided I think.

PR #791, opened 2018-05-19
[1.x] Set `APPLICATION_EXTENSION_API_ONLY` to `YES`

ikesyo:
This enables us to use Swift internal classes from Objective-C side with importing the generated header.

Fixes #749. See also https://github.com/Quick/Nimble/pull/525.

PR #803, opened 2018-06-21
[1.x] Fix unrecognized selector crash on static linking

### LINKED ISSUE #785: Unrecognized Selector crash in QuickSpec.m on Xcode 9.3

ikesyo:
~~Extension methods or properties for NSObject subclasses are somehow invisible from the Objective-C runtime on static linking, so let's make a wrapper class.~~

Extension methods or properties for NSObject subclasses are invisible from the Objective-C runtime on static linking unless the consumers add `-ObjC` linker flag, so let's make a wrapper class to mitigate that situation.

Fixes #785.

### DIFF
--- a/Sources/Quick/NSString+C99ExtendedIdentifier.swift
+++ b/Sources/Quick/NSString+C99ExtendedIdentifier.swift
@@ -35,11 +35,11 @@ extension NSString {
 ///
 /// See: https://github.com/Quick/Quick/issues/785
 @objc
-public class QCKObjCStringUtils: NSObject {
+class QCKObjCStringUtils: NSObject {
     override private init() {}
 
     @objc
-    public static func c99ExtendedIdentifier(from string: String) -> String {
+    static func c99ExtendedIdentifier(from string: String) -> String {
         return string.c99ExtendedIdentifier
     }
 }
--- a/Sources/Quick/NSString+C99ExtendedIdentifier.swift
+++ b/Sources/Quick/NSString+C99ExtendedIdentifier.swift
@@ -1,8 +1,7 @@
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Foundation
 
-public extension NSString {
-
+extension NSString {
     private static var invalidCharacters: CharacterSet = {
         var invalidCharacters = CharacterSet()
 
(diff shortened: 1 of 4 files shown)


ikesyo:
@zcsipler Could you try this branch?

ikesyo:
Oh I completely forgot the linker flag! 💦 Indeed, adding the flag to test targets which uses Quick (e.g. `Quick - macOSTests`, `QuickFocused - macOSTests`, and so on), not to Quick itself, fixes the issue.

But forcing consumers to add the flag to their test targets just for this internal implementation detail will not be ideal I think. What do you think?

Sources/Quick/NSString+C99ExtendedIdentifier.swift
```diff
@@ -22,12 +21,25 @@ public extension NSString {
         return invalidCharacters
     }()
 
-    @objc(qck_c99ExtendedIdentifier)
```
sharplet: Because there's a workaround, I think we should preserve this public API.

Sources/Quick/NSString+C99ExtendedIdentifier.swift
```diff
@@ -22,12 +21,25 @@ public extension NSString {
...
 
         return result.isEmpty ? "_" : result
     }
 }
+
+/// Extension methods or properties for NSObject subclasses are somehow invisible
+/// from the Objective-C runtime on static linking, so let's make a wrapper class.
+///
+/// See: https://github.com/Quick/Quick/issues/785
+@objc
```
sharplet: Might be a good idea to prefix this new class?

Sources/Quick/NSString+C99ExtendedIdentifier.swift
```diff
@@ -22,12 +21,25 @@ public extension NSString {
...
     var c99ExtendedIdentifier: String {
         let validComponents = components(separatedBy: NSString.invalidCharacters)
         let result = validComponents.joined(separator: "_")
 
         return result.isEmpty ? "_" : result
     }
 }
+
+/// Extension methods or properties for NSObject subclasses are somehow invisible
+/// from the Objective-C runtime on static linking, so let's make a wrapper class.
```
sharplet: Because we know that `-ObjC` will expose these to the runtime, this comment might need an update.

Sources/Quick/NSString+C99ExtendedIdentifier.swift
ikesyo: I'm okay to preserve this for 1.x but I'd like to hide this in 2.x, because I assume the reason why this API is public is to expose it to the generated header. We can now expose internal Swift APIs to the generated header thanks to #791, so there will be no need to make it public.

sharplet: Can we explicitly deprecate it?

ikesyo: I tried adding `@available(*, deprecated)` but `__attribute__ ((deprecated))` is not exposed to the generated header unfortunately.

ikesyo:
@sharplet Thanks for the detailed review! 💖 I updated the pull request.

ikesyo:
Thanks!

> Is there somewhere in the documentation that we could add a tip about the -ObjC linker flag?

I will try to update `InstallingQuick.md` in a separate pr. Let's metge this as is.

PR #821, opened 2018-09-27
Fix Xcode 10 warning

spekke:
This PR will suppress the following warning:
```
Pods/Quick/Sources/Quick/Behavior.swift#L7: static declarations are implicitly ‘final’; use ‘public’ instead of ‘open’ 
open static var name: String { return String(describing: self) }
```

### DIFF
--- a/Sources/Quick/Behavior.swift
+++ b/Sources/Quick/Behavior.swift
@@ -4,7 +4,7 @@
 
 open class Behavior<Context> {
 
-    open static var name: String { return String(describing: self) }
+    public static var name: String { return String(describing: self) }
     /**
      override this method in your behavior to define a set of reusable examples.
 


Sources/Quick/Behavior.swift
```diff
@@ -4,7 +4,7 @@
 
 open class Behavior<Context> {
 
-    open static var name: String { return String(describing: self) }
+    public static var name: String { return String(describing: self) }
```
ikesyo: This makes sense to me :+1:

But there is another possibility that `open class var` is the original intent. How do you think @Quick/core?

wongzigii:
I am good with this.

wongzigii:
Fix #823

Issue #656, opened 2016-12-12
All tests interrupted, continues with Selected tests, ends with single failure Exit status: 65


antonys1:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

I used fastlane scan to run my test target.

### What did you expect to happen?

I expected to see `All tests` in the console output and all tests pass. (This happens on one machine)

```
[09:52:26]: ▸ 	 Executed 124 tests, with 0 failures (0 unexpected) in 1.811 (1.917) seconds
[09:52:26]: ▸ 
[09:52:29]: ▸ Test Succeeded
+--------------------+-------+
|        Test Results        |
+--------------------+-------+
| Number of tests    | 124   |
| Number of failures | 0     |
+--------------------+-------+
```

### What actually happened instead?

I see `All tests` in the console output, the app then quits in the simulator, I see `Selected tests` in the console output, the app starts in the simulator which runs selected tests and always ends with one failure. The failed test on inspection cannot genuinely be failing. So I suspect something else has failed coinciding with the execution of this test.

Given that the expected result occurs on one machine, I was hoping someone may be able to explain what might cause the observed behaviour on other machines to occur? Specifically, what causes `Selected tests` to appear on one machine and not another? (both machines are pulling from the same repo, and have the same version of Xcode, Cocoapods, Quick, Nimble, fastlane, scan)

```
[11:02:57]: Running Tests: ▸ Touching ExampleTests.xctest
[11:02:57]: ▸ Signing /Users/jenkins/Library/Developer/Xcode/DerivedData/Example-btezuwhrwluvblektjlreaboyhtd/Build/Intermediates/CodeCoverage/Products/Debug-iphonesimulator/Example.app/PlugIns/ExampleTests.xctest
[11:03:05]: ▸ All tests
[11:03:05]: ▸ Quick.framework
[11:03:05]: ▸ Test Suite ExampleTests.xctest started
[11:03:05]: ▸ Example1Test
[11:03:06]: ▸ ✓ testExample1_A (0.081 seconds)
[11:03:06]: ▸ ✓ testExample1_B (0.060 seconds)
[11:03:06]: ▸ Example2Test
[11:03:06]: ▸ ✓ testExample2_A (0.023 seconds)
[11:03:06]: ▸ ✓ testExample2_B (0.023 seconds)
[11:03:06]: ▸ Example3Test
[...]

Issue #600, opened 2016-09-03
[Swift 3.0] `swift test` fails with `fatal error: use of unimplemented initializer 'init(invocation:)'`


norio-nomura:
Using `swift-3.0` branch, `swift test` fails on macOS as following:

``` console
$ swift build
Cloning https://github.com/norio-nomura/Nimble
HEAD is now at 8ed5cb9 Merge pull request #6 from ikesyo/swift3-beta-6
Resolved version: 5.0.0-alpha.30p6
Compile Swift Module 'Nimble' (46 sources)
Compile Swift Module 'Quick' (21 sources)
$ swift test
Compile Swift Module 'QuickTests' (20 sources)
Linking ./.build/debug/QuickPackageTests.xctest/Contents/MacOS/QuickPackageTests
Test Suite 'All tests' started at 2016-09-03 11:15:16.837
Test Suite 'QuickPackageTests.xctest' started at 2016-09-03 11:15:16.838
Test Suite 'AfterEachTests' started at 2016-09-03 11:15:16.838
Test Case '-[QuickTests.AfterEachTests testAfterEachIsExecutedInTheCorrectOrder]' started.
/Users/norio/github/Quick/Sources/Quick/QuickSpec.swift: 6: 12: fatal error: use of unimplemented initializer 'init(invocation:)' for class 'Quick.QuickSpec'
```

Current implementation of Quick needs instance of `QuickSpec` for enumerating test methods. But It fails to create instance in Swift.
Because:
- `QuickSpec.swift` is used instead of `QuickSpec.m` because building Objective-C is not supported by Swift Package Manager
- `NSInvocation` is unimplemented on Swift
- Swift implicitly generates codes for catching calls of `init(invocation:)` that uses unimplemented type in Swift
- The wrapper function is generated if `public required override init()` is declared in `QuickSpec`
- `init(invocation:)` is called from `init()` by superclass of `XCTestCase`

Call stack on fatal error:

``` console
$ lldb `xcrun -find xctest` -- .build/debug/QuickPackageTests.xctest
(lldb) target create "/Applications/Xcode-beta.app/Contents/Developer/usr/bin/xctest"
Current executable set to '/Applications/Xcode-beta.app/Contents/Developer/usr/bin/xctest' (x86_64).
(lldb) settings set -- target.run-args  ".build/debug/QuickPackageTests.xctest"
(lldb) process launch
[...]

Issue #427, opened 2015-11-22
Tests partially appearing in test inspector or not at all


mutech:
Test inspector shows only some tests (in this case here 4/5 but sometimes it only shows a single test with unspecifiedMethod_(null) as test name, sometimes 3/5 and other combinations). I think it's a coincidence that it does not show the failed test only, right before it did not show a succeeded test, can't reproduce it anymore.

I cleared derived data, rebuild in various combinations. Results are fuzzy.

Using Xcode 7.1, the framework I'm testing is an Objective-C framework.

Here the test inspector view:

![Test inspector](https://cloud.githubusercontent.com/assets/3248524/11321985/afff4fa4-90d5-11e5-97ef-4f885ad15864.png)

The corresponding test results

```
04:51:56.174 xctest[83634:11803643] _XCT_testBundleReadyWithProtocolVersion:minimumVersion: reply received
04:51:56.176 xctest[83634:11803643] _IDE_startExecutingTestPlanWithProtocolVersion:16
Test Suite 'All tests' started at 2015-11-22 04:51:56.218
Test Suite 'AKABeaconQuickTests.xctest' started at 2015-11-22 04:51:56.219
Test Suite 'UITextViewBindingsSpec' started at 2015-11-22 04:51:56.220
Test Case '-[AKABeaconQuickTests.UITextViewBindingsSpec UITextView__textBinding__when_bound__does_not_change_its_initial_text_UsersmugitakaiosbeaconAKABeaconAKABeaconQuickTestsUITextViewBindingsSpecswift_68]' started.
Test Case '-[AKABeaconQuickTests.UITextViewBindingsSpec UITextView__textBinding__when_bound__does_not_change_its_initial_text_UsersmugitakaiosbeaconAKABeaconAKABeaconQuickTestsUITextViewBindingsSpecswift_68]' passed (0.009 seconds).
Test Case '-[AKABeaconQuickTests.UITextViewBindingsSpec UITextView__textBinding__when_bound__when_observing_changes__updates_text_to_bound_value_UsersmugitakaiosbeaconAKABeaconAKABeaconQuickTestsUITextViewBindingsSpecswift_76]' started.
Test Case '-[AKABeaconQuickTests.UITextViewBindingsSpec UITextView__textBinding__when_bound__when_observing_changes__updates_text_to_bound_value_UsersmugitakaiosbeaconAKABeaconAKABeaconQuickTestsUITextViewBindingsSpecswift_76]' passed (0.000 seconds).
[...]

PR #829, opened 2018-12-26
Alternative way to integrate Quick an Nimble without losing test registration and code completion, also adds full swift4.2 support

doozMen:
This possible relates to the following issues: #656 , #427, #219, #534, #238, #600, 

* I moved all framework search paths for all targets up to the project and rest inherits
* I added $(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME) so it looks for the right framework for that platform and the build configuration (Debug / Release)
* Added XCTest as an explicit linked dependency to every target
* Made the target swift_version = 4.2 for every target
* Added example integration with subproject approach
*  Made a dependend build on Nimble via Carthage with submodule

 - [x] Does this have tests?
 - [ ] Does this break the public API (Requires major version bump)? No
 - [ ] Is this a new feature (Requires minor version bump)? No


ikesyo:
Closing this for the same reason as https://github.com/Quick/Nimble/pull/618. Thanks for the PR though!

PR #843, opened 2019-03-29
[SwiftPM] Update Package.swift to support Swift 5

devxoul:
As upgrading to Swift 5, without having `platforms` the default deployment target is set to the latest version of each OS. It causes a compile error.

* Add `platforms`
* Add `.v5` to `swiftLanguageVersions`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,9 +1,12 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 
 import PackageDescription
 
 let package = Package(
     name: "Quick",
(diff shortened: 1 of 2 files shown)


ikesyo:
Thanks for the PR, though this breaks backward compatibility.

ikesyo:
Ref: [SE-0236](https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md)

ikesyo:
I'll merge this once CI passes.

PR #634, opened 2016-11-04
Bot for Quick

wongzigii:
Discuss in #631  and #575 

This PR aims at creating a bot @QuickBot (using danger) to handle chore job (e.g. tagging a label, automatic response new coming PR or issues, CI fails attention)

Would love to hear voice (rules) about @QuickBot to met your appetite. Please checkout the Dangerfile which includes some rules for this bot.

Known warning/fails conditions: 
- [x] Modifying *immutable* files(LICENSE / CONTRIBUTING.md / COC.md)
- [x] Big PR
- [x] Work in process
- [x] Specific files (Quick.podspce) edits by external contributors
- [x] Empty description for PR
- [x] Warn if there's no tests when modifying swift source
- [x] SwiftLint

Require [GitHub api](https://developer.github.com/v3/issues/labels/#add-labels-to-an-issue): 
- [ ] <strike>Mark PRs with "WIP" in the title with don't merge yet<strike>
- [ ] <strike>Automatically marked PRs with GH Review: needs-review<strike>

PR #848, opened 2019-04-06
Support QuickSpec.current for SwiftPM builds as well on Darwin

ikesyo:
Ref: #645

Linux support will be added in a separate PR.

And also this is a preparation for #680 which uses `QuickSpec.current`.

 - [x] Does this have tests? Already exists
 - [x] Does this have documentation? Already exists


Sources/Quick/QuickSpec.swift
```diff
@@ -72,12 +84,13 @@ open class QuickSpec: QuickSpecBase {
...
     private static func addInstanceMethod(for example: Example, classSelectorNames selectorNames : inout Set<String>) -> Selector {
-        let block: @convention(block) (QuickSpec) -> Void = { _ in
+        let block: @convention(block) (QuickSpec) -> Void = { spec in
+            spec.example = example
             example.run()
         }
         let implementation = imp_implementationWithBlock(block as Any)
 
-        let originalName = example.name
+        let originalName = example.name.c99ExtendedIdentifier
```
ikesyo: There had been incompatibility with Xcode builds so fixed it.

https://github.com/Quick/Quick/blob/7e6b060b41e0110a5be1f4af324000acbfc21ae7/Sources/QuickObjectiveC/QuickSpec.m#L113

ikesyo:
https://github.com/Quick/Quick/pull/634#issuecomment-480521569

I don't have an access to @QuickBot account, we might need to disable running danger (for a while).

phatblat:
Opened #849 to track the danger/quickbot issue.

ikesyo:
Thanks @phatblat for the review and filing the issue!

PR #850, opened 2019-04-08
Support QuickSpec.current on Linux

ikesyo:
Ref: #645, #848

This is a preparation for #680 which uses `QuickSpec.current`.

 - [x] Does this have tests? Already exists
 - [x] Does this have documentation? Already exists

### DIFF
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -4,42 +4,43 @@ import Quick
 @testable import QuickTests
 
 Quick.QCKMain([
-    FunctionalTests_AfterEachSpec.self,
     AfterSuiteTests.self,
+    Configuration_AfterEachSpec.self,
+    Configuration_BeforeEachSpec.self,
+    CurrentSpecTests.self,
+    FunctionalTests_AfterEachSpec.self,
     FunctionalTests_BeforeEachSpec.self,
     FunctionalTests_BeforeSuite_BeforeSuiteSpec.self,
     FunctionalTests_BeforeSuite_Spec.self,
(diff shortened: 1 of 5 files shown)


Tests/LinuxMain.swift
```diff
@@ -4,42 +4,43 @@ import Quick
 @testable import QuickTests
 
 Quick.QCKMain([
-    FunctionalTests_AfterEachSpec.self,
     AfterSuiteTests.self,
+    Configuration_AfterEachSpec.self,
+    Configuration_BeforeEachSpec.self,
+    CurrentSpecTests.self,
```
ikesyo: This is added.

Tests/QuickTests/QuickTests/FunctionalTests/CurrentSpecTests.swift
```diff
@@ -2,12 +2,18 @@ import Quick
...
 class CurrentSpecTests: QuickSpec {
     override func spec() {
         it("returns the currently executing spec") {
-            expect(QuickSpec.current.name).to(match("currently_executing_spec"))
+            let name: String = {
+                let result = QuickSpec.current.name
+                #if canImport(Darwin)
+                return result.replacingOccurrences(of: "_", with: " ")
+                #else
+                return result
```
ikesyo: Because we don't use example name to selector name conversion on Linux (there is no Objective-C runtime off-course 😃), we don't need to care about whether white spaces are used or not in example names.

ikesyo:
Please let me go.

PR #856, opened 2019-04-14
Unify test target product names

ikesyo:
This is important for "Objective-C Generated Interface Header Name".


ikesyo:
Superseded by #858.

PR #872, opened 2019-04-20
[refactoring][QuickSpec] Remove manual call of `[self testInvocations]`

ikesyo:
The method is automatically called by the XCTest runtime through `defaultTestSuite`/`testSuiteForTestCaseClass` so no need to call it manually.

There should be no behavioral change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,18 @@ branches:
 language: generic
 matrix:
   include:
+    - name: CocoaPods Lint & Danger
+      os: osx
+      osx_image: xcode10.1
+      language: ruby
+      script:
+        - PODSPEC=1 ./script/travis-script-macos
+        - bundle exec danger
     - &xcode
       name: Xcode 10.1 / Swift 4.2
       os: osx
+      env:
+        - XCODE_ACTION="build-for-testing test-without-building"
       osx_image: xcode10.1
       script:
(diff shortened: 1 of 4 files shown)


ikesyo:
Hm it seems that the code was originally added in https://github.com/Quick/Quick/pull/388. Is that still required?

ikesyo:
I don't think we should keep a workaround for xctool.

Ref: https://blog.travis-ci.com/2018-08-15-deprecating-xctool

ikesyo:
@sharplet I ran the test suite with xctool with the commit f81ff04 and it worked successfully: https://travis-ci.org/Quick/Quick/builds/522682399

So at this point we should be able to remove `[self testInvocations];` line from `+[QuickSpec initialize]`.

phatblat:
What sort of testing should be done on this in order to validate? I can try out this branch on several projects to see if anything breaks.

sharplet:
The main thing I guess would be working through all the different ways to run a test suite in Xcode & `xcodebuild`, including (off the top of my head) running the entire suite, running randomly, running in parallel, and running matching tests or test cases by name. I’m not in front of a computer right now, but there might be others in, for example, Xcode’s “Perform Action” submenu and the `xcodebuild` man page.

ikesyo:
I don't have so much time to validate those all (maybe including xctool integration?), I withdraw this.

PR #873, opened 2019-04-20
Don't call `+[QuickConfiguration initialize]` method manually

### LINKED ISSUE #853: Quick crashes after upgrading XCode 10.2
- [X] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?
I have updated to latest version of XCode 10.2 
When i switch back to previous version of XCode it just works.

Tests still executing

UITests thrown an exception when running them. 

Screenshot 2019-04-10 at 08 37 46

Screenshot 2019-04-10 at 08 37 57

### Environment

 - Quick: *2.0.0*
 - Nimble: *8.0.1*
 - Xcode Version: *10.2 (10E125)* (Open Xcode; In menubar: Xcode > About Xcode)
 - Swift Version: *XCode Default* (Open Xcode Preferences; Components > Toolchains. If none, use `Xcode Default`.)

 - Cocoapods: *1.5.3* (Use `pod --version` in Terminal)

ikesyo:
Fixes #853.

This fixes a runtime crash when a subclass of QuickSpec is subclassed and the subclass has a Swift struct property. This avoids nested calls of `+[QuickConfiguration initialize]` and `objc_getClassList`.

### DIFF
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -26,7 +26,7 @@ static QuickSpec *currentSpec = nil;
  included an expectation outside of a "it", "describe", or "context" block.
  */
 + (void)initialize {
-    [QuickConfiguration initialize];
+    [QuickConfiguration class];
 
     World *world = [World sharedWorld];
     [world performWithCurrentExampleGroup:[world rootExampleGroupForSpecClass:self] closure:^{
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -18,14 +18,12 @@ static QuickSpec *currentSpec = nil;
 #pragma mark - XCTestCase Overrides
 
 /**
- The runtime sends initialize to each class in a program just before the class, or any class
- that inherits from it, is sent its first message from within the program. QuickSpec hooks into
- this event to compile the example groups for this spec subclass.
+ QuickSpec hooks into this event to compile the example groups for this spec subclass.
 
  If an exception occurs when compiling the examples, report it to the user. Chances are they
  included an expectation outside of a "it", "describe", or "context" block.
  */
-+ (void)initialize {
++ (XCTestSuite *)defaultTestSuite {
     [QuickConfiguration class];
 
     World *world = [World sharedWorld];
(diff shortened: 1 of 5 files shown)


ikesyo:
Reopened to trigger CI.

Sources/QuickObjectiveC/QuickSpec.m
```diff
@@ -26,7 +26,7 @@ @implementation QuickSpec
  included an expectation outside of a "it", "describe", or "context" block.
  */
 + (void)initialize {
-    [QuickConfiguration initialize];
+    [QuickConfiguration class];
```
ikesyo: By accessing the class, let the Objective-C runtime call `+initialize` method of the class.

phatblat:
The new test case is great, allowed me to confirm that the change in this PR does resolve the test issue. However, I still get a crash in gzafra/QuickCrashTest#1

Screen Shot 2019-04-20 at 9 57 30 AM
Screen Shot 2019-04-20 at 9 57 27 AM

phatblat:
FYI, Travis does has a "Trigger build" button now so you shouldn't need to close/reopen the PR just to trigger a build.
Screen Shot 2019-04-20 at 10 14 29 AM

ikesyo:
> However, I still get a crash in gzafra/QuickCrashTest#1

Thanks @phatblat, I also confirmed that. So the right direction would be to avoid using `+initialize` to gather examples for each spec classes.

ikesyo:
> So the right direction would be to avoid using `+initialize` to gather examples for each spec classes.

Replacing `initialize` with `defaultTestSuite` did fix the issue.

```diff
diff --git a/Sources/QuickObjectiveC/QuickSpec.m b/Sources/QuickObjectiveC/QuickSpec.m
index 1418fec..080cf52 100644
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -25,7 +25,7 @@ static QuickSpec *currentSpec = nil;
  If an exception occurs when compiling the examples, report it to the user. Chances are they
  included an expectation outside of a "it", "describe", or "context" block.
  */
-+ (void)initialize {
++ (XCTestSuite *)defaultTestSuite {
     [QuickConfiguration class];
 
     World *world = [World sharedWorld];
@@ -48,6 +48,8 @@ static QuickSpec *currentSpec = nil;
         }
         [self testInvocations];
     }];
+
+    return [super defaultTestSuite];
 }
 
 /**
```

PR #874, opened 2019-04-21
[CI] Specifiy simulator sdk option for xcodebuild command explicitly

ikesyo:
Just a CI environment improvement.


ikesyo:
The tvOS build failure will be fixed by https://github.com/Quick/Nimble/pull/648.

jessesquires (approved):

jessesquires:
Looks like this creates CI failures.

It isn't a critical change and I don't have time to investigate, so I'm going to close.

Issue #435, opened 2015-12-04
Remove World singleton


jeffh:
This requires a lot of internal refactoring, but it would be ideal to not have a `World.sharedInstance` that's used all the time.

mkauppila:
I'd love to help with this issues :smile:  

I read through the discussion on #431 and agree about the need to move away from singleton world when considering parallelism. Maybe I'm lacking the overall picture of the codebase (and/or imagination) but how would you guys replace the `World.sharedInstance`? 

As a container of all the state, it's accessed from all around the codebase including the DSL. So to me, it seems that injecting the world instance to all the places that needs it is impossible. Also dropping the world instance altogether feels rather impractical because of its role. 

I'm just asking and wondering aloud :smile_cat:

modocache:
I actually think the `World` singleton may be the key to implementing #433. It'd be great to consider tackling this either in conjunction with that issue, or after that issue is completed.

morganchen12:
@mkauppila my initial idea was to move the example groups off of `World` and attach each `ExampleGroup` to a `QuickSpec` subclass. Then the `describe` family of functions could just be instance methods on `QuickSpec`, reading from their local instance's examples instead of being global functions pulling state from `World`. My main motivation behind this is it's okay to have `World` still around, as long as its omniscience is limited, it isn't constantly in flux while running tests, and other things aren't dependent on it--it shouldn't need to be called by other things, but it's okay for it to call other things. With this in place, the objc macros would be redefined to simply wrap the instance method (hiding away `self`), and swift would still work since `self` isn't needed to call instance methods.

I'm not 100% sure if this would work and I wanted to hammer out my ideas and get something up and running first, but I haven't put the time in yet.

jessesquires:
Hey there! 👋🏼 

I'm a new maintainer for this project and I'm trying to get the next release out ASAP and also clean up old issues and old PRs. I'm closing all issues that no longer seem relevant or are very old / stale.

This **does not** mean this issue is necessarily being rejected. If you are still interested in contributing the changes specified in this issue, please comment to request it be reopened.

We appreciate you opening this issue and acknowledge the time and effort you put in to contribute! You're awesome. 💯 However, we are all volunteers here with limited capacity working for free. Unfortunately, that means we must close out stale issues and PRs to move the project forward in a reasonable way.

We apologize for the inconvenience and appreciate your understanding! 😎 ✌🏼

PR #879, opened 2019-04-30
Run additional test suites in a separate World instance (only with Xcode for now)

ikesyo:
This is a step to correctly test #680 in SwiftPM. And also this is a bit related to #435 I think.


ikesyo:
I'm merging this. Please feel free to leave a comment if there is any concern.

PR #887, opened 2019-05-09
Add (before|after)EachAutoreleasepool hooks

VojtaStavik:
This PR wraps the example execution into `autoreleasepool` and adds two more hooks: `beforeEachAutoreleasepool` and `afterEachAutoreleasepool`.

## Motivation
The main motivations for adding `afterEachAutoreleasepool` is to provide an easy way to test memory leaks. Example:
```swift
var viewController: UIViewController!
weak var weakViewController: UIViewController?

beforeEach { 
  viewController = .init() 
  weakViewController = viewController
}

describe("something") { ... }

afterEach { 
  // Remove the strong reference to the controller
  viewController = nil
}

afterEachAutoreleasepool { 
  // Check the view controller is no longer in the memory
  expect(weakViewController).to(beNil())
}
```

It's possible to test _certain_ objects for memory leaks without wrapping them into `autoreleasepool` with the following approach:
```swift
afterEach { 
  weakObject = object
  object = nil
  expect(weakObject).to(beNil())
}
```
However, this **doesn't work for `UIViewController`s**! 
_(see: https://stackoverflow.com/questions/45890020/uiviewcontroller-variables-persist-after-deallocation/45910208)_ 

`UIViewController` is such a fundamental class in iOS, I think it deserves to have a special treatment like this. Especially, when it's so susceptible to memory leaks.

`beforeEachAutoreleasepool` was added mainly to maintain the `before`/`after` symmetry with other hooks.

## Alternative
This PR is a concrete implementation of the use case suggested by @pcantrell at https://github.com/Quick/Quick/pull/820. I find the solution in this PR easier to use and less error-prone.

## Checklist
_What behavior was changed?_
None.

_What code was refactored / updated to support this change?_
The example execution was wrapped into `autoreleaspool`. This shouldn't have any effect on the functionality.

_What issues are related to this PR? Or why was this change introduced?_
I couldn't find an easy way to test whether my view controllers leak memory.

 - [x] Does this have tests?

### DIFF
--- a/Quick.xcodeproj/project.pbxproj
+++ b/Quick.xcodeproj/project.pbxproj
@@ -24,6 +24,9 @@
 		1B77503622844254002C6143 /* AutoreleasepoolTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B77503522844254002C6143 /* AutoreleasepoolTests.swift */; };
 		1B77503722844254002C6143 /* AutoreleasepoolTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B77503522844254002C6143 /* AutoreleasepoolTests.swift */; };
 		1B77503822844254002C6143 /* AutoreleasepoolTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B77503522844254002C6143 /* AutoreleasepoolTests.swift */; };
+		1BDDC5FC228461500060F8E2 /* AutoreleasepoolTests+ObjC.m in Sources */ = {isa = PBXBuildFile; fileRef = 1BDDC5FA228461380060F8E2 /* AutoreleasepoolTests+ObjC.m */; };
+		1BDDC5FD228461510060F8E2 /* AutoreleasepoolTests+ObjC.m in Sources */ = {isa = PBXBuildFile; fileRef = 1BDDC5FA228461380060F8E2 /* AutoreleasepoolTests+ObjC.m */; };
+		1BDDC5FE228461520060F8E2 /* AutoreleasepoolTests+ObjC.m in Sources */ = {isa = PBXBuildFile; fileRef = 1BDDC5FA228461380060F8E2 /* AutoreleasepoolTests+ObjC.m */; };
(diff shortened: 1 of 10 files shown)


ikesyo:
Thanks for the PR! However, at a first glance, I think that this is too specific for your usecase and will be a bit difficult to understand the meaning of `afterEachAutoreleasepool`: we can't see any `autoreleasepool` existence from spec (example and hooks) definitions, so that may be confusing.

And also wrapping all example (and hooks) executions into `autoreleasepool` seems too aggressive and not ideal. There is no way to opt-out it. I assume normal XCTest does not wrap each test method execution into the pool.

So I'm rather for #820 as a more general, abstract mechanism.

ikesyo:
I want to hear other opinions from the core team.

@Quick/core

vst-ableton:
> we can't see any `autoreleasepool` existence from spec (example and hooks) definitions, so that may be confusing.
Fair point 👍 

>  I assume normal XCTest does not wrap each test method execution into the pool.
I actually think this is exactly what's happening. That's at least the observation I made when trying to understand what's happening.

VojtaStavik:
> we can't see any `autoreleasepool` existence from spec (example and hooks) definitions, so that may be confusing.

Fair point 👍 

>  I assume normal XCTest does not wrap each test method execution into the pool.

I actually think this is exactly what's happening. That's at least the observation I made when trying to understand what's happening.

pcantrell:
FWIW, I do plan to get #820 off my back burner and merged eventually!

With #820, the `afterEachAutoreleasepool` example at the of the OP is:

```swift
aroundEach {
    autoreleasepool { $0() }
    expect(weakViewController).to(beNil())
}
```

Whether that deserves additional sugar, I’m not sure.

VojtaStavik:
The longer I think about it the more I also believe #820 is a better solution for the problem.

It seems that what's missing there are tests and Objective-C support, is that correct? I can help with that if you're interested @pcantrell. I'm not attached to this PR. I would just like to have a solution to my problem :)

pcantrell:
There were some corner cases I wanted to make sure to get in the tests. I’ll try to finish up that work later this week or early next. @VojtaStavik, I’d love help putting together the Obj-C API if you want to chip in on that!

PR #901, opened 2019-07-15
Fix empty 'Selected tests' suites

### LINKED ISSUE #891: Trying to run a specific spec does not run any tests
What did you do?

Running a spec with the play button in Xcode results in no tests being run. I raised this earlier in #889 but it was immediately closed. 

Upon further investigation I've discovered the fix in #873 is causing this issue. Specifically the `defaultTestSuite` is not called when running a specific spec.

What did you expect to happen?

The specific spec to be run

What actually happened instead?

No tests were run

Environment

Quick: latest from master 32804c3
Nimble: 8.0.1
Xcode Version: 10.2.1 (Open Xcode; In menubar: Xcode > About Xcode)
Swift Version: Xcode Default (Open Xcode Preferences; Components > Toolchains. If none, use Xcode Default.)

Cocoapods: 1.6.1

VojtaStavik:
This PR fixes the regression introduced in a3ece4f. Closes #891 

The `defaultTestSuite` method wasn't called when the 'Selected tests' type of test suite was executed. This caused the `spec` method not being invoked and the examples for the given class were never built.

This PR changes the behaviour and introduces a new `buildExamplesIfNeeded` method. This method can be called safely multiple times. We call it from both the `defaultTestSuite` and `testInvocations` methods. This ensures the examples are built properly no matter which method is called (or in which order).

 - [x] Does this have tests?

### DIFF
--- a/Quick.xcodeproj/project.pbxproj
+++ b/Quick.xcodeproj/project.pbxproj
@@ -119,6 +119,9 @@
 		64076D231D6D7E6B00E2B499 /* AfterSuiteTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 64076D1D1D6D7D0B00E2B499 /* AfterSuiteTests.swift */; };
 		64076D261D6D80B500E2B499 /* AfterSuiteTests+ObjC.m in Sources */ = {isa = PBXBuildFile; fileRef = 64076D241D6D80B500E2B499 /* AfterSuiteTests+ObjC.m */; };
 		64076D271D6D80B500E2B499 /* AfterSuiteTests+ObjC.m in Sources */ = {isa = PBXBuildFile; fileRef = 64076D241D6D80B500E2B499 /* AfterSuiteTests+ObjC.m */; };
+		79FE27DF22DC955400BA013D /* QuickSpec_SelectedTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FE27DE22DC955400BA013D /* QuickSpec_SelectedTests.swift */; };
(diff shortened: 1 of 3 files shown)


VojtaStavik:
Since the bug happens only when the tests are run from the IDE, we should restrict the tests in 
9722799 only for non-Darwin platforms?

VojtaStavik:
@Quick/core can someone please take a look at this PR? Thank you!

ikesyo:
It seems that by this fix, a crash occurs when using focused tests 😕 

```
2019-09-12 08:09:24.170519+0900 xctest[11754:14170830] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An exception occurred when building Quick's example groups.
Some possible reasons this might happen include:

- An 'expect(...).to' expectation was evaluated outside of an 'it', 'context', or 'describe' block
- 'sharedExamples' was called twice with the same name
- 'itBehavesLike' was called with a name that is not registered as a shared example

Here's the original exception: 'NSInternalInconsistencyException', reason: 'A shared example named 'combineLatest examples' has already been registered.', userInfo: '(null)''
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff3d62d2fd __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff67d0ba17 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff3d62d12f +[NSException raise:format:] + 201
	3   Quick                               0x0000000103642857 __34+[QuickSpec buildExamplesIfNeeded]_block_invo
libc++abi.dylib: terminating with uncaught exception of type NSException
```

VojtaStavik:
This crash looks familiar. Let me double check again. it’s been a while :)

PR #905, opened 2019-07-22
Add documentation for `Behavior<Context>`

VojtaStavik:
I use Quick for several years and recently I accidentally discovered `Behavior`. I really love the feature! I realized that I didn't know about it because it's not mentioned anywhere in the documentation.

This PR adds another option to the `Reducing Test Boilerplate` documentation presenting an alternative solution using `Behavior<Context>`.

### DIFF
--- a/Documentation/en-us/README.md
+++ b/Documentation/en-us/README.md
@@ -29,7 +29,7 @@ testing, consider reading them in the order they're introduced below:
   Read this to learn more about testing code that uses the AppKit and UIKit frameworks.
 - **[Testing with test doubles](TestUsingTestDoubles.md)**:
   Read this to learn what test doubles are and how to use them.
-- **[Reducing Test Boilerplate with Behaviors and Shared Assertions](SharedExamples.md)**:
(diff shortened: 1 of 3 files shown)


Documentation/en-us/README.md
```diff
@@ -29,7 +29,7 @@ testing, consider reading them in the order they're introduced below:
   Read this to learn more about testing code that uses the AppKit and UIKit frameworks.
 - **[Testing with test doubles](TestUsingTestDoubles.md)**:
   Read this to learn what test doubles are and how to use them.
-- **[Reducing Test Boilerplate with Shared Assertions](SharedExamples.md)**:
+- **[Reducing Test Boilerplate with Behaviors and Shared Assertions](SharedExamples.md)**:
```
ikesyo: I think we can define "Shared Assertions" as a concept including both `Behavior` and `sharedExample` implementation, so I'd like to leave this as is.

ikesyo:
Thanks for the documentation improvement!

PR #906, opened 2019-07-22
Make `name` variable of `Behavior` open

VojtaStavik:
The PR is a byproduct of [my article](https://vojtastavik.com/2019/07/22/advanced-testing-using-behavior-in-quick/) about using `Behavior` in Quick.

I believe the `name` variable of `Behavior` should be `open` so it's possible to override it and provide custom name, if needed.

### DIFF
--- a/Sources/Quick/Behavior.swift
+++ b/Sources/Quick/Behavior.swift
@@ -4,7 +4,11 @@
 
 open class Behavior<Context> {
 
-    public static var name: String { return String(describing: self) }
+    /**
+     Override this variable if you want to provide custom name for this example group.
+    */
+    open class var name: String { return String(describing: self) }
+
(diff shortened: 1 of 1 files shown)


Sources/Quick/Behavior.swift
```diff
@@ -4,7 +4,11 @@
 
 open class Behavior<Context> {
 
-    public static var name: String { return String(describing: self) }
+    /**
+     Override this variable if you want to provide custom name for this example group.
+    */
+    open class var name: String { return String(describing: self) }
```
ikesyo: This is exactly what I thought before and is not a breaking change, so it's safe to merge.

https://github.com/Quick/Quick/pull/821#discussion_r220907011

PR #922, opened 2019-09-16
[refactoring] Build examples only for classes without root example group

VojtaStavik:
This commit fixes the crash reported in https://github.com/Quick/Quick/pull/901#issuecomment-530600416 .

### DIFF
--- a/Sources/Quick/World.swift
+++ b/Sources/Quick/World.swift
@@ -110,6 +110,17 @@ final internal class World: _WorldBase {
     }
 
     /**
+     Returns `true` if the root example group for the given spec class has been already initialized.
+
+     - parameter specClass: The QuickSpec class for which is checked for the existing root example group.
+     - returns: Whether the root example group for the given spec class has been already initialized or not.
+     */
+    internal func isRootExampleGroupInitialized(forSpecClass specClass: QuickSpec.Type) -> Bool {
+        let name = String(describing: specClass)
+        return specs.keys.contains(name)
+    }
+
+    /**
         Returns an internally constructed root example group for the given
         QuickSpec class.
 
--- a/Sources/QuickObjectiveC/QuickSpec.m
+++ b/Sources/QuickObjectiveC/QuickSpec.m
@@ -84,13 +84,12 @@ static QuickSpec *currentSpec = nil;
     [QuickConfiguration class];
     World *world = [World sharedWorld];
 
-    ExampleGroup *rootExampleGroup = [world rootExampleGroupForSpecClass:self];
-
-    if ([rootExampleGroup examples].count > 0) {
+    if ([world isRootExampleGroupInitializedForSpecClass:[self class]]) {
         // The examples fot this subclass have been already built. Skipping.
         return;
     }
 
+    ExampleGroup *rootExampleGroup = [world rootExampleGroupForSpecClass:[self class]];
     [world performWithCurrentExampleGroup:rootExampleGroup closure:^{
         QuickSpec *spec = [self new];
 


VojtaStavik:
@ikesyo Oh, I see you have fixed this already in https://github.com/Quick/Quick/pull/917 . I wasn't sure about the solution and thought I'd do something more proper. That's why I closed the PR.

ikesyo:
@VojtaStavik I'm sorry I already merged #917 (which is based on #916) and released [v2.2.0](https://github.com/Quick/Quick/releases/tag/v2.2.0). I will accept this PR as a refactoring if you resolve conflicts with the current `master`.

VojtaStavik:
No worries. I didn't know you wanted to release a new version and this was a blocker. I would have fixed it immediately. Let me know next time ;-)

VojtaStavik:
Technically, we can revert the commits making `[rootExampleGroup examples]`  possible. However, I guess it doesn't hurt to have `examples` accessible from Objective-C, too.

Sources/QuickObjectiveC/QuickSpec.m
```diff
@@ -84,13 +84,12 @@ + (void)buildExamplesIfNeeded {
...
 
-    ExampleGroup *rootExampleGroup = [world rootExampleGroupForSpecClass:self];
-
-    if ([rootExampleGroup examples].count > 0) {
+    if ([world isRootExampleGroupInitializedForSpecClass:[self class]]) {
         // The examples fot this subclass have been already built. Skipping.
         return;
     }
 
+    ExampleGroup* rootExampleGroup = [world rootExampleGroupForSpecClass:[self class]];
```
ikesyo: `ExampleGroup* rootExampleGroup` should be `ExampleGroup *rootExampleGroup` for consistency.

VojtaStavik: Sorry, old habit. Fixed.

Sources/Quick/World.swift
```diff
@@ -109,6 +109,18 @@ final internal class World: _WorldBase {
         isConfigurationFinalized = true
     }
 
+    /**
+     Returns `true` if the root example group from the given spec class has been already
+     initialized.
+
+     - parameter specClass: The QuickSpec class for which is checked for the existing root
+     example group.
+     */
```
ikesyo: I'd like to tweak the doc comment block as:

```swift
    /**
     Returns `true` if the root example group for the given spec class has been already initialized.

     - parameter specClass: The QuickSpec class for which is checked for the existing root example group.
     - returns: Whether the root example group for the given spec class has been already initialized or not.
     */
```

(adjusting line length and adding `returns`)

VojtaStavik: Hm, I tried to match the line length of the doc for `func rootExampleGroup(forSpecClass specClass: QuickSpec.Type) -> ExampleGroup` 🤷‍♂ 

I'll add the `returns` doc; it seemed to be a bit redundant to me.

VojtaStavik: I think it's checked by default, isn't it? I see it checked.

Screen Shot 2019-09-17 at 08 32 08

ikesyo: Hm then it's weird that I couldn't push to your fork. 😕

ikesyo: > Hm, I tried to match the line length of the doc for `func rootExampleGroup(forSpecClass specClass: QuickSpec.Type) -> ExampleGroup` 🤷‍♂ 

I'll add SwiftLint line length rule configuration with 120 (or more).

PR #953, opened 2020-03-18
Update the Objective-C QuickSpec template to use modules (semantic import)

KeithMorning:
help you if you're not sure:

 - What behavior was changed?
I change the OC Quick Template, from this issue https://github.com/Quick/Nimble/issues/642
The  Quick Spec Class.xctemplate OC file ___FILEBASENAME___.m
https://github.com/Quick/Nimble/issues/642

 - [x] Does this have tests?
 - [] Does this break the public API (Requires major version bump)?
 - [] Is this a new feature (Requires minor version bump)?

### DIFF
--- a/Quick Templates/Quick Spec Class.xctemplate/Objective-C/___FILEBASENAME___.m
+++ b/Quick Templates/Quick Spec Class.xctemplate/Objective-C/___FILEBASENAME___.m
@@ -6,8 +6,8 @@
 //  ___COPYRIGHT___
 //
 
-@import Nimble;
 @import Quick;
+@import Nimble;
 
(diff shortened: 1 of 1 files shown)


Quick Templates/Quick Spec Class.xctemplate/Objective-C/___FILEBASENAME___.m
```diff
@@ -6,8 +6,8 @@
 //  ___COPYRIGHT___
 //
 
-#import <Quick/Quick.h>
-#import <Nimble/Nimble.h>
+@import Nimble;
+@import Quick;
```
ikesyo:
```suggestion
@import Quick;
@import Nimble;
```

Issue #886, opened 2019-05-08
fdescribe, fit, fcontext stopped working.


fousa:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

I want to use `fdescribe` to make sure just on test runs.

When I use `github "Quick/Quick"` is works for an iOS test target. On tvOS I have to downgrade to `github "Quick/Quick" == 2.0`.

But when I switch to `github "Quick/Quick" "master"` it seems to be broken on both platforms.

### What did you expect to happen?

Only the tests in this file run.

### What actually happened instead?

All the tests run until I hit the tests in this file and Xcode stops the test suite.

### Environment

List the software versions you're using:

 - Quick: *2.1.0*
 - Nimble: *commit 0edb48d936f31f4ce657047884b77b6e3b0bef4b*
 - Xcode Version: *10.2.1 (10E1001)*
 - Swift Version: *Xcode Default*

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - Carthage: *0.32.0* (Use `carthage version` in Terminal)

cooksimo:
I have the same issue, also trying to run a specific spec via the button in Xcode results in no tests being run

ikesyo:
[The test](https://github.com/Quick/Quick/blob/master/Tests/QuickTests/QuickFocusedTests/FocusedTests.swift) in our test suite is passing in all supported platforms (macOS, iOS, tvOS, SwiftPM on macOS, SwiftPM on Linux). It would be great if you could provide a reproducible case/project.

ikesyo:
#873 may be related.

cooksimo:
I am seeing the same issue, will see if I can come up with a sample project tomorrow.

fousa:
I tried using Quick from the master branch (after the #873 fix was merged) and the probleem seems to persist. I'll see if I can upload a sample project this weekend.

ikesyo:
I didn't say #873 may be fix for this, but said #873 may be the cause of this.

fousa:
Ah alright, my bad.

pimnijman:
Any updates on this? I'm running into this issue right now and since the last comment is from May I was hoping you found a fix.

mikkoseppa:
Okay so I spent some time trying to figure this out.

**Example setup with three specs, single test focused using `fit`:**
- QuickSpec1
  - it("TestCase1")
[...]

PR #962, opened 2020-04-07
[WIP] Fix focus behavior

marcelofabri:
help you if you're not sure:

 - What behavior was changed?

The `f`it & friends behavior was changed in 2.1.0, as described in [this comment](https://github.com/Quick/Quick/issues/886#issuecomment-569627450). 

This PR adds a test case demonstrating that. Would love some pointers on how to fix it.

#886

 - [x] Does this have tests?


ikesyo:
Superseded by #980. Thanks for looking into this though!

PR #980, opened 2020-05-27
Fix focus behavior

### LINKED ISSUE #886: fdescribe, fit, fcontext stopped working.

ikesyo:
Let's configure all example groups of QuickSpec subclasses at first using XCTestObservation's `testBundleWillStart(_:)`.

refs:
- #873 
- Fixes #886


ikesyo:
I confirmed this works by myself.

PR #993, opened 2020-06-02
Update CI config for 2.x-branch

ikesyo:
ref: #991

### DIFF
--- a/Sources/Quick/QuickSpec.swift
+++ b/Sources/Quick/QuickSpec.swift
@@ -123,14 +123,14 @@ open class QuickSpec: QuickSpecBase {
 
     // MARK: Delegation to `QuickSpec.current`.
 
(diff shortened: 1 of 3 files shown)


Sources/Quick/QuickSpec.swift
```diff
@@ -129,8 +129,8 @@ open class QuickSpec: QuickSpecBase {
         atLine lineNumber: Int,
         expected: Bool
     ) {
-        guard self === Self.current else {
-            Self.current.recordFailure(
+        guard self === QuickSpec.current else {
+            QuickSpec.current.recordFailure(
```
ikesyo, line 133: refs:
- #991 
- https://github.com/apple/swift-evolution/blob/master/proposals/0068-universal-self.md

PR #1006, opened 2020-08-25
add cxx setting CLANG_ENABLE_MODULES

### LINKED ISSUE #997: Compile error with the SwiftPM-generated Xcode project on v3.0.0

### What did you do?

Add Quick to SPM Package.swift as a dependency and try to run tests.

Test target would compile and run. 

Test target failed to compile. Got the following error: 

```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_XCTestCase", referenced from:
      _OBJC_CLASS_$__QuickSpecBase in QuickSpecBase.o
  "_OBJC_METACLASS_$_XCTestCase", referenced from:
      _OBJC_METACLASS_$__QuickSpecBase in QuickSpecBase.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Environment

 - Quick: 3.0.0
 - Xcode Version: 11.5
 - Swift Version: Xcode Default

 - Swift Package Manager: 5.2.0

Just ran `swift package init` and added Quick as a dependency to the package and test target. 

```
// swift-tools-version:5.2
import PackageDescription

let package = Package(
    name: "DemoSPM",
    products: [
        .library(name: "DemoSPM", targets: ["DemoSPM"]),
    ],
    dependencies: [
        .package(url: "https://github.com/Quick/Quick", from: "3.0.0"),
    ],
    targets: [
        .target(name: "DemoSPM", dependencies: []),
        .testTarget(name: "DemoSPMTests", dependencies: [
            .target(name: "DemoSPM"),
            .product(name: "Quick", package: "Quick"),
        ]),
    ]
)
```

makleso6:
help you if you're not sure:

 - What behavior was changed? Add cxx setting in Package.swift
 - What code was refactored / updated to support this change? - Nothing
 - What issues are related to this PR? Or why was this change introduced? Try to fix #997

 - [ - ] Does this have tests? 
 - [ - ] Does this have documentation?
 - [ - ] Does this break the public API (Requires major version bump)?
 - [ - ] Is this a new feature (Requires minor version bump)?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,7 +1,10 @@
 // swift-tools-version:5.2
 
 import PackageDescription
+let cxxSettings: [CXXSetting] = [
+    .define("CLANG_ENABLE_MODULES", to: "YES"),
 
+]
 let package = Package(
     name: "Quick",
     platforms: [
@@ -35,7 +38,7 @@ let package = Package(
         ]
 #if os(macOS)
         targets.append(contentsOf: [
-            .target(name: "QuickObjCRuntime", dependencies: []),
+            .target(name: "QuickObjCRuntime", dependencies: [], cxxSettings: cxxSettings),
             .target(name: "Quick", dependencies: [ "QuickObjCRuntime" ]),
         ])
 #else


ikesyo:
I think that having an umbrella header would be preferable to this approach. I'll tackle it.

ref: https://forums.swift.org/t/recent-swiftpm-features-and-enhancements/13807

> Embedable Xcode projects:
> For various complex reasons, C language targets are not mapped to "real" framework targets in the generated Xcode project. In practice, this means that the "defines module" and "enable modules" build settings are turned off in the generated framework target. This causes issues when trying to use the generated project as a subproject.
> 
> The issue is resolved for C language targets that have an umbrella header. To create an umbrella header, place a file named <targetname>.h inside the include/ directory of that target and include all the public headers in the umbrella header.

ikesyo:
Superseded by #1023.

Issue #1007, opened 2020-08-31
Package file specifies incorrect macOS version


danielsaidi:
[ x] I have read CONTRIBUTING and have done my best to follow them.

## What did you do?

Opened a project that requires Nimble in Xcode 12 and tried to build it.

## What did you expect to happen?

The project should build.

## What actually happened instead?

The project doesn't build due to a Nimble build error.

## Environment

List the software versions you're using:

* Quick: 3.0.0
* Nimble: 8.1.1
* Xcode Version: 12 beta
* Swift Version: 5.3

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

* Swift Package Manager 5.3.0 (swiftpm-5.3.0)

## Solution

The Package file specifies

```
platforms: [
    .macOS(.v10_10), .iOS(.v8), .tvOS(.v9)
],
```

but `recordFailure` is only available in 10.13 or newer.

I guess the package files should be updated to reflect this properly.

![](https://user-images.githubusercontent.com/429927/91754809-54644500-ebca-11ea-8fd3-eb94db9239e9.png)

ikesyo:
Duplicate of #995.

PR #1008, opened 2020-08-31
Update macOS version in Package file

danielsaidi:
help you if you're not sure:

 - What behavior was changed?

The package file specified the incorrect macOS version, which results in Quick not building on macOS, since `recordFailure` requires 10.13 or later. 

I only updated the macOS version in the package file.

https://github.com/Quick/Quick/issues/1007

 - [ ] Does this have tests?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 let package = Package(
     name: "Quick",
     platforms: [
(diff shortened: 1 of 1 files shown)


ikesyo:
Looks like this is not an issue anymore at least on Xcode 12.0 beta 6. ☺️

PR #1028, opened 2020-11-15
Make NSBundle.currentTestBundle() public

rbaumbach:
I've had various scenarios where I'd like to get the spec bundle in my projects.  I usually create an extension to do this, but I noticed this extension is already present in `Quick`, but is marked `internal`.

The only change in this PR that I marked this static method as `public`, so that others can use it if they desire.


rbaumbach:
I'm following up on this PR.  @ikesyo are you an appropriate contributor to contact about getting this PR merged?

ikesyo:
Thanks for the pull request @rbaumbach. However this is an internal implementation detail and is not  meant to be used as a public API. I'm in favor of reducing public API surface as much as possible.

PR #1061, opened 2021-04-09
Fix `afterSuite` blocks not running when having excluded tests

### LINKED ISSUE #809: afterSuite blocks do not run if there are excluded tests
- [x ] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

When there are excluded examples (using 'xcontext,' 'xit,' etc), the count of examples used by Example.run() (_world.includedExampleCount_) is incorrect (it's too large), which causes the logic that decides whether to call _world.suiteHooks.executeAfters()_ to fail (because the _numberOfExamplesRun_ count never reaches the _numberOfIncludedExamples_).

The calculation of _numberOfIncludedExamples_ was apparently introduced to correct https://github.com/Quick/Quick/issues/482, because the count was not taking focused tests into account.  This calculation *also* needs to take excluded tests into account or the count will still be wrong.

For example, if you have 100 examples, but 3 of them are excluded with 'xit', the check to run the "afters" will be watching for example # 100, but it only gets to 97, so the afters never run, and the test run ends.

I have a fix for the issue, but haven't had a chance to run all the tests yet.  I'll open a PR when I get a chance to do that.  

the afterSuite block to be executed after all the tests ran

the afterSuite block did not run

### Environment

 - Quick: 1.2.0 (but the behavior is unchanged in master)
 - Nimble: N/A
 - Xcode Version: 9.3.1 (9E501) 
 - Swift Version: 4.1 

 - Cocoapods: 1.5.2

revolter:
help you if you're not sure:

 - What behavior was changed?
The logic used to determine whether the `afterSuite` blocks should be called or not.
It also adds missing gems required for the `bundle exec rake` command mentioned in the `CONTRIBUTING.md` file.

`World.swift`

Fixes #809.
Related to #488.

 - [x] Does this have tests?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,4 @@ source "https://rubygems.org"
 gem 'cocoapods', '~> 1.10'
 gem 'danger'
 gem 'rake'
+gem 'xcpretty'
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,4 @@ source "https://rubygems.org"
 
 gem 'cocoapods', '~> 1.10'
 gem 'danger'
+gem 'rake'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -111,6 +111,7 @@ GEM
     rake (13.0.3)
     rchardet (1.8.0)
     rexml (3.2.4)
+    rouge (2.0.7)
     ruby-macho (1.4.0)
     ruby2_keywords (0.0.4)
     sawyer (0.8.2)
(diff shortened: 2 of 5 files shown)


jessesquires:
thanks @revolter ! 💯

revolter:
Wow, I forgot about this. Thanks for merging!

Issue #1084, opened 2021-06-08
Feature request: async support


NachoSoto:
Just like it was added in [XCTest](https://twitter.com/arekholko/status/1402031047014035458?s=21), it would be great to be able to do this:

```swift
it("") {
  expect(await f()) == 1
}
```

hernangonzalez:
Really looking forward to this :)

gshahbazian:
It would be awesome to get built in async/await support, but until that is ready I've adopted [this linux testing example](https://www.swiftbysundell.com/articles/unit-testing-code-that-uses-async-await/#linux-compatibility) for use with Quick:

```swift
public func asyncIt(_ description: String, flags: FilterFlags = [:], file: FileString = #file, line: UInt = #line, closure: @MainActor @escaping () async throws -> Void) {
    it(description, flags: flags, file: file, line: line) {
        var thrownError: Error?
        let errorHandler = { thrownError = $0 }
        let expectation = QuickSpec.current.expectation(description: description)

        Task {
            do {
                try await closure()
            } catch {
                errorHandler(error)
            }

            expectation.fulfill()
        }

        QuickSpec.current.wait(for: [expectation], timeout: 60)

        if let error = thrownError {
            XCTFail("Async error thrown: \(error)")
        }
    }
}

public func asyncBeforeEach(_ closure: @MainActor @escaping (ExampleMetadata) async -> Void) {
    beforeEach({ exampleMetadata in
        let expectation = QuickSpec.current.expectation(description: "asyncBeforeEach")
        Task {
            await closure(exampleMetadata)
            expectation.fulfill()
        }
        QuickSpec.current.wait(for: [expectation], timeout: 60)
    })
}

public func asyncAfterEach(_ closure: @MainActor @escaping (ExampleMetadata) async -> Void) {
    afterEach({ exampleMetadata in
        let expectation = QuickSpec.current.expectation(description: "asyncAfterEach")
        Task {
            await closure(exampleMetadata)
            expectation.fulfill()
        }
        QuickSpec.current.wait(for: [expectation], timeout: 60)
    })
}
```

Does anyone see any potential issues with these extensions?

LuAndreCast:
any status on this? 🙏🏽

zbencz3:
> Does anyone see any potential issues with these extensions?

@gshahbazian 
`FilterFlags` are internal in Quick 5.

bvirlet:
@gshahbazian did you implement `async` variants of `expect`?

[...]

PR #1085, opened 2021-06-08
Async tests proof of concept

NachoSoto:
Inspired by https://twitter.com/arekholko/status/1402031047014035458?s=21.
Proof of concept for #1084.

### Issues:
- This does not yet build.
- Not backwards compatible. I'm not sure how easy it will be to have async and regular variants of all APIs.
- `Example.run` implementation does not work. I made it block using `DispatchGroup` because the method can't be `async`, as it's used from `Objective-C`. I haven't looked too much into why this implementation is deadlocking, but I suspect this won't be the right way of doing this. Ideally we just hook into `XCTest`s ability to run `async` functions.
- As others have pointed out, `async`/`await` leads to an explosion of call-sites needing to become async. It actually got me thinking how useful it would be for `async` to work like `throws`: just like a method only throws if a called closure does, it would be great if `async` could propagate in the same manner.
- `Nimble` will need support for this too (for example, `Expression`).

### DIFF
--- a/Sources/Quick/DSL/DSL.swift
+++ b/Sources/Quick/DSL/DSL.swift
@@ -126,7 +126,7 @@ public func afterEach(_ closure: @escaping AfterExampleWithMetadataClosure) {
     - parameter file: The absolute path to the file containing the example. A sensible default is provided.
     - parameter line: The line containing the example. A sensible default is provided.
 */
-public func it(_ description: String, file: FileString = #file, line: UInt = #line, closure: @escaping () throws -> Void) {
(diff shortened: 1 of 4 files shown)


jessesquires:
Hey @NachoSoto I'm inheriting this project and doing some clean up. 

I'm going to close this due to it being "proof of concept" / non-critical. But feel free to re-open if you are still interested in this! 😄 ✌🏼

NachoSoto:
No problem!

Unless somebody can think of some clean way, I think this won't be worth doing without `reasync`.

Issue #1025, opened 2020-10-19
Sporadic crashes after updating to v2.2.1 in QuickSpec.enumerateSubclasses


lordzsolt:
- [X] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/master/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Tried running the unit tests on CI.
Also happens locally, when running the tests for the first time.

### What did you expect to happen?

Tests run.

### What actually happened instead?

Crash with `EXC_BAD_ACCESS (SIGSEGV)`

### Environment

 - Quick: *3.0.0*
 - Nimble: *9.0.0*
 - Xcode Version: *Any. Xcode 12.0.1 (12A7300), 12.2 beta (12B5018i), Xcode 11.7 (11E801a)*
 - Swift Version: *Xcode Default*

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - CocoaPods: *1.9.3*

### Project that demonstrates the issue

Cannot reproduce it reliably. Happens in Private codebase.

Here's a crash log with the relevant stack trace:

```
Process:               Redacted [5926]
Path:                  /Users/USER/Library/Developer/CoreSimulator/Devices/57F35BB1-5B28-40C1-833F-8F4326FE73DF/data/Containers/Bundle/Application/A5A93B63-64B3-4CAA-93EB-A974A77BD415/Redacted.app/Redacted
Identifier:            Redacted
Version:               8.3.0 (1156)
Code Type:             X86-64 (Native)
Parent Process:        launchd_sim [862]
Responsible:           SimulatorTrampoline [807]
User ID:               501

Date/Time:             2020-10-19 13:01:09.867 +0000
OS Version:            Mac OS X 10.15.5 (19F101)
Report Version:        12
Anonymous UUID:        4763FA26-E06C-409D-9FAC-0174438B0AC7


Time Awake Since Boot: 600 seconds

System Integrity Protection: disabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x00000001000007ac
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [5926]

VM Regions Near 0x1000007ac:
--> 
[...]

PR #1086, opened 2021-06-22
Fixing bug #1025

thekushpatel:
This PR fixes bug #1025. 

The `enumerateSubclasses` function in `QuickConfiguration` and `QuickTestObservation` uses the `is` keyword to determine if a class is an instance of `QuickConfiguration` or `QuickSpec`. For some reason, this does not work with all classes and causes a [crash](https://github.com/Quick/Quick/issues/1025#issuecomment-717888950) when running tests.

The solution implemented in this PR was proposed by @iDevPro.

More information on the issue can be found here #1025.

### DIFF
--- a/Sources/Quick/Configuration/QuickConfiguration.swift
+++ b/Sources/Quick/Configuration/QuickConfiguration.swift
@@ -37,6 +37,7 @@ extension QuickConfiguration {
             guard
                 let subclass = classes[Int(index)],
                 let superclass = class_getSuperclass(subclass),
+                type(of: superclass) == QuickConfiguration.Type.self,
                 superclass is QuickConfiguration.Type
                 else { continue }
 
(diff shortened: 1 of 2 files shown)


Sources/Quick/Configuration/QuickConfiguration.swift
```diff
@@ -37,6 +37,7 @@ extension QuickConfiguration {
             guard
                 let subclass = classes[Int(index)],
                 let superclass = class_getSuperclass(subclass),
+                type(of: superclass) == QuickConfiguration.Type.self,
```
ikesyo, line 40: Unfortunately this doesn't work for cases of indirect subclasses:

```swift
class BaseConfig: QuickConfiguration { ... }
class FooConfig: BaseConfig { ... }
```

That is the reason why we use `superclass is QuickConfiguration.Type` here.

ref: #975

PR #1088, opened 2021-07-22
Excluding Info.plist files from Swift Package

heyzooi:
help you if you're not sure:

 - What behavior was changed?
No

 No

Warnings are being generated by Xcode when the Quick was used using SPM

 - [ ] Does this have tests?


ikesyo (approved):
👍 

ref: https://github.com/Quick/Nimble/pull/878

PR #1096, opened 2021-08-29
Use Objective-C runtime's `isSubclass(of:)` API for checking superclass

ikesyo:
A possible fix for #1025.


krris:
Unfortunately this doesn't help. I can still reproduce the crash.

ikesyo:
@krris Thanks for the confirmation. Could you please take a look at #1103 as well?

PR #1098, opened 2021-09-01
Report skipped tests to Xcode

### LINKED ISSUE #1093: Feature request: Mark skipped tests as skipped in Xcode.
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/main/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Compare what happens when skipping a test in native `XCTest`:

```swift
import XCTest

class SampleTest: XCTestCase {
	func test_this_test_should_run() {
		
	}
	
	func test_this_test_is_skipped() throws {
		throw XCTSkip("Skipping this test")
	}
}
```

Versus what happens when you skip a test using Quick (e.g. using `xit`)::

```swift
import Quick

class SampleSpec: QuickSpec {
	override func spec() {
		it("this test should run") {

		}
		xit("this test is skipped") {
			fatalError("This should never be called")
		}
	}
}
```

The Xcode sidebar should have shown the test as skipped, as it does when using `XCTest`:

image

And the skipped line could have been highlighted, as it does when using `XCTest`:

Screen Shot 2021-08-25 at 12 36 52 PM

The Quick framework just pretends the test doesn't exist at all, rather than explicitly showing that it *does* exist, but was skipped:

image

And no highlights were made that the test was skipped:

Screen Shot 2021-08-25 at 12 38 12 PM

### Environment

- Quick: *4.0.0*
- Nimble: *9.2.0*
- Xcode Version: *13.0 (13A5201i)*
- Swift Version: *5.5*

- Swift Package Manager *5.5.0*

amomchilov:
help you if you're not sure:

 - What behavior was changed?
 
     When running on macOS, skipped tests will be reported in the Xcode test navigator, like so:
  
    image
 
    The skipped line will *not* be marked in the source code editor. It appears that Xcode only does that for statically defined tests (those marked static test), not runtime tests (those marked run-time test), which is what Quick uses exclusively.
  
     Future direction: when some tests are "focused", should we mark all other tests as skipped?

     The general error-handling path was extracted into a private `reportFailedTest` method.
   
    Resolves #1093

 - [x] Does this have tests?
 - [ ] Does this have documentation? *Idk, Should it?*
 - [ ] Does this break the public API (Requires major version bump)? No.

### DIFF
--- a/Sources/Quick/Example.swift
+++ b/Sources/Quick/Example.swift
@@ -62,7 +62,7 @@ final public class Example: _ExampleBase {
         Executes the example closure, as well as all before and after
         closures defined in the its surrounding example groups.
     */
-    public func run() { // swiftlint:disable:this function_body_length
+    public func run() {
         let world = World.sharedWorld
 
         if world.numberOfExamplesRun == 0 {
@@ -83,31 +83,11 @@ final public class Example: _ExampleBase {
             do {
                 try closure()
             } catch {
-                let description = "Test \(name) threw unexpected error: \(error.localizedDescription)"
-                #if SWIFT_PACKAGE
-                let file = callsite.file.description
-                #else
-                let file = callsite.file
-                #endif
-
-                // XCTIssue is unavailable (not implemented yet) on swift-corelibs-xctest (for non-Apple platforms)
-                #if canImport(Darwin)
-                let location = XCTSourceCodeLocation(filePath: file, lineNumber: Int(callsite.line))
-                let sourceCodeContext = XCTSourceCodeContext(location: location)
-                let issue = XCTIssue(
-                    type: .thrownError,
-                    compactDescription: description,
-                    sourceCodeContext: sourceCodeContext
-                )
-                QuickSpec.current.record(issue)
-                #else
-                QuickSpec.current.recordFailure(
-                    withDescription: description,
-                    inFile: file,
-                    atLine: Int(callsite.line),
-                    expected: false
-                )
-                #endif
+                if let testSkippedError = error as? XCTSkip {
+                    self.reportSkippedTest(testSkippedError, name: name, callsite: callsite)
+                } else {
+                    self.reportFailedTest(error, name: name, callsite: callsite)
+                }
             }
 
             self.group!.phase = .aftersExecuting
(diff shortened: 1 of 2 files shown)


Sources/Quick/Example.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+import XCTest
```
amomchilov, line 2: Needed for the `XCTSkip` error type.

Sources/Quick/Example.swift
```diff
@@ -136,6 +119,65 @@ final public class Example: _ExampleBase {
         }
         return aggregateFlags
     }
+	
+	private func reportSkippedTest(_ testSkippedError: XCTSkip) {
+		#if _runtime(_Native)
+			return // This functionality is Only supported by Apple's proprietary XCTest, not by swift-corelibs-xctest
+		#elseif _runtime(_ObjC)
```
amomchilov, line 126: What's the correct way check that we're using Apple's XCTest and not [corelibs-xctest](https://github.com/apple/swift-corelibs-xctest)? I see several options:

* `#if canImport(Darwin)`
* `#if _runtime(_ObjC)`
* `#if #if os(macOS) || os(iOS) || os(tvOS)`, seems fragile, it'll omit future targets that may be supported.

I also asked about this on the Swift forums: https://forums.swift.org/t/whats-the-right-way-to-distinguish-between-corelibs-frameworks-from-apple-propreitary-ones/51699

Sources/Quick/Example.swift
```diff
@@ -136,6 +119,65 @@ final public class Example: _ExampleBase {
...
+					the \(NSStringFromSelector(selector)) message necessary to report skipped tests to Xcode.
+					
+					If nobody else has done so yet, please submit an issue to https://github.com/Quick/Quick/issues
+					
+					For now, we'll just benignly ignore skipped tests.
+				""")
+				return
+			}
+		
+			testRun.perform(selector, with: testSkippedError.message, with: sourceCodeContext)
```
amomchilov, line 148: Would it be better to add this in a header file, and make the call-site "normal" as far as Swift is concerned?

amomchilov (commented):
Hey @ikesyo , could you please give some early feedback on my first draft?

Also, do you have any advice on how I should go about testing this?

Sources/Quick/Example.swift
ikesyo: Let's use `#if canImport(Darwin)`.

- https://github.com/Quick/Quick/pull/1040#discussion_r711749726
- #1104

Sources/Quick/Example.swift
ikesyo: I'm okay using `perform`.

ikesyo (changes requested):
Could you please fix the followings at first?

- indentation: use spaces over tabs
- SwiftLint violations (you can see them in the "Files changed" tab)

amomchilov:
> Could you please fix the followings at first?

@ikesyo I sure can, and I did. Though this is still an early stages draft. I was less concerned about the linting, I'm worried more about the substance.

I was hoping you could give me some guidance on how to write tests for this.

jessesquires:
Hey @amomchilov! 👋🏼 

I'm a new maintainer here trying to get v5.0 released. This looks like a great change and the diff is pretty small, so I think we can include it.

Are you willing and interested in following up on this fix?

jessesquires:
I have rebased the branch on main for you

amomchilov:
Thanks @jessesquires, I'd love to get this moving

> Are you willing and interested in following up on this fix?

I am, but I got stuck, and never heard back. My qualm still exists, I don't know to properly cover this with a test.

Could you help with that?

amomchilov:
I rebased again. There's two outstanding issues:

1. Testing
2. Deciding what to do about `QuickSpec.current.testRun`. Is it safe to force-unwrap it?

Sources/Quick/Example.swift
```diff
@@ -141,6 +121,94 @@ final public class Example: _ExampleBase {
         }
         return aggregateFlags
     }
+
+    private func reportSkippedTest(_ testSkippedError: XCTSkip, name: String, callsite: Callsite) { // swiftlint:disable:this function_body_length
+        #if !canImport(Darwin)
+            return // This functionality is only supported by Apple's proprietary XCTest, not by swift-corelibs-xctest
+        #endif
+
+        guard let testRun = QuickSpec.current.testRun else { fatalError() } // FIXME: improve this
```
jessesquires, line 130: I think keeping the guard is fine.

Let's just just `print()` and `return` like we do for the other guards below. 😊

jessesquires:
Thanks @amomchilov ! 🙌🏼 

> I rebased again. There's two outstanding issues:
>
> Testing

@younata Do you have any thoughts here?

> Deciding what to do about QuickSpec.current.testRun. Is it safe to force-unwrap it?

I left comments in-line.

jessesquires:
I would probably be ok omitting tests for this since it is a bit odd and doesn't seem like a massive risk.

Also, ignore the Danger check. The failure is expected for forks.

jessesquires:
Looks like the SwiftPM failures are valid.

amomchilov:
> I would probably be ok omitting tests for this since it is a bit odd and doesn't seem like a massive risk.

There's a certain irony to a test framework not being well tested, isn't there? 😛 

> Looks like the SwiftPM failures are valid.

Ah yes, `NSStringFromSelector` is only available to Linux. I gated it.

Let's see wht CI says this time

younata:
> Testing
> @younata Do you have any thoughts here?

Yes! Similar to how we verify that test failures are reported to Xcode (see [`ItTests.testImplicitErrorHandling()`](https://github.com/Quick/Quick/blob/main/Tests/QuickTests/QuickTests/FunctionalTests/ItTests.swift#L167-L174)), we can add another set of fixture tests that do the right thing and then verify with the test run that the correct skip count was recorded.

Something like this (I have not tried to run this code)

```swift
// This is the fixture
final class FunctionalTests_SkippingTestsSpec: QuickSpec {
    override func spec() {
        it("supports skipping tests") { XCTSkip("This test is intentionally skipped") }
        it("supports not skipping tests") { }
    }
}

final class ItTests: // ...
    func testSkippingExamplesAreCorrectlyReported() {
        let result = qck_runSpec(FunctionalTests_SkippingTestsSpec.self)!
        XCTAssertTrue(result.hasSucceeded)
        XCTAssertEqual(result.executionCount, 2)
        XCTAssertEqual(result.skipCount, 1)
        XCTAssertEqual(result.totalFailureCount, 0)
    }
    // ...
```

jessesquires:
@younata thank you! 🙌🏼 

@amomchilov looks like CI is ✅ now! 🎉 Let's just implement @younata's testing suggestion and then we can merge.

amomchilov:
This fails in some CI configurations, but not others. I can't figure out the cause, any ideas?

younata:
Maybe the API for marking tests as skipped changed in Xcode 12.5? All those test failures are referencing an Xcode 12.4 toolchain.

I'm ok with skipping these tests (hehe) on Xcode 12.4, especially considering that #1138 means we'll be dropping Xcode <13 anyway.

jessesquires:
> Maybe the API for marking tests as skipped changed in Xcode 12.5? All those test failures are referencing an Xcode 12.4 toolchain.

I vaguely remember seeing some tests that only ran for specific Xcode versions, but maybe that was something else?

Is there a good way to gate the tests to be ignored/skipped on Xcode 12.4?

> I'm ok with skipping these tests (hehe) on Xcode 12.4, especially considering that https://github.com/Quick/Quick/issues/1138 means we'll be dropping Xcode <13 anyway.

I'm ok with this too, but I'd like to do the following:

1. Let's submit a new PR that drops Xcode 12.4 support. This would mean updating any relevant code, as well as the GH workflows.
2. Merge that PR, then rebase this PR on top of it.

amomchilov:
> I'm ok with skipping these tests (hehe) on Xcode 12.4

How would I do that? Is there a way to make macros around Xcode versions?

amomchilov:
I found the issue with Xcode 12.4. Their API was just slightly different:

`XCTSkippedTestContext` used to have:

* `filePath: NSString`
* `lineNumber: NSString`

Which was passed to `-[XCTestRun recordSkipWithDescription:inFile:atLine:`.

After Xcode 12.4, this `filePath` and `lineNumber` was extracted into a new `sourceCodeContext: XCTSkippedTestContext` property, which is passed directly to a new method: `-[XCTestRun recordSkipWithDescription:sourceCodeContext:]`.

I added a shim to support Xcode 12.4, because it was pretty straight forward. We can ship this now (it should all be green! :D). In the future, if we want to go out of our way to drop 12.4, we can delete this shim

jessesquires (approved):
Thanks so much @amomchilov ! 🙌🏼

amomchilov:
Ah, I forgot to make some Objective-C dependant code to be guarded by `canImport(Darwin)`. Fixed.

amomchilov:
Third time's the charm?

amomchilov:
Woohoo, we're green!

@ikesyo requested changes a while ago, but apart from that, I think this is ready to merge

amomchilov:
@jessesquires What do we do if `@ikesyo` doesn't respond to re-review this?

jessesquires:
> What do we do if @ikesyo doesn't respond to re-review this?

It's all good, no worries! We only need 1 approval.

I'll merge this tomorrow -- and potentially tag the v5.1 release (still deciding on that and coordinating with Nimble).

ikesyo (approved):
LGTM! 👏

amomchilov:
@jessesquires Was it decided whether this will be tagged 5.1, or when it would be merged?

jessesquires:
Thanks @amomchilov ! 💯 

Decided to just keep this for v6.0 -- unfortunately I don't have capacity to do a v5.1 right now, and we haven't landed any other big changes. 😊

jessesquires:
But, there isn't _a ton_ of work to do for v6.0 -- so we can probably get that out within the next few months.

jessesquires:
@amomchilov invited you to be a contributor 😄

Issue #992, opened 2020-06-02
Delegate some XCTestCase methods to `QuickSpec.current`


ikesyo:
## Implementations
- #990

PR #1102, opened 2021-09-17
QuickSpec: replace deprecated recordFailure api with XCTIssue

bartosz-pawelczyk:
For swift 5.3 and above XCTIssue is used with record: api.

### DIFF
--- a/Sources/Quick/QuickSpec.swift
+++ b/Sources/Quick/QuickSpec.swift
@@ -139,12 +139,20 @@ open class QuickSpec: QuickSpecBase {
             return
         }
 
-        super.recordFailure(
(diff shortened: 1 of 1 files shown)


ikesyo (commented):
ref:
- #990 
- #992 

Two points:
- XCTIssue API is not implemented (yet) in [swift-corelibs-xctest](https://github.com/apple/swift-corelibs-xctest) so can't be used on non-Apple platforms.
- Calling `record(_ issue: XCTIssue)` in the implementation of overriding `recordFailure` is weird. It might be desirable to add delegation for `record(_ issue: XCTIssue)`.

Thanks for the PR though!

PR #1113, opened 2021-11-16
Fix Swift docs for non-named `configure` argument.

alexhayes:
This change fixes the docs for the `configure` method - which takes a non-named argument.

 - What behavior was changed? None
 - What code was refactored / updated to support this change? None
 - What issues are related to this PR? Or why was this change introduced? Because the docs were wrong.

 - [ ] Does this have tests?


jessesquires:
Thanks @alexhayes !

jessesquires (approved):

Issue #1123, opened 2022-03-16
QuickSpec tests are not referenced in Xcode13.3


nnsnodnb:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/main/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

`QuickSpec` tests were not eligible for execution in Xcode 13.3.

<img width="550" src="https://user-images.githubusercontent.com/9856514/158536272-f530bbad-3d73-4461-848e-40bcb675f758.png">

All tests are correctly recognized when installed with the Swift Package Manager.
But when installed with CocoaPods, all tests are ignored.

### What did you expect to happen?

Tests are executed.

### What actually happened instead?

All tests were ignored.

### Environment

List the software versions you're using:

 - Quick: *4.0.0*
 - Nimble: *?.?.?* not included
 - Xcode Version: *13.3 (13E113)* (Open Xcode; In menubar: Xcode > About Xcode)
 - Swift Version: *5.6* use `Xcode Default` (Open Xcode Preferences; Components > Toolchains. If none, use `Xcode Default`.)

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - CocoaPods: *1.11.3* (Use `pod --version` in Terminal)
 - Carthage: *?.?* (Use `carthage version` in Terminal) not included
 - Swift Package Manager *5.6.0 (swiftpm-???)* (Use `swift build --version` in Terminal)

### Project that demonstrates the issue

[Sample.zip](https://github.com/Quick/Quick/files/8259688/Sample.zip)

mimibuduo123:
Xcode13.3 testInvocations -> testBundleWillStart

mimibuduo123:
< Xcode13.3 testBundleWillStart < testInvocations

paweldudek:
We did a quick & dirty fix this way: https://github.com/toptal/Quick/commit/8c332c41770d87e62718531ed0951763c58f2345

But I also don't know what are the ramifications of building examples right before they're run (this also is slightly inefficient b/c the code to build examples is run before _every_ test - there's an internal check that prevents from duplicating tests for each class _but_ the check needs to be run for every test - not the best complexity here 😕). 

Would be nice for someone from Core team to chime in with direction, I can work out the details & submit a PR 🙌🏻

ignotusverum:
Adding `Execute in parallel(if possible)` flag in test scheme options - for unknown reason fixes issue with execution 

[...]

PR #1124, opened 2022-03-16
Fix XCTest lifecycle in Xcode13.3

mimibuduo123:
help you if you're not sure:

 - What behavior was changed?

 - [x] Does this have tests?

### DIFF
--- a/Sources/Quick/QuickTestObservation.swift
+++ b/Sources/Quick/QuickTestObservation.swift
@@ -12,16 +12,24 @@ import XCTest
 @objc internal final class QuickTestObservation: NSObject, XCTestObservation {
     @objc(sharedInstance)
     static let shared = QuickTestObservation()
-
-    // Quick hooks into this event to compile example groups for each QuickSpec subclasses.
-    //
-    // If an exception occurs when compiling examples, report it to the user. Chances are they
-    // included an expectation outside of a "it", "describe", or "context" block.
-    func testBundleWillStart(_ testBundle: Bundle) {
+    
+    override init() {
+        super.init()
         QuickSpec.enumerateSubclasses { specClass in
             // This relies on `_QuickSpecInternal`.
             (specClass as AnyClass).buildExamplesIfNeeded()
         }
+    }
+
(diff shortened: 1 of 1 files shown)


mimibuduo123:
#1123

Sources/Quick/QuickTestObservation.swift
```diff
@@ -12,16 +12,24 @@ import XCTest
 @objc internal final class QuickTestObservation: NSObject, XCTestObservation {
     @objc(sharedInstance)
     static let shared = QuickTestObservation()
+    
+    override init() {
+        super.init()
+        QuickSpec.enumerateSubclasses { specClass in
```
paweldudek, line 18: Note that the init will be called when types are instantiated (b/c of https://github.com/Quick/Quick/pull/1124/files#diff-6288283bc6481ff91171d7266acdfb23cfb0b8d5dba40176ba0e05046cd6b5f4R14). This results in the runtime subclass enumeration being called while classes are still being loaded into runtime. 

In our codebase this resulted in random crashes happening when the tests started running (I also tried a poor mans approach by wrapping this in an async call but that was already too late).

mimibuduo123: Could you tell me about the crash scene? A large number of case tests found no problems

mimibuduo123: and,   
```
    override init() {
        super.init()
        
        DispatchQueue.main.async {
            QuickSpec.enumerateSubclasses { specClass in
                // This relies on `_QuickSpecInternal`.
                (specClass as AnyClass).buildExamplesIfNeeded()
            }
        }
    }
```
is it better？

paweldudek: In our rather large code base (200k LoC with multiple modules) doing the init-based approach results in:

```
runtime assert: unexpected initialization order
```

This totally makes sense - this code is trying to walk through runtime while runtime is still being loaded (see screenshot below for details). 

Also as mentioned above: I tried wrapping this into an async call and that's too late - no tests are being registered 😞 

Screenshot 2022-03-18 at 13 14 54

PR #1128, opened 2022-03-27
Fix test cases discoverage in Xcode 13.3 

BobCatC:
Fix for #1123 

# How to reproduce?

1. Create a project with Quick using Xcode 13.3
2. Add unit tests
3. Run all tests

Expected results:
* Your tests evaluated

Actual results:
* Xcode couldn't discover any test and didn't run it
 
# What behavior was changed?

When running tests in Xcode 13.3 notification `testBundleWillStart` is called after method `testInvocations` of each QuickTest subclass. 
This means that when Xcode asks your QuickTest subclass for test cases to run (calling `testInvocations`) it gets empty array because test cases array will be computed on `testBundleWillStart` call.

# A less complex solution

The first solution that comes to mind is just to call `[self buildExamplesIfNeeded]` inside `testInvocations` method. This one works well at first sight and could help to remove `QuickTestObservation` and `QuickTestExamplesLoader` at all. 
But.
But there was [a commit](https://github.com/Quick/Quick/commit/6cf120fdd875c3077a37aa2d0561ec326e8a4e70) in which @ikesyo removes `[self buildExamplesIfNeeded]` from `testInvocations` (in past it already was there) and creates `QuickTestObservation`. And the reason of this changes is broken focus behaviour. Focus feature needs all examples to be computed before any test execution.

### DIFF
--- a/Sources/Quick/QuickTestExamplesLoader.swift
+++ b/Sources/Quick/QuickTestExamplesLoader.swift
@@ -1,0 +1,76 @@
+#if !SWIFT_PACKAGE
+
+import Foundation
+
+/// Singleton which can call `buildExamplesIfNeeded` on each `QuickSpec` subclass once per lifetime of application
+@objc internal final class QuickTestExamplesLoader: NSObject {
+    @objc(sharedInstance)
+    static let shared = QuickTestExamplesLoader()
+
+    private var didBuildExamples = false
+
+    private override init() {}
+
+    /// Iterates over all subclasses of `QuickSpec` and calls `buildExamplesIfNeeded` method on each one once per lifetime of application.
+    ///
+    /// If an exception occurs when compiling examples, report it to the user. Chances are they
+    /// included an expectation outside of a "it", "describe", or "context" block.
+    @objc func buildExamplesIfNeeded() {
+        guard didBuildExamples == false else { return }
+        didBuildExamples = true
+
+        QuickSpec.enumerateSubclasses { specClass in
+            // This relies on `_QuickSpecInternal`.
+            (specClass as AnyClass).buildExamplesIfNeeded()
+        }
+    }
+}
+
+/// A dummy protocol for calling the internal `+[QuickSpec buildExamplesIfNeeded]` method
+/// which is defined in Objective-C from Swift.
+@objc private protocol _QuickSpecInternal {
+    static func buildExamplesIfNeeded()
+}
+
+// swiftlint:disable:next todo
+// TODO: Unify this with QuickConfiguration's equivalent
+private extension QuickSpec {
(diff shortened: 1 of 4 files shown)


Sources/Quick/QuickTestExamplesLoader.swift
```diff
@@ -0,0 +1,76 @@
+#if !SWIFT_PACKAGE
+
+import Foundation
+
+/// Singleton which can call `buildExamplesIfNeeded` on each `QuickSpec` subclass once per lifetime of application
+@objc internal final class QuickTestExamplesLoader: NSObject {
```
BobCatC, line 6: Note that this class consists of 99% extracted logic from QuickTestObservation
The only thing changed - I've added a flag `didBuildExamples`

Sources/QuickObjectiveC/QuickSpec.m
```diff
@@ -24,6 +24,10 @@ @implementation QuickSpec
  @return An array of invocations that execute the newly defined example methods.
  */
 + (NSArray *)testInvocations {
+    // Xcode 13.3 hack, see this issue for more info: https://github.com/Quick/Quick/issues/1123
+    // In case of fix in later versions next line can be removed
+    [[QuickTestExamplesLoader sharedInstance] buildExamplesIfNeeded];
```
BobCatC, line 29: And here is the main change which helped to solve the problem

BobCatC:
Looks like there is another pull request merged in main with this fix

Sources/Quick/QuickTestExamplesLoader.swift
jessesquires: I think this PR will have a better chance of getting reviewed and merged more quickly by maintainers if you remove all the refactoring and keep the diff as small as possible.

The refactoring (which I agree with) can come in a 2nd PR after the main fix is merged. 😊

BobCatC: Hm, good idea
Created another one: #1129

BobCatC:
Closing this PR in order to create a new one with smaller diff: #1129 
Maybe it will be merged faster

jessesquires:
Hey @BobCatC 😊 

Now that https://github.com/Quick/Quick/pull/1129 is merged, would you like to open a PR with the refactoring?

BobCatC:
@jessesquires of course 😁
I was going to do it today
Also it looks like I have to add test cases for that fix (according to the comments by @ashfurrow)

jessesquires:
@BobCatC yes, tests would be appreciated!

BobCatC:
@jessesquires see #1134 with refactoring and new test

PR #1133, opened 2022-04-14
[Breaking] Rename Configuration -> QCKConfiguration

### LINKED ISSUE #1038: Name clash in generated Obj-C headers

This isn't necessarily a bug in Quick, but I believe it's useful to have it for the record and to avoid more tickets like this as more people try the beta.

### What did you do?

Tried to build my project for testing with Xcode 12.5 beta.

I expected it would build as usual.

Build of one of the test targets failed due to a name clash:

Screen Shot 2021-02-05 at 10 26 31
Screen Shot 2021-02-05 at 10 25 39

Both Quick and another library have `Configuration` class, and they clash when emitted to Obj-C headers.

This looks very much like https://developer.apple.com/forums/thread/658012, even the workaround mentioned there to set `SWIFT_INSTALL_OBJC_HEADER` to `NO` works (well, as long as one doesn't need Quick's Obj-C interface, that is).

### Environment

 - Quick: *3.0.0*
 - Nimble: *9.0.0*
 - Xcode Version: *12.5 beta (12E5220o)*, new build system
 - Swift Version: *5.4*

 - CocoaPods: *1.10.1*

The other library involved is a closed source one, unfortunately.

jessesquires:
Closes #1038


kmcbride:
Couldn't you have avoided breaking Swift compatibility simply by annotating the type with `@objc(QCKConfiguration)`?

PR #1129, opened 2022-04-05
Critical! Fix test cases discoverage in Xcode 13.3

BobCatC:
Fix for #1123 

# Why critical?

Because Quick does not work with Xcode 13.3.
 
## What behavior was changed?

When running tests in Xcode 13.3 notification `testBundleWillStart` is called after method `testInvocations` of each QuickTest subclass. 
This means that when Xcode asks your QuickTest subclass for test cases to run (calling `testInvocations`) it gets empty array because test cases array will be computed on `testBundleWillStart` call.

## Step 2

As suggested @jessesquires this is step 1 with most important changes that fix the problem. Step 2 will contain a bit of refactoring with extracting examples loading into separate class as now it looks a bit strange that `QuickSpec.testInvocations` method calls `QuickSpecObservation`.

Issue #1115, opened 2022-02-08
Failing test succeeds


zbencz3:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/main/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

When running tests, the developer has the option to click on the red diamond on Xcode's gutter to re-run tests. When the top most diamond is actioned, the tests correctly failed, but when the failing test's local diamond is actioned, Xcode reports the test as a success despite the code not being changed. See the sample project and the video demoing the problem.

https://user-images.githubusercontent.com/5381246/153002826-fc23651b-a61e-4e0e-ac3a-2a3d78e22895.mov

### What did you expect to happen?

I expected the test to correctly fail or succeed.

### What actually happened instead?

The test succeeded even though it should have failed.

### Environment

List the software versions you're using:

 - Quick: 4.0.0
 - Nimble: 9.2.1
 - Xcode Version: Version 13.2.1 (13C100)
 - Swift Version: Xcode Default

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - CocoaPods: 1.11.2

### Project that demonstrates the issue

[QN.zip](https://github.com/Quick/Quick/files/8024417/QN.zip)

zbencz3:
Related note: I was hoping that I could use the `control + option + command + U` shortcut to re-run the selected current test case, but that seems to only work for XCTest methods, not for Quick/Nimble. See how that menu item is disabled. Any known workaround for that?

<img width="471" alt="Disabled" src="https://user-images.githubusercontent.com/5381246/153004868-e5155292-437c-44c6-9ff2-bacd6bb4520e.png">

PR #1134, opened 2022-04-14
Refactoring after fixing test cases discoverage in Xcode 13.3

BobCatC:
# Intro

Next step after #1129 

As we agreed with @jessesquires this is a refactoring PR after small fix of Xcode 13.3 issue. What code was refactored read below.

## `AllExamplesBuilder` class

Just extracted logic of calling method `buildExamplesIfNeeded` on each `QuickSpec` subclass from  `QuickTestObservation` into separate class.
You can note that the instance has a flag - `didBuildAllExamples` which helps to execute method only once per lifetime of the object. And also you can note that the object is used as a property of `World` instance. So its lifetime equals world's lifetime (which in real projects acts like a singleton).

## Tests for the issue

I've added a test for Xcode 13.3 issue. And to be able to test that behaviour I had to add into function `qck_runSpec` new param - `gatherExamples`. Passing `false` simulates Xcode 13.3 problem, passing `true` (which is default) behaves like before.

## Checklist

 - **Does this have tests?** Yes
 - **Does this have documentation?** No (needed?)
 - **Does this break the public API (Requires major version bump)?** No
 - **Is this a new feature (Requires minor version bump)?** No


BobCatC:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

BobCatC:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Big PR</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

BobCatC:
Danger fails as usual because I'm not a contributor. Ran it locally - it's ok (see comment above).

jessesquires (changes requested):
Thanks @BobCatC ! 💯 

I think this looks good. 

One request:

Can you comment to call out what was simply moved vs what was changed? I think I understand what's happening, but this would be very helpful — and good for posterity.

jessesquires:
@BobCatC oh also, I've marked this for the v5.0 release. 

This would be the final PR to merge for that.

It doesn't seem very risky, but I would appreciate your thoughts on the risk level as well.

jessesquires:
@BobCatC one last question. Does this also address #1115?

Sources/Quick/AllExamplesBuilder.swift
```diff
@@ -0,0 +1,68 @@
...
+
+        QuickSpec.enumerateSubclasses { specClass in
+            // This relies on `_QuickSpecInternal`.
+            (specClass as AnyClass).buildExamplesIfNeeded()
+        }
+    }
+}
+
+// swiftlint:disable:next todo
+// TODO: Unify this with QuickConfiguration's equivalent
```
jessesquires, line 28: I think we should address this before merging. 

But I don't have a ton of context here.

jessesquires:
Hey @younata -- not sure if you saw, but GitHub restored all the previously-deleted Russian account contributions. So this PR is back now. 😊  https://www.jessesquires.com/blog/2022/04/19/github-suspending-russian-accounts/#updated-21-april-2022

Issue #1139, opened 2022-04-16
5.0.0 execute in parallel broken


skywalkerdude:
- [ x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/main/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Related to https://github.com/Quick/Quick/issues/1123, but I updated my library to version 5.0.0 and tried running my Quick tests with "execute in parallel" on.

### What did you expect to happen?

Tests would run

### What actually happened instead?

Tests failed with error
```
 Assertions: System: Unexpected TestCaseDidStart. If you believe this error represents a bug, please attach the result bundle at ...DerivedData/Sample-acinzmmaeypryxgcsczdnkgqvssb/Logs/Test/Test-Sample-2022.04.16_00-09-09--0700.xcresult
```
<img width="1512" alt="Screen Shot 2022-04-16 at 12 14 01 AM" src="https://user-images.githubusercontent.com/1427524/163665679-4db510f0-07c4-47de-9ed5-916124f97427.png">

### Environment

List the software versions you're using:

 - Quick: *5.0.0*
 - Nimble: *N/A*
 - Xcode Version: *13.3*
 - Swift Version: `Xcode Default`

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - CocoaPods: *1.11.3*
 - Carthage: *N/A*
 - Swift Package Manager: *N/A*

### Project that demonstrates the issue
[Sample.zip](https://github.com/Quick/Quick/files/8499458/Sample.zip)

NachoSoto:
Same here:
<img width="475" alt="Screen Shot 2022-04-16 at 15 04 25" src="https://user-images.githubusercontent.com/685609/163676059-dfe16015-28a2-421d-bee0-06bd142518bd.png">

younata:
This is also an issue in Xcode 13.2 at least.

This is likely the result of #1129 solving the issue in an incomplete way. I'll investigate it some more this evening.

jessesquires:
Fixed in #1143

PR #1143, opened 2022-04-19
Fix running tests in parallel

younata:
Solves #1139

This reverts 90ec83d37cc8e943944ee07ed0ea9cf2f541c6b1. The original PR that introduced it has since been deleted, so I'm unsure exactly of the intention of that contribution. The commit mentions being able to run a selected test, which is still doable with this revert.

Tests here is setting some of the test targets for each of iOS, tvOS, and macOS to run in parallel.

### DIFF
--- a/Quick.xcodeproj/xcshareddata/xcschemes/Quick-iOS.xcscheme
+++ b/Quick.xcodeproj/xcshareddata/xcschemes/Quick-iOS.xcscheme
@@ -62,8 +62,7 @@
             </BuildableReference>
          </TestableReference>
          <TestableReference
-            skipped = "NO"
-            parallelizable = "YES">
+            skipped = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "DA9876B11A4C70EB0004AA17"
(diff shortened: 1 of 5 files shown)


younata:
@jessesquires Given the breaking issue of this bug, I think we should release a bug fix version immediately/shortly after this is merged.

jessesquires:
> The original PR that introduced it has since been deleted, so I'm unsure exactly of the intention of that contribution.

The same thing happened with the Xcode 13.3 fix PR. Both of these users completely deleted the GH Accounts. 

wtf is going on. 🤔 🤷🏼‍♂️

jessesquires (approved):
Thanks @younata ! 💯

tsapeta:
Thanks guys for taking care of Quick and Nimble, you're doing a great job for the community! 🙏 

> > The original PR that introduced it has since been deleted, so I'm unsure exactly of the intention of that contribution.
> 
> The same thing happened with the Xcode 13.3 fix PR. Both of these users completely deleted the GH Accounts.
> 
> wtf is going on. 🤔 🤷🏼‍♂️

Looks like GitHub started suspending accounts of Russian devs 🤔

jessesquires:
> Looks like GitHub started suspending accounts of Russian devs 🤔

Ohh fuck. Yeah, that explains it.

Wow, that's super fucked up. These are just civilian individuals, as far as I know.

PR #1146, opened 2022-04-21
Raise minimum supported versions to macOS 10.15, iOS 13, tvOS 13

### LINKED ISSUE #1138: Drop old OS and Swift versions
I don't see much of a reason or benefit to keep these around and it will simplify many things, including CI.

### Suggested changes:

- iOS/tvOS 13+
- macOS 10.15+
- Swift 5.5+
- Xcode 13+

### Todo:

- [ ] Update project to drop old deployment targets
- [ ] Update podspec
- [ ] Update Package.swift
- [ ] Update CI

younata:
Resolves #1138

Change minimum supported versions in the the xcodeproj so that it's easier in the future to change this if we desire.

**This drops support for Xcode 12 and earlier, macOS 10.10 through 10.14, iOS 9 through 12, and tvOS 9 through 12**


younata:
The linux tests are broken because, apparently, when you specify `swift-tools-version:5.5`, you can no longer call bare functions in LinuxMain that aren't `XCTMain`? If we specify 5.4 or earlier, then this isn't an issue. I don't want to do that because I'd rather actually solve this issue.

I'm currently researching how to address this.

jessesquires:
Actually, on second thought, I'm wondering if we should do iOS 12+ and macOS 10.14+ instead? What do you think @younata? 

A couple notes:

- Don't forget to update the podspec.
- When this is ready, I can remove the required status checks for Xcode 12.4, 12.5, etc.
- While we're here, let's get rid of the `swift nightly` status check -- I don't think that's valuable.

younata:
> I'm wondering if we should do iOS 12+ and macOS 10.14+ instead

Sure! I'm mostly concerned with dropping support for Xcode 12, supporting the past 4 instead of 3 major OSs is fine with me.

> Don't forget to update the podspec.

Thank you! I had forgotten about it.

> While we're here, let's get rid of the swift nightly status check -- I don't think that's valuable.

Good point. This was likely much more valuable in the swift 2 or 3 days, but now that swift has stabilized, I agree that there's not much point keeping the nightly check around.

> The linux tests are broken because, apparently, when you specify swift-tools-version:5.5, you can no longer call bare functions in LinuxMain that aren't XCTMain? If we specify 5.4 or earlier, then this isn't an issue. I don't want to do that because I'd rather actually solve this issue.
>
>I'm currently researching how to address this.

I discovered that this is caused by [SE-0323](https://github.com/apple/swift-evolution/blob/main/proposals/0323-async-main-semantics.md), and we can address it by wrapping QCKMain calls like so:

```swift
@main struct Main {
    static func main() {
        QCKMain ...
    }
}
```

I'll update our documentation to mention this (and actually write documentation for using Quick with SPM).

Documentation/en-us/InstallingQuick.md
```diff
@@ -141,13 +141,81 @@ This is not "the one and only way" to use Carthage to manage dependencies.
...
+    	.testTarget(name: "MyPackageTests", dependencies: ["Quick", "Nimble"])
+    ]
+)
+```
+
+#### Linux Support
+
+On Linux, you will need to also add a `LinuxMain.swift` at the root of the `Tests` subdirectory. The contents of this branches depending on the `swift-tools-version` your `Package.swift` file specifies.
+
+##### Swift Tools Version 5.4 and earlier
```
jessesquires, line 180: Is it possible / easy for us to support only Swift 5.6 and above?

Supporting old Swift versions isn't as important as supporting old OS versions -- the language versions are compatible with all OSes essentially (and we're dropping Xcode 12 and below), and the Swift compiler has very good support for mixing Swift versions. So even if someone is using Swift 5.4 or 5.5 for some reason, this shouldn't have a huge impact on their project.

Package.swift
```diff
@@ -1,11 +1,11 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.5
 
 import PackageDescription
 
 let package = Package(
     name: "Quick",
     platforms: [
-        .macOS(.v10_10), .iOS(.v9), .tvOS(.v9)
+        .macOS(.v10_15), .iOS(.v13), .tvOS(.v13)
```
jessesquires, line 8:
```suggestion
        .macOS(.v10_14), .iOS(.v12), .tvOS(.v12)
```

jessesquires (changes requested):
Looks good 😎

jessesquires:
Let's be sure to coordinate this with #1098.

I'm thinking it might be worth releasing a 5.1 with that change, then we could land this after.

Documentation/en-us/InstallingQuick.md
```diff
@@ -141,13 +141,81 @@ This is not "the one and only way" to use Carthage to manage dependencies.
...
+import PackageDescription
+
+let package = Package(
+    name: "MyPackage",
+    products: [
+    	.library(name: "MyPackage", targets: ["MyPackage"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/Quick/Quick.git", from: "5.0.1"),
+        .package(url: "https://github.com/Quick/Nimble.git", from: "9.2.1"),
```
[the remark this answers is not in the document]
younata: Good point. Might be worth waiting until nimble v10, so that we can update the minimum version of Nimble here.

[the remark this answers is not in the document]
jessesquires: Sounds good. I'll tag the Nimble v10 release today or tomorrow

Documentation/en-us/InstallingQuick.md
younata: As far as I can tell, the only way for us to support only 5.6 and above is to start using 5.6 features without guarding them with the appropriate `#if swift(>=5.6)`.

younata:
Getting back to this because I want to add async/await support to Quick and Nimble, and this feels like a pre-req to doing that.

younata:
> Actually, on second thought, I'm wondering if we should do iOS 12+ and macOS 10.14+ instead?

My only objection to iOS 12+ instead of 13+ is we'd have to add a bunch of `@available` attributes for any async/await APIs we add. Which isn't that much of an issue.

jessesquires:
> My only objection to iOS 12+ instead of 13+ is we'd have to add a bunch of @available attributes for any async/await APIs we add. Which isn't that much of an issue.

Ah, you know what, in that case let's do 13+

jessesquires:
🎉 thanks @younata !

PR #1153, opened 2022-05-10
version up Nimble in Package.swift

kimxwan0319:
I version up Nimble in Package.swift to 10.0.0


younata (approved):
Oof. Yeah. This is probably presenting an issue for people using SPM with this. Should've done this as soon as Nimble 10 was released. I'll add updating Quick with current Nimble to the release checklist for Nimble.

younata:
The "waiting for status to be reported" tests will never run because they're no longer valid.

PR #1154, opened 2022-06-01
Support watchOS

JosephDuffy:
This is a counterpart to https://github.com/Quick/Nimble/pull/916, which added support for watchOS to Nimble.

Note that this is a draft until a new version of Nimble is released with watchOS support. Until then this points at `main`.

I'm also not sure if the tests will run on CI correctly because the tests require watchOS 7.4 to run and I'm not sure what version of watchOS the simulators are configured with.

I also found that the tests weren't compiling (using Xcode 13.2.1) so I've updated these too.


MikeChugunov:
Hey @JosephDuffy, thanks for doing this! Looks like this is still a draft for a reason… 😅 Is this doable or is there some major blocker?

JosephDuffy:
@MikeChugunov I just merged in the latest changes, I think this is good to go :)

chugunomi:
Hey @younata! I know you done some work on making watchOS support happen for Nimble. Can you please look into this one too?

PR #1156, opened 2022-06-06
Fix sporadic crashes caused by finding classes that don't play well with isSubclass(of:)

### LINKED ISSUE #1155: Risky subclass checking in QuickSpec and QuickConfiguration
### What did you do?

When running tests, I run into sporadic crashes in `QuickConfiguration.configureSubclassesIfNeededWithWorld`. 

Similar to these issues
https://github.com/Quick/Quick/issues/1114
https://github.com/Quick/Quick/issues/1025

I expect there to be no crash.

It crashes.

### Environment

 - Quick: *4.0.0*
 - Nimble: *9.2.1*
 - Xcode Version: *13.3.1*
 - Swift Version: *5.6*

 - CocoaPods: *1.11.3*

younata:
Also grabs the list of objc classes in a much more thread-safe way.

Fixes #1155

---

This changes how we determine if a given class is a subclass of QuickSpec or QuickConfiguration. This is likely slower than using `isSubclass(of:)`. However, according to the radar @tiarnann filed, this should eliminate the sporadic crashes in subclass detection that keep coming up. So that's definitely worth fixing the issue.

Additionally, this replaces usage of `objc_getClassList` with `objc_copyClassList`. Apparently the get method has potential race conditions that can cause crashes, while the copy method does not.

### DIFF
--- a/Sources/Quick/QuickTestObservation.swift
+++ b/Sources/Quick/QuickTestObservation.swift
@@ -27,46 +27,10 @@ import XCTest
         guard !didBuildAllExamples else { return }
         didBuildAllExamples = true
 
-        QuickSpec.enumerateSubclasses { specClass in
+        enumerateSubclasses { (specClass: QuickSpec.Type) in
             // This relies on `_QuickSpecInternal`.
             (specClass as AnyClass).buildExamplesIfNeeded()
         }
-    }
-}
-
-// swiftlint:disable:next todo
-// TODO: Unify this with QuickConfiguration's equivalent
-extension QuickSpec {
-    internal static func enumerateSubclasses(
-        subclasses: [QuickSpec.Type]? = nil,
-        _ block: (QuickSpec.Type) -> Void
-    ) {
-        let subjects: [QuickSpec.Type]
-        if let subclasses = subclasses {
-            subjects = subclasses
-        } else {
-            // See https://developer.apple.com/forums/thread/700770.
-            var classesCount: UInt32 = 0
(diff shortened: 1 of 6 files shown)


Sources/Quick/QuickTestObservation.swift
```diff
@@ -45,24 +45,18 @@ extension QuickSpec {
         if let subclasses = subclasses {
             subjects = subclasses
         } else {
-            let classesCount = objc_getClassList(nil, 0)
+            // See https://developer.apple.com/forums/thread/700770.
+            var classesCount: UInt32 = 0
+            guard let classList = objc_copyClassList(&classesCount) else { return }
+            defer { free(UnsafeMutableRawPointer(classList)) }
+            let classes = UnsafeBufferPointer(start: classList, count: Int(classesCount))
```
jessesquires, line 52: Could we extract this into a helper function and reuse it?

jessesquires (approved):
thanks @younata! 💯 looks good

jessesquires:
@younata danger is failing because you are pulling from a fork. 

if you branch directly from the repo, danger should run successfully. but not a big deal.

Sources/Quick/QuickTestObservation.swift
younata: Done.

younata:
> @younata danger is failing because you are pulling from a fork.
> 
> if you branch directly from the repo, danger should run successfully. but not a big deal.

Oops. Future PRs from me will be from a branch on this repo.

PR #1159, opened 2022-07-01
Support for Swift Concurrency.

younata:
This PR provides basic swift concurrency support. Now, you can `await` functions inside of `beforeEach`, `afterEach`, `aroundEach`, `it`, `fit`, and `xit` blocks.

For example, the following tests are now possible:

```swift
describe("my awesome tests") {
    beforeEach {
        await someAsyncSetupFunction()
    }

    afterEach {
        await teardownButAsync()
    }

    it("can run tests asynchronously!") {
        let result = await methodBeingTested()
        expect(result).to(equal(expectedValue))
    }
}
```

Objective-C support is unchanged - this provides the same API to Objective-C, and Objective-C tests will run without any noticeable differences.

## Todo

As-is, this PR will not allow the common way scoping variables to example groups. For example, the following code will **fail to compile**:

```swift
describe("some tests") {
    var someSharedProperty: Int = 0
    beforeEach {
        someSharedProperty = 1 // Error: Mutation of captured var 'exampleMetadata' in concurrently-executing code
    }
}
```

On the surface, I'm somewhat ok with this, provided we also give a migration path. This way of scoping variables is an antipattern, because these variables are scoped to the `describe` block, which runs long before the test blocks run. In fact, this commonly causes test failures or even crashes when a variable is not properly reset before being accessed again. That said, I'd love to solve this in a way that doesn't require everyone who uses Quick to rewrite their tests. Note: This is something @amomchilov's [result builder PR](https://github.com/Quick/Quick/pull/1149) would address.

Nimble also will need to add Async support, which XCTest's assertion functions don't even support as of this writing.


younata:
Wrong branch. Sorry about that.

PR #1163, opened 2022-07-09
Support stopping test execution once a test has ended.

### LINKED ISSUE #249: Provide an API around continueAfterFailure for XCTestCases
More historic context is on [Nimble#69](https://github.com/Quick/Nimble/issues/69).

By default, XCTestCase sets this as `YES`, which makes this continue running the test, even when a failure is reached.

I'm posing this a few questions:
1. Should Quick provide an API that wraps this behavior (eg - global application of a QuickConfiguration setting)?
2. what is the default value for this setting, if we're providing one.

I'm my votes are yes for the first one and `continueAfterFailure = true` (whatever XCTest defaults to).

younata:
Resolves #249

This adds a public property on `Example` for discontinuing the current running test upon a test failure. This relies on XCTest's `continueAfterFailure` mechanism to work.

Usage looks like:

```swift
beforeEach { exampleMetadata in
    exampleMetadata.example.continueRunningTestAfterFailure = false
}

it("stops running tests after the first failure") {
    fail("")
    fail("You should not see this")
}
```

Note: Due to how we implement `afterEach` and `aroundEach`, setting this property will also prevent the `afterEach` blocks from running. Likewise, it will also stop any part of `aroundEach` blocks that are after the `runExample` closure call. I can live with the `aroundEach` issue, because the primary usage of that feature is things like `autoreleasepool { runExample() }`, which wouldn't be affected by this anyway. But I'd like to also run the `afterEach` blocks afterwards. Hence, draft.

Presumably, if we actually use the `addTeardownBlock` API in XCTest for the `afterEach` blocks, then they would be run. I'll be taking a look at that in a bit, but I wanted a preview PR up while I figure out how to reimplement example.run() to use `addTeardownBlock`.


younata:
I don't think this is necessary now that https://github.com/Quick/Quick/pull/1165 is merged.

damonZDM:
beforeEach {
    $0.example.continueRunningTestAfterFailure = false
}

Why can't I find this property in version 7.3.0?

PR #1147, opened 2022-04-23
Create funding.yml

jessesquires:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository

PR #1164, opened 2022-08-06
Add @younata to funding.yml

younata:
Following up from https://github.com/Quick/Quick/pull/1147.

I finally got around to setting up GitHub sponsors, so adding myself to the people to sponsor for Quick and Nimble.


jessesquires:
🎉

Issue #209, opened 2014-11-30
justBeforeEach?


modocache:
Apparently [Ginkgo](https://github.com/onsi/ginkgo) has a [`JustBeforeEach`](http://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach) node. Perhaps Quick should also have such a feature? Pull requests or discussion on the topic welcome.

PR #1169, opened 2022-09-07
Introduce an `assignBefore` operator

esilverberg:
**Overview**
Quick is missing the ability to assign values to variables in interior blocks, but read those variables in a common exterior block. This is similar to the lazy evaluation of the `let` operation in rspec. This change introduces the ability to assign variable values prior to the execution of the `beforeEach` blocks.

**What behavior was changed?**
A new operator, `assignBefore`, was added to the DSL and logic of the `Example` class

**What code was refactored / updated to support this change?**
I have amended the core logic of the `Example` class, added `assignBefore` to the DSL, and added a new array of closures stored by the `ExampleHooks` class.

**What issues are related to this PR? Or why was this change introduced?**
I have written a blog post describing my rationale. See [this link](https://medium.com/perry-street-software-engineering/making-quickspec-lazier-proposing-the-assignbefore-operator-2abc9df47b8d).

 - [x] Does this have tests?

If this is a direction you guys want to move in, I am happy to cleanup and do this additional work! If not, no worries, will keep as a fork. Lastly, if you have a better solution in Swift than this modification, I am all ears!

### DIFF
--- a/Sources/Quick/DSL/DSL.swift
+++ b/Sources/Quick/DSL/DSL.swift
@@ -163,6 +163,19 @@ public func aroundEach(_ closure: @escaping AroundExampleWithMetadataClosure) {
 }
 
 /**
+     Defines a closure to be run prior to each example but after any beforeEach blocks.
+     This closure is not run for pending or otherwise disabled examples.
+     An example group may contain an unlimited number of justBeforeEach. They'll be
+     run in the order they're defined, but you shouldn't rely on that behavior.
+
+    - parameter closure: The closure to be run prior to each example and after any beforeEach blocks
+*/
+
+public func justBeforeEach(_ closure: @escaping BeforeExampleClosure) {
+    World.sharedWorld.justBeforeEach(closure)
+}
+
+/**
     Defines an example. Examples use assertions to demonstrate how code should
     behave. These are like "tests" in XCTest.
 
--- a/Documentation/en-us/QuickExamplesAndGroups.md
+++ b/Documentation/en-us/QuickExamplesAndGroups.md
@@ -221,10 +221,6 @@ In some cases you will want to have a group of tests with a common test setup bu
 ```swift
 // Swift
 
-enum ClickError: Error {
-  case dolphinIsSleeping
-}
-
 describe("a sleeping dolphin") {
     var sleeping: Bool!
     var click: Click!
--- a/Documentation/en-us/QuickExamplesAndGroups.md
+++ b/Documentation/en-us/QuickExamplesAndGroups.md
@@ -214,6 +214,49 @@ of typing!
 
 To execute code *after* each example, use `afterEach`.
 
+### Sharing Setup/Teardown Code Using `justBeforeEach`
+
+In some cases you will want to have a group of tests with a common test setup but different configuration of that setup code in the individual tests. This makes the most sense when you have an API that you are mocking, and logic on top of that API that will make different decisions based on the result.
+
+```swift
+// Swift
+
+enum ClickError: Error {
+  case dolphinIsSleeping
+}
+
+describe("a sleeping dolphin") {
+    var sleeping: Bool!
+    var click: Click!
+
+    justBeforeEach {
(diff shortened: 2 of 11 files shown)


Sources/Quick/DSL/DSL.swift
```diff
@@ -309,3 +309,9 @@ public func fitBehavesLike<C>(_ behavior: Behavior<C>.Type, file: FileString = #
 }
 
 // swiftlint:enable line_length
+
+/* es added */
```
younata, line 313: What does this comment mean?

younata (approved):
Heya @esilverberg! Thanks so much for the contribution!

To provide some background, this feature exists in a few other test suites. [Cedar](https://github.com/cedarbdd/cedar) calls it [`subjectAction`](https://github.com/cedarbdd/cedar/wiki/Writing-specs#extracting-common-setup---subjectaction-blocks), [ginkgo](https://github.com/onsi/ginkgo) calls it [`justBeforeEach`](https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach). There was a [discussion years ago about this exact feature](https://github.com/Quick/Quick/issues/209), but I'm really happy to see a PR for this.

Not to bikeshed, but I'm curious why you came up with the name `assignBefore`. Obviously it can be used for more than just that (hence why ginkgo calls it `justBeforeEach`).

I'm also curious what your thoughts are in regard to allowing more than one `assignBefore` per test, and what you think the order of these should be. Obviously, the code allows more than one per test, and this would execute them in the order they're specified in, under the same tree-like structure that `beforeEach` follows.

As for some next steps, I'd appreciate it if you could add some documentation. It doesn't need to be in all the languages we have documentation for, but certainly for the English language documentation. I think that [Documentation/en-us/QuickExamplesAndGroups.md](https://github.com/Quick/Quick/blob/main/Documentation/en-us/QuickExamplesAndGroups.md) would be a good place for it.

esilverberg:
@younata Thank you for your feedback!

I have read the documentation of Gingko's `justBeforeEach` and modified my PR accordingly. Now, the keyword is indeed `justBeforeEach` and it behaves as does Gingko's, where a developer should put their setup in `justBeforeEach` blocks and configuration in `beforeEach` blocks.

I have added documentation and tried to follow the Dolphin example.

This code will allow multiple `justBeforeEach` blocks in an example. This is aligned with the behavior of `beforeEach`. Personally I think that is not a good idea as a test writer, but I am inclined to mirror the current behavior.

The one thing I did not do is attempt to add objc support. Please advise if that is a requirement. No one I know continues to write in objc so I felt it was not necessary.

younata:
Hey @esilverberg, thanks for the update and documentation! We really appreciate that!

> This code will allow multiple justBeforeEach blocks in an example. This is aligned with the behavior of beforeEach. Personally I think that is not a good idea as a test writer, but I am inclined to mirror the current behavior.

While I agree with you on this, I also think that this is a style choice that should be left up to individual developers.

> The one thing I did not do is attempt to add objc support. Please advise if that is a requirement. No one I know continues to write in objc so I felt it was not necessary.

Unfortunately, objc support is a requirement. We do have users who are still using Quick with Objective-C. This should be as simple as adding the appropriate definitions in `Quick/Sources/QuickObjectiveC/DSL/QCKDSL.h` (`qck_justBeforeEach`, and `justBeforeEach`), as well as implementing `qck_justBeforeEach` to call `[[World sharedWordl] justBeforeEach:closure]` in `QCKDSL.m`. We'd also appreciate adding an objc version of `JustBeforeEachTests` to the `Quick/Tests/QuickTests/FunctionalTests/ObjC/` directory.

Again, thank you so much!

esilverberg:
OK I have added objc support and written an objc test that mirrors the test I wrote in Swift.

younata:
Thank you so much @esilverberg! I/we really appreciate this!

esilverberg:
FYI I have updated by blog post to include discussion of this new feature! https://esilverberg.medium.com/making-quickspec-lazier-proposing-the-assignbefore-operator-2abc9df47b8d

Issue #1178, opened 2022-11-03
Add a property wrapper to auto-nil-out test variables.


younata:
A very common pattern in writing tests is to have most variables used in test to be force-unwrapped optionals, like so:

```swift
var value: SomeObject!

beforeEach {
    value = SomeObject()
}

afterEach {
    value = nil
}

it("is used in test") {
    // do something with `value`
}
```

I think that a property wrapper (`@TestState` perhaps?) that automatically nils out the wrapping variable in an `afterEach` would be a good thing to add. This property wrapper could optionally take in a closure to initialize it. To be used like this:

```swift
@TestState(SomeObject())
var value: SomeObject!

it("is used in test") {
    expect(value).to(equal(SomeObject())
}
```

younata:
This was resolved in #1186.

PR #1186, opened 2022-11-24
Added a `TestState` property wrapper.

CraigSiemens:
Adds a `TestShared` property wrapper as suggested in
- #1178

It will always reset a property to nil at the end of a test. If the property wrapper is created with a value, the value will be set at the start of a test.

 - [x] Does this have tests?


younata (approved):
This is awesome! Thank you for taking care of it!

CraigSiemens:
Any idea why the tests are failing? Each `expect` call in the tests I added is getting the error
```
error: expression is 'async' but is not marked with 'await'
```

I'm not seeing anything obviously different between my tests and the existing ones.

younata:
> Any idea why the tests are failing? Each `expect` call in the tests I added is getting the error
> 
> ```
> error: expression is 'async' but is not marked with 'await'
> ```
> 
> I'm not seeing anything obviously different between my tests and the existing ones.

Oh. This is because Quick is using the 11.0 version of Nimble. This was fixed in the 11.2 version of Nimble. I just merged a [PR that brings it in](https://github.com/Quick/Quick/pull/1187). I updated your branch with now-current main. Hopefully there's nothing else.

younata:
Ok, great. Tests have all passed. I'm sorry for this confusion. I'll merge it now.

Thank you so much for your contribution, @CraigSiemens!

PR #1187, opened 2022-11-25
Update Nimble to 11.2.1, remove now-unnecessary usage of awaits in tests. Drop support for swift 5.6/Xcode 13.3.1

younata:
[Nimble 11.2.0](https://github.com/Quick/Nimble/releases/tag/v11.2.0) fixed the issue where any usage of async behavior in a test would cause the compiler to think that any usage of `expect` was taking in an async closure. This PR pulls in the now-current [11.2.1 version of Nimble](https://github.com/Quick/Nimble/releases/tag/v11.2.1), and sets the Package.swift to require at least 11.2.0.

This should make life significantly easier for those contributing to Quick.

### DIFF
--- a/.github/workflows/ci-swiftpm.yml
+++ b/.github/workflows/ci-swiftpm.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: macos-12
     strategy:
       matrix:
-        xcode: ["13.3.1", "14.0.1"]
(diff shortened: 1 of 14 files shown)


younata:
Oh. Right. 11.1.0 dropped support for Swift 5.6. Cool. I'll put this PR on hold for now.

Issue #1189, opened 2022-12-12
Upgrading a large codebase to Quick 6 is really problematic and `toEventually` is broken in Obj-C


ViktorSimko:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/main/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

Tried upgrading to Quick 6.

### What did you expect to happen?

Quick 6 has an incremental way of adopting concurrency and `toEventually` isn't broken in Objective-C.

### What actually happened instead?

Quick 6 added support for concurrency in a way that currently makes incremental adoption impossible.

The codebase that has to be upgraded to Quick 6 has a huge number specs across hundreds of modules and in order to upgrade all the problems introduced by the breaking changes have to be fixed in all those modules at once. Also generally the repo has a large amount of open PRs which would also need to be checked if they contain broken code and be fixed at the same time too.

Details:

#### Now tests run off the main thread (and off the MainActor) by default

- a lot of tests use Main actor-isolated properties from the iOS SDK, these tests now won't build
- a lot of tests use main thread only APIs which crash when called off the main thread

To temporarily fix these each spec could be marked as `@MainActor` but that wouldn't solve the cases where `it`s or `beforeEach`s and its variants are used in a local function, since inside those the closures passed won't be `@MainActor`:

```
@MainActor
final class SomeSpec: QuickSpec {
    override func spec() {
        describe("...") {
            context("...") {
                beforeEach { // this is @MainActor
                    ...
                }

                it { // this too
                    ...
                }

                func someFunc(...) {
                    beforeEach { // this isn't @MainActor
                        ...
                    }

                    it("...") { // this also isn't
                        ...
                    }
                }

                someFunc()
            }
        }
    }
}
```

Another possible temporary solution is to mark all the `async` closures in all the specs `@MainActor` but that would result in an immense amount of changes.

#### Now the closures passed to `it`, `beforeEach` and the variants of the latter are all `async` and there isn't a non-async alternative

- because of this (of course in addition to the above mentioned things) only the latest version of `Nimble` can be used which provides the `async` versions of `waitUntil`, `toEventually` and its variants and all the current usages can only be fixed all at once

- those are also used extensively and it's likely that numerous open PRs would also include new usages

#### `toEventually` is broken in Objective-C because it runs in an async context

Adding this testcase to `FunctionalTests_ItSpec_ObjC` shows that `toEventually` is broken:

```
[...]

Issue #1182, opened 2022-11-08
`QuickConfiguration` could inherit `@MainActor` attribute from spec class. 


matheusalano:
- [x] I have read [CONTRIBUTING](https://github.com/Quick/Quick/blob/main/CONTRIBUTING.md) and have done my best to follow them.

### What did you do?

When you have a `QuickSpec` with the `@MainActor` and a `QuickConfiguration` with (for example) an `afterEach`, the closure of this `afterEach` won't use the main actor. 

**The example below is using NSLog to show which process Id each closure is being run from. This is to easily tell from the console whether they’re both running on the same thread or not.**

```swift
@MainActor
final class Actor_Spec: QuickSpec {
    override func spec() {
        describe("mainActor") {
            it("should be true") {
                NSLog("Log from spec class")
                expect(Thread.isMainThread) == true
            }
        }
    }
}

final class ThreadTestConfiguration: QuickConfiguration {
    override class func configure(_ configuration: QCKConfiguration) {
        configuration.afterEach {
            NSLog("Log from configuration class")
        }
    }
}
```

### What did you expect to happen?

The process ID from the spec and configuration classes should match. 

### What actually happened instead?

The process ID from the spec and configuration classes don't match. 
Adding `configuration.afterEach { @MainActor in` makes the process IDs match. 

### Environment

List the software versions you're using:

 - Quick: v6.0.0
 - Nimble: v11.1.0
 - Xcode Version: 14.1
 - Swift Version: 5.7.1

Please also mention which package manager you used and its version. Delete the
other package managers in this list:

 - Carthage: 0.38.0

younata:
Huh. I 100% missed that line in [SE-0316](https://github.com/apple/swift-evolution/blob/main/proposals/0316-global-actors.md#using-global-actors-on-a-type) prior to this.

I don't think we should do this. I think that tagging the configuration closure with `@MainActor in` is sufficient enough. We don't expect there to be more than a few QuickConfiguration subclasses, so adding the `@MainActor in` syntax isn't that much of a lift. Additionally, not all the contents of every `configuration.X` will need to run on the main actor, and I feel that it's bad practice to unnecessarily force everything to run on the main actor.

I'm going to leave this open for now. I'd like to see more evidence/arguments for why you think this is desirable. But I don't think this is a convincing-enough reason to do this.

matheusalano:
Yeah, I honestly don't know if this suggestion would be a bad practice or if it's even possible.
[...]

Issue #1180, opened 2022-11-06
How to get rid of MainActor warning on all "it" tests?


Kalzem:
### What did you do?

Using Quick v6, I had to add a lot of @MainActor to it("...") tests such as

```swift
class TutorialManagerSpec: QuickSpec {
    override func spec() {
        describe("foo") {
            it("can do bar") { @MainActor in // <- Xcode Warning
                // Some test
            }
        }
    }
}
```

That's because of the architecture of the project which is based on singletons which must run on the main thread at all costs.

This generates an xcode warning for every test (so there are hundreds and hundreds of warnings).

> Converting function value of type '@ MainActor () -> ()' to '() throws -> Void' loses global actor 'MainActor'

Is there a specific syntax to get rid of this warning?

### Environment

List the software versions you're using:

 - Quick: 6.0.0
 - Nimble: 11.1.0
 - Xcode Version: 14.0.1
 - Swift Version: The one of 14.0.1
 - CocoaPods: 1.11.3

younata:
That's odd. You shouldn't be getting those warnings with Quick 6. It might be some weird caching behavior? I'll look into it some more later.

---

Regardless, to improve your developer experience: Since you're compiling with swift 5.7 (Xcode 14), you can annotate the entire class as `@MainActor`, and your tests will all run on the main actor without needing tag each individual test with `@MainActor`.

That is, instead of writing the following:

```swift
class MySpec: QuickSpec {
    override func spec() {
        it("") { @MainActor in
            expect(Thread.isMainThread).to(beTrue())
        }
    }
}
```

You can instead write:

```swift
@MainActor
class MySpec: QuickSpec {
    override func spec() {
        it("runs on the main actor") {
            expect(Thread.isMainThread).to(beTrue())
        }
    }
}
```

slava-semeniuk-ek:
It seems this doesn't work, since originally `func spec()` is not MainActor at `QuickSpec` class
```swift
@MainActor
class MySpec: QuickSpec {
    override func spec() {
        it("runs on the main actor") {
            expect(Thread.isMainThread).to(beTrue())
        }
    }
}
```

bvirlet:
Hello, 

Is there a recommend solution for this, or should `func spec()` be MainActor in `QuickSpec`? 

Thank you!

younata:
This should be resolved as of Quick 7. Please re-open if it isn't.

Issue #1179, opened 2022-11-03
beforeSuite running on the main actor when not tagged to do so. WAS: XCTestExpectation not working anymore in v6 in beforeSuite


Kalzem:
### What did you do?

I have a XCTestExpectation for an API fetch that I can't mock properly (third party SDK) so I call it in beforeSuite.
Before v6, the following code worked. Now it keeps failing even though `fulfill()` is indeed called.

```swift
final class PurchaseManagerSpec: QuickSpec {
    override func spec() {
        beforeSuite {
            let exp = self.expectation(description: "Fetching IAPs")
            PurchaseManager.shared.fetchAllProducts { status in
                switch status {
                case .success: break
                case .error: print("some errors but I don't care")
                }
                exp.fulfill()
            }
            self.waitForExpectations(timeout: 20) // <- Always failing with the wrong message
        }
    ...
}
```

I also tried adding `@MainActor` to beforeSuite, it doesn't change anything.

### What did you expect to happen?

v6 should work like v5 as I don't see mention of XCTestExpectation with breaking changes in the changelog.

### What actually happened instead?

XCTestExpectation not working in beforeSuite anymore.

Also I noticed that Xcode is lost and doesn't understand which test failed.

The issue comes from PurchaseManagerSpec but the error I get from line `self.waitForExpectations(timeout: 20)` is

```
An_AnalyticsManager__can_send_all_achievements_events:(): 
  Asynchronous wait failed: Exceeded timeout of 20 seconds, with unfulfilled expectations: "Fetching IAPs".
```

which is another Spec class.

### Environment

 - Quick: 6.0.0
 - Nimble: 11.1.0
 - Xcode Version: 14.0.1
 - Swift Version: The one of Xcode 14.0.1
 - CocoaPods: 1.11.3

younata:
I'm not able to replicate your error when doing this in a `beforeSuite`. Here's my test code:

```
final class FunctionalTests_XCTestExpectationSpec: QuickSpec {
    override func spec() {
        beforeSuite {
            let exp = self.expectation(description: "beforeSuite")
            Task {
                exp.fulfill()
            }
            self.wait(for: [exp], timeout: 10)
        }

        beforeEach {
            let exp = self.expectation(description: "beforeEach")
            Task {
                exp.fulfill()
            }
            self.wait(for: [exp], timeout: 10)
        }

        afterEach {
            let exp = self.expectation(description: "afterEach")
            Task {
                exp.fulfill()
            }
[...]

PR #1192, opened 2023-01-10
(Temporarily) Remove async support from Quick

younata:
Async support was released before it was ready.

There have been several issues about the way Quick forces async behavior. [Migrating is a massive pain in the ass. There's no way to get Nimble's `toEventually` to work with Objective-C tests](https://github.com/Quick/Quick/issues/1189). There's no clear way to force a test (or part of a test) to run on the main actor (the method [mentioned in the documentation](https://github.com/Quick/Quick/blob/main/Documentation/en-us/AsyncAwait.md#running-test-code-on-the-main-thread) doesn't work if you use the Swift Package Manager version of Quick). (issues: [1182](https://github.com/Quick/Quick/issues/1182), [1180](https://github.com/Quick/Quick/issues/1180), [1179](https://github.com/Quick/Quick/issues/1179)). It's been a mess.

Quick 7 will introduce a different (better) way of handling async. I'm still in the process of prototyping that out, but it will involve a separate test class (I'm using `AsyncSpec` for now). My current prototype is based on ResultBuilder, but I want to expand on that and see how it scales before committing to that direction. I'll post a draft PR when I have something semi-usable.

- This PR is targeting the `quick_7` branch. I'm going to target all of the async changes to this branch. I'll also keep it up to date with main to make this relatively easy to keep track of.
- The `main` branch will represent Quick 6 until I merge in the `quick_7` branch.


tahirmt:
Does this mean big changes in v7.0.0 potentially needing to revert many of the things that were changed in v6?

younata:
> Does this mean big changes in v7.0.0 potentially needing to revert many of the things that were changed in v6?

In a sense, yes. In v7, QuickSpec subclasses will revert back to the older, v5 behavior of running all tests synchronously, on the main thread. However, Async specs will 100% still be a thing in v7. They'll need to subclass a different class in order to get the async behavior (e.g. `class MyAsyncSpec: AsyncSpec` instead of `class MyAsyncSpec: QuickSpec`).

I'm still figuring out/testing how to get this all to work out. For example, [using result builders](https://github.com/Quick/Quick/pull/1194) was one attempt that didn't pan out. But it was certainly educational in terms of guiding other paths to try.

tahirmt:
Thank you for the context. The reason I asked is because I’ve spent the better part of the last month migrating our tests to Quick 6. A total of around 17000 tests. Previously we had written async versions of `it`, `beforeEach` etc. so we’d swap to `asyncIt` whenever we needed to break into an async context. This refactor of Quick 6 has been among the biggest in our tests so far. The branches are not merged yet still. 

Are you imagining `AsyncSpec` to behave how `QuickSpec` behaves in Quick 6 where the entire spec would run on the cooperative threads and we would need to use `MainActor` for things that require the main thread?

younata:
> This refactor of Quick 6 has been among the biggest in our tests so far. The branches are not merged yet still.

My apologies for the breakage I caused.

> Are you imagining AsyncSpec to behave how QuickSpec behaves in Quick 6 where the entire spec would run on the cooperative threads and we would need to use MainActor for things that require the main thread?

Yes. Additionally, I'd like to add some kind of convenience to force all of the closures to run on the main thread, but I don't know what that will look like yet.

PR #1214, opened 2023-05-10
[doc] Corrected to class method

coffmark:
Hi, I'm @coffmark. Thanks for always maintaining !

help you if you're not sure:

 - What behavior was changed?
   - I converted `override func spec() { ... }` to `override class func spec() { ... }` in docs.
   - Docs.
   - None.

 - [ ] Does this have tests?


younata (approved):
Omg, thank you so much! Really appreciate this.

PR #1222, opened 2023-06-08
Consolidate Quick-macOS, Quick-iOS and Quick-tvOS into a single Quick target

### LINKED ISSUE #1221: Integrate individual platform targets into a single Multiplatform target.
We currently have separate `Quick-macOS`, `Quick-iOS`, and `Quick-tvOS` targets. Multiplatform targets have been supported since Xcode 13. We should integrate these together into a single target. Would make adopting watchOS and visionOS so much simpler.

younata:
Fixes #1221

Makes it easier to maintain the xcodeproj and adding new platforms (sup visionOS) by consolidating our multiple, platform-dependent targets into a single Quick target (and one each for the different Quick tests).

This breaks the public API: People who embed the xcodeproj (i.e. using git submodules) will need to specify the Quick xcframework. Also, people who use Carthage (and are using intel) will now have to use the `--use-xcframeworks` option.

 - [x] Does this break the public API (Requires major version bump)?


Jesus13390 (commented):
QUE OPINA

PR #1233, opened 2023-07-06
Added a TestState property wrapper, again :D

CraigSiemens:
Adds a `TestState` property wrapper
- As suggested in #1178
- Then added in #1186
- Then removed in #1199

The main difference from the original implementation is it tries to detect which "World" is currently running tests, then add's the `afterEach`/`aroundEach` to that world. I'm not 100% sure if that's the right approach or if there's another way to add those to the current world.

From the original PR description

> It will always reset a property to nil at the end of a test. If the property wrapper is created with a value, the value will be set at the start of a test.

 - [x] Does this have tests?


younata (approved):
Thanks for re-adding this!

tahirmt:
@younata when it was removed from Quick, we created separate property wrappers namespaced within QuickSpec and AsyncSpec. Is this approach better? Is it possible to have any errors in the approach in this PR?

younata:
> Is this approach better?

There's no namespacing required with this approach. I don't know enough about how you implemented this to know whether it'll interfere with your approach.

> Is it possible to have any errors in the approach in this PR?

If XCTest ever does run test discovery in parallel, then there's the potential for errors as `AsyncWorld.sharedWorld.currentExampleGroup` and `World.sharedWorld.currentExampleGroup` could both be non-nil at the same time. Right now, XCTest runs test discovery synchronously. If this ever changes, we will have to either figure out a way to explicitly disable parallelized test discovery, or do a significant amount of engineering work to get Quick's DSL to work with parallelized test discovery.

I don't think it's worth worrying about parallelized test discovery at the moment.

I'm more than happy to release this as a beta, if you'd like. Otherwise, I was planning to release it as part of 7.2.0 on Friday morning.

younata:
Yes. That's correct. `currentExampleGroup` is determined at test discovery time, not during test runtime. Parallelized testing won't have any effect here.

PR #1240, opened 2023-08-31
Updated TestState to remove the value after the test.

CraigSiemens:
Currently, a property with `@TestState` will have its value reset to `nil` before any `afterEach` in the same `ExampleGroup` is run. This is because:
- the `@TestState` needs to be defined first so the tests can use it
- the `@TestState` was adding an `afterEach` to the current `ExampleGroup`
- the `afterEach` closures are run in the order they are defined in an `ExampleGroup`

This causes the `afterEach` resetting the `TestState` value to always be run before the `afterEach` closures defined in the same `ExampleGroup`.

This is an issue when a test needs to do additional cleanup using the value from `TestState. Ideally the value in the `TestState` would only be cleaned up after all the code in the test has finished running.

## Example

```swift
@TestState var tempDir: URL!
        
beforeEach {
    tempDir = FileManager.default
        .temporaryDirectory
        .appendingPathComponent(randomDirName())
    try? FileManager.default.createDirectory(at: tempDir, withIntermediateDirectories: true)    
}

afterEach {
    try? FileManager.default.removeItem(at: tempDir)
}
```

When this runs, the `afterEach` will crash because `tempDir` has already been reset to nil.

## Solution

~`TestState` has been updated to use an `afterEach` set in a `QuickConfiguration` to ensure that the value is reset after the current example group has finished running. This works because configurations behave like a parent `ExampleGroup` of the root `ExampleGroup` for the spec.~

~On one hand this does feel a little wierd to for Quick itself to be defining a configuration, but it was the simplest approach I could find.~

It's been updated to add a teardown block to the test case to set the value back to nil since that will always be run after the `afterEach` blocks.

### Alternatives considered
Update `ExampleGroup`/`ExampleHooks` to a `alwaysAfterEach` array that would function similar to `justBeforeEach` where it's a second array of closures that are always run last. This would have been a larger change to add support to the Sync/Async versions of `World`, `ExampleGroup`, and `ExampleHooks`

## Checklist

 - [x] Does this have tests?

### DIFF
--- a/Sources/Quick/TestState.swift
+++ b/Sources/Quick/TestState.swift
@@ -16,7 +16,7 @@ public struct TestState<T> {
     public init() {
         if AsyncWorld.sharedWorld.currentExampleGroup != nil {
             AsyncWorld.sharedWorld.beforeEach { [container] in
-                TestStateConfiguration.afterEach.append {
+                AsyncSpec.current.addTeardownBlock {
                     container.value = nil
                 }
             }
@@ -24,7 +24,7 @@ public struct TestState<T> {
 
         if World.sharedWorld.currentExampleGroup != nil {
             World.sharedWorld.beforeEach { [container] in
-                TestStateConfiguration.afterEach.append {
+                QuickSpec.current.addTeardownBlock {
                     container.value = nil
                 }
             }
@@ -35,7 +35,7 @@ public struct TestState<T> {
         if AsyncWorld.sharedWorld.currentExampleGroup != nil {
             AsyncWorld.sharedWorld.beforeEach { [container] in
                 container.value = wrappedValue()
-                TestStateConfiguration.afterEach.append {
+                AsyncSpec.current.addTeardownBlock {
                     container.value = nil
                 }
             }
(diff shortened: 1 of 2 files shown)


Sources/Quick/TestState.swift
```diff
@@ -15,53 +15,56 @@ public struct TestState<T> {
     /// Resets the property to nil after each test.
     public init() {
         if AsyncWorld.sharedWorld.currentExampleGroup != nil {
-            AsyncWorld.sharedWorld.afterEach { [container] in
-                container.value = nil
+            AsyncWorld.sharedWorld.beforeEach { [container] in
+                TestStateConfiguration.afterEach.append {
```
younata, line 19: I wonder if this might be a good place to use [`XCTestCase.addTeardownBlock(_:)`](https://developer.apple.com/documentation/xctest/xctestcase/3815521-addteardownblock), as that should run when XCTest would invoke the `tearDown` method, which will necessarily be after all of the `afterEach` blocks run.

younata (approved):
Thanks for this! Really appreciate it.

Also, would you like commit privileges to Quick and Nimble? You've been doing some great work, and I'd like to invite you to the Quick team.

Sources/Quick/TestState.swift
CraigSiemens: Nice. I updated it to use `QuickSpec.current` and `AsyncSpec.current` to get the current test case. Let me know if that that's not the right way to do that, it's the only way  could find.

CraigSiemens:
> Also, would you like commit privileges to Quick and Nimble? You've been doing some great work, and I'd like to invite you to the Quick team.

I could take it, though I probably wouldn't do much with it until I felt more comfortable with the repo.

younata (approved):

PR #1280, opened 2024-03-29
Add QuickLint, a CLI tool to detect focused specs, and optionally remove the focus.

younata:
Oftentimes, teams who use Quick's spec focusing either have to be really diligent about removing focus, or write a script to check for focus (and either block the push, or fail CI). Usually the latter is chosen. In an effort to alleviate writing those scripts, I wrote QuickLint. Which detects focused specs in the current directory, and also allows you to defocus the specs.

QuickLint also serves as an example of how to write testable code, and how to use Quick, [Nimble](https://github.com/Quick/Nimble), and the new [Swift Fakes](https://github.com/Quick/swift-fakes) packages.

Upcoming in this PR are also build tool and command plugins (for SwiftPM and Xcode) to lint and defocus your specs.

New feature, doesn't break the public API. So... minor version bump!

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mattgallagher/CwlCatchException.git",
       "state" : {
-        "revision" : "3b123999de19bf04905bc1dfdb76f817b0f2cc00",
-        "version" : "2.1.2"
+        "revision" : "3ef6999c73b6938cc0da422f2c912d0158abb0a0",
+        "version" : "2.2.0"
       }
     },
     {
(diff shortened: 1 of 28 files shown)


younata:
Apparently, to ship a binary with cocoapods, you have to actually build it yourself and then include that. Which is not something I'm interested in doing at the moment.

younata:
Lol, on Xcode 15.3, command plugins just... disappear after the first time you run them.

Screenshot 2024-03-29 at 8 08 12 PM

younata:
Ok, I think the actual commands are ready. Next steps are to figure out actually adding binaries (and jobs to build the binaries) so that we can ship quicklint with cocoapods & as pre-build commands.

PR #1283, opened 2024-05-06
named the context param aContext to match the actual protocol

Noobish1:
Hi there, in the SharedExamples docs, the use of the Behavior class is slightly incorrect in that it names the context param `context` instead of `aContext` like the class itself. This has the nasty consequence of shadowing any use of `context` blocks within the `spec` func.

### DIFF
--- a/Documentation/en-us/SharedExamples.md
+++ b/Documentation/en-us/SharedExamples.md
@@ -35,7 +35,7 @@ like `SomethingEdible`:
 
 ```swift
 class SomethingEdible: Behavior<Edible> {
-  override class func spec(_ context: @escaping () -> Edible) {
+  override class func spec(_ aContext: @escaping () -> Edible) {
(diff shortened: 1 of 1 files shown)


Documentation/en-us/SharedExamples.md
```diff
@@ -35,7 +35,7 @@ like `SomethingEdible`:
 
 ```swift
 class SomethingEdible: Behavior<Edible> {
-  override class func spec(_ context: @escaping () -> Edible) {
+  override class func spec(_ aContext: @escaping () -> Edible) {
     var edible: Edible!
     beforeEach {
       edible = context()
```
younata, line 41: This should be using `aContext` to match the new variable name.

```suggestion
      edible = aContext()
```

younata (approved):
Great catch! Thanks!

PR #1289, opened 2024-06-21
Updated closures in QuickSpec to be annotated with MainActor

CraigSiemens:
This is following the discussion I started here.
- https://github.com/Quick/Quick/discussions/1288

## Background

`XCTestCase` always calls synchronous tests on the main thread. This is to support testing UIKit types which can't be used on background threads. 

However when testing custom types annotated with `@MainActor` the compiler produces errors on uses in the tests like the following.

```
Main actor-isolated property 'stateSubject' can not be referenced from a non-isolated context
```
```
Call to main actor-isolated initializer 'init()' in a synchronous nonisolated context
```

For some unknown reason, the concurrency checking doesn't give the same errors for `@MainActor` UIKit types.

## Changes

This PR adds `@MainActor` annotations to the `QuickSpec` before/after/it closures so the compiler knows that they will be executed on the `MainActor`, matching the existing runtime behavour.

This allows tests to use types annotated with `@MainActor` without needing to switch to `AsyncSpec` and annotate every closure with `@MainActor`.

 - Does this have tests?
    - No. I'm not sure how to write a test for how the project compiles. I build quick with complete concurrency checking enabled and no new warnings were added. I ran the Quick tests and tried it with our project's tests and both successfully ran.
      
 - Does this have documentation?
   - No since the behaviour didn't change, it's just adding details so the compiler knows about it.
 - Does this break the public API (Requires major version bump)? / Is this a new feature (Requires minor version bump)?
   - The only changes we had to make in out project were to a couple funcs (< 10) defined in `context` and `describe` that now required a `@MainActor` annotation so they can be called in `it`. The scope of the impact will entirely depend on the code base, but it could warrant a major version bump.

### DIFF
--- a/Sources/Quick/DSL/DSL.swift
+++ b/Sources/Quick/DSL/DSL.swift
@@ -205,7 +205,7 @@ extension SyncDSLUser {
      - parameter file: The absolute path to the file containing the example. A sensible default is provided.
      - parameter line: The line containing the example. A sensible default is provided.
      */
-    public static func it(_ description: String, file: FileString = #file, line: UInt = #line, closure: @escaping () throws -> Void) {
+    public static func it(_ description: String, file: FileString = #file, line: UInt = #line, closure: @escaping ExampleClosure) {
         World.sharedWorld.it(description, file: file, line: line, closure: closure)
     }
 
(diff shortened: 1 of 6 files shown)


younata (approved):
Yeah, this makes sense to me.

tahirmt:
@younata when can we expect this to be released?

younata:
@tahirmt thanks for the reminder! [7.6.1](https://github.com/Quick/Quick/releases/tag/v7.6.1) is available now!

CraigSiemens:
@seth-folley Filing a new issue with the steps to reproduce it would be the best.

I built Quick with Xcode 16.0 Beta 6, Swift 6, and `SWIFT_STRIT_CONCURRENCY = Complete` and while it did have errors, none match the warning you've posted so I'm unsure what the cause could be.

seth-folley:
@CraigSiemens will do. This isn't an issue with building Quick, rather an issue using `QuickSpec` with strict concurrency enabled.

seth-folley:
https://github.com/Quick/Quick/issues/1298

PR #1292, opened 2024-07-03
Spike: BDD DSL on top of Swift Testing

younata:
Swift Testing is the new, mostly open source testing framework for Swift. It provides a pytest-style approach to testing, with native support for nested tests, like what Quick and other BDD-style frameworks provide.

This PR is an attempt to create a Quick-Style DSL on top of Swift Testing. This work explicitly does not support Objective-C, using resultbuilders to build its DSL.

This PR, for now, adds a new library target, QuickTesting, just to provide a separate space for these new type names. I intend to later merge this new target back in to the main Quick target.

---
Previous Work

I previously tried to implement Quick's Async support using resultbuilders, which was quite the learning experience: https://github.com/Quick/Quick/pull/1194

### DIFF
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -1,0 +1,115 @@
+// swift-tools-version:6.0
+
+import PackageDescription
+
+let package = Package(
+    name: "Quick",
+    platforms: [
+        .macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .visionOS(.v1)
+    ],
+    products: [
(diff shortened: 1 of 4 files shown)


sushant-here:
Thanks for giving this a go @younata - Ive started looking at SwiftTesting and can say its definitely a step up from XCTest! Keen to use it. However I really like BDD and the test structure that Quick provides. So i am stuck between a rock and a hard place!

I am keen to see where you get with this. Happy to get involved with any beta testing/feedback you may be looking for.

PR #1300, opened 2024-09-09
Added MainActor annotations to the example group closures

CraigSiemens:
This adds `@MainActor` annotations to the `describe` and `context` closures as well as `QuickSpec.spec()`.

This helps resolve a warning when using Swift 6 with complete concurrency checking enabled when a variable is defined in `spec()` or an example group (`describe`/`context`) and used in a test. The compiler would warn when accessing a non-isolated variable from an isolated context (`it`). Now everything has the same isolation.


CraigSiemens:
It looks like the builds are failing because of the call to `MainActor.assumeIsolated` which was introduced in [Swift 5.9](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md).

Some of those workflows would need to be updated to use newer versions of Xcode and Swift. I'm no sure if that's something that should be done in this PR or another one.

Noobish1:
Any updates on this PR?

PR #1322, opened 2025-04-06
Bump Cwl dependencies to fix Xcode 16 issues

balavor:
Bumped SPM dependencies that cause issues when building targets with a `carthage`, which takes dependencies straight from `.resolved` file, using ones that have to build issues with Xcode 16.2.

image

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mattgallagher/CwlCatchException.git",
       "state" : {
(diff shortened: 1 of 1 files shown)


younata (approved):

younata:
Thanks for doing this!

balavor:
@younata, no worries! Would you also be able to bump `Quick/Nimble` version in `Package.resolved` file before releasing the new `Quick` version (e.g. from `"13.2.1"` to `"13.7.2"` or whatever comes next)? 🙏

